PR #2544, opened 2026-03-27
Fix owner preselection mismatch on /portfolio and /performance root routes

### LINKED ISSUE #2539: Bug: Portfolio and Performance pages show "Select an owner/member" despite owner being pre-selected in dropdown
## Description

On the Portfolio (`/portfolio`) and Performance (`/performance`) pages, the owner dropdown is pre-populated with a default value (e.g., "Alex Leonard"), but the page displays a message saying "Select an owner." or "Select a member." and shows no data. The user must manually re-select the owner from the dropdown to trigger data loading.

## Steps to Reproduce

1. Navigate directly to `/portfolio` or `/performance`
2. 2. Observe the owner dropdown shows "Alex Leonard" already selected
3. 3. Observe the page content shows "Select an owner." (portfolio) or "Select a member." (performance) with no data
## Expected Behaviour

If a default owner is pre-selected in the dropdown on page load, the data should load automatically without requiring the user to re-interact with the dropdown.

## Root Cause

The component likely initialises state with a default value but doesn't trigger the data fetch/effect when the dropdown has a pre-selected value. The `onChange` handler is only called when the user actively changes the selection, not on initial render with a default value.

## Fix Suggestion

Either:
1. Use `useEffect` to trigger data loading when the component mounts if an owner value is already set
2. 2. Or initialise the state to empty/null and don't pre-fill the dropdown (consistent UX - force explicit selection)
## Affected Pages

- `/portfolio` - shows "Select an owner."
- - `/performance` - shows "Select a member."

leonarduk:
### Motivation
- Users navigating to `/portfolio` or `/performance` could see a pre-selected owner in the dropdown while the app state remained empty and no data was loaded, producing a stale "Select an owner/member" UI state.
- The route-handling logic only triggered a default owner for portfolio when there was exactly one owner, and did not handle the performance root at all, causing inconsistent behaviour.
- Closes #2539 

### Description
- Updated `frontend/src/App.tsx` to default `selectedOwner` and navigate to a concrete owner route when visiting the root owner routes; `/portfolio` now chooses the first available owner when `owners.length > 0` and `/performance` receives equivalent handling. (`frontend/src/App.tsx`) 
- Broadened portfolio defaulting from `owners.length === 1` to `owners.length > 0` to keep UI state and route consistent when owners are preloaded.
- Added unit tests to `frontend/tests/unit/App.test.tsx` to cover both flows: redirecting `/portfolio` to `/portfolio/:owner` and `/performance` to `/performance/:owner`, and verifying the portfolio fetch and performance dashboard render for the chosen owner.

### Testing
- Ran the focused frontend unit tests: `npm --prefix frontend run test -- App.test.tsx --run` and verified the updated `App.test.tsx` tests passed. 
- The added/updated tests exercise the new defaulting logic and assert that the portfolio fetch and performance dashboard render for the first owner; these tests passed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c6ebd724e48327b3ca017d4c397431)

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -348,13 +348,14 @@ export default function App({ onLogout }: AppProps) {
         ? "performance"
         : null;
 
-    const fallbackOwner = owners[0]?.owner ?? selectedOwner;
-    if (ownerRoot && fallbackOwner) {
-      // Intentionally auto-select the first available owner on owner-root routes
-      // so route, selected owner, and loaded data stay consistent.
-      const targetPath = `/${ownerRoot}/${fallbackOwner}`;
-      if (selectedOwner !== fallbackOwner) {
-        setSelectedOwner(fallbackOwner);
+    // Only redirect when we are at an owner-root route AND at least one owner has loaded.
+    // Do not fall back to selectedOwner — that could redirect to an unverified owner
+    // when the owners list is empty (e.g. stale state during async load).
+    if (ownerRoot && owners.length > 0) {
+      const firstOwner = owners[0].owner;
+      const targetPath = `/${ownerRoot}/${firstOwner}`;
+      if (selectedOwner !== firstOwner) {
+        setSelectedOwner(firstOwner);
       }
       if (location.pathname !== targetPath) {
         navigate(targetPath, { replace: true });


leonarduk:
@codex

Context: PR #2544 is meant to fix the stale preselected-owner mismatch on `/portfolio` and `/performance` root routes so the URL, selected owner, and loaded data stay in sync.

**Real blockers**

* The current change appears to fix the route handling, but the remaining blocker is not feature intent; it is proof and correctness around the broadened auto-selection behavior.
* This PR now auto-selects the first owner whenever `owners.length > 0`, not just when there is exactly one owner. That is a real UX change for multi-owner accounts, not just a bug fix. Either keep it deliberately and make the behavior explicit/consistent, or narrow it. Do not leave it implicit.
* The test suite must prove the redirect logic is stable and not just that a component eventually renders. We need evidence that the redirect happens to the expected concrete route, happens once, and does not override explicit owner routes.
* Ignore `security/snyk` here; it is quota-limited and this PR does not change security-sensitive code.

**Required code fixes**

* Keep the root-route redirect logic route-driven and consistent for both `/portfolio` and `/performance`.
* Make the intended behavior explicit in code: visiting an owner-root route with no selected owner and at least one available owner should replace-navigate to the first owner route. If that is not the intended product behavior for multi-owner households, restore the narrower guard and solve the stale-state bug another way.
* Ensure explicit owner routes such as `/portfolio/:owner` and `/performance/:owner` are treated as authoritative and are not cleared or overridden by fallback selection logic.
* Keep `navigate(..., { replace: true })`.

**Required tests / validation**

* Update or keep tests that assert actual route transition to:

  * `/portfolio/alice`
  * `/performance/alice`
* Add or keep explicit coverage for:

  * `/portfolio` with multiple owners -> redirects to first owner route
  * `/performance` with multiple owners -> redirects to first owner route
  * `/portfolio` with zero owners -> no redirect, no crash
  * `/performance` with zero owners -> no redirect, no crash
  * owners loading asynchronously after initial render -> redirect occurs once when owners become available
  * direct `/portfolio/:owner` and `/performance/:owner` routes -> no override to first owner
* Run the relevant frontend test file and report a clean passing result. Do not claim merge readiness without a clean pass.

**Scope constraints**

* Out of scope: `security/snyk` quota failure, unrelated refactors, stylistic cleanup, comment-only churn.
* Do not weaken the fix by relying on render-based assertions instead of route assertions.

**Reply format**
Root cause:
Code changes made:
Tests added/updated:
Validation run:
Remaining concerns or none:

PR #2545, opened 2026-03-27
Ignore non-finite/negative holding values in allocation charts

### LINKED ISSUE #2540: Bug: Allocation page chart not rendering (empty chart area)
## Description

On the Allocation page (`/allocation`), all three tabs (Instrument Types, Industries, Regions) show an empty chart area. The totals legend at the bottom is visible, but the chart itself does not render any visualisation.

## Steps to Reproduce

1. Navigate to `/allocation`
2. 2. Observe a large blank/empty chart area below the account checkboxes
3. 3. The bottom of the page shows the legend with correc

leonarduk:
### Motivation
- Allocation pie charts on the `/allocation` page were rendering as a blank area (Closes #2540) because invalid holding values (e.g. `NaN` or negative numbers) produced invalid slices for Recharts.
- The goal is to harden the frontend aggregation so charts receive only safe numeric inputs and do not fail silently.

### Description
- Added an `asFiniteNumber` helper in `frontend/src/pages/AllocationCharts.tsx` that coerces values to finite numbers and returns `0` for non-finite inputs. 
- Aggregation now skips non-positive values (`<= 0`) to avoid feeding invalid slices into the pie chart. 
- Added a regression unit test `ignores non-positive values without crashing` in `frontend/tests/unit/pages/AllocationCharts.test.tsx` that supplies negative/`NaN`/positive holdings and verifies the page renders.
- No backend or API contract changes; this is a defensive frontend-only change that drops zero/negative holdings from pie datasets.

### Testing
- Ran the targeted unit tests with `npm --prefix frontend run test -- AllocationCharts.test.tsx --run` and the test file passed (`3 tests, 0 failures`).
- Ran the repo lint command `npm --prefix frontend run lint -- src/pages/AllocationCharts.tsx tests/unit/pages/AllocationCharts.test.tsx`, which surfaced pre-existing repo-wide lint errors unrelated to this change; these lint issues did not affect the targeted unit test outcome.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c6ebefe2a883279d5dc3ff56c4ec43)

### DIFF
--- a/frontend/src/pages/AllocationCharts.tsx
+++ b/frontend/src/pages/AllocationCharts.tsx
@@ -37,10 +37,7 @@ const isInvalidNumericInput = (value: unknown): boolean => {
   return !Number.isFinite(numeric);
 };
 
-export const isDevEnvironment = (): boolean => import.meta.env.DEV;
-export const allocationChartRuntime = {
-  isDev: isDevEnvironment(),
-};
+const isDevEnvironment = (): boolean => import.meta.env.MODE !== "production";
 
 export type AllocationChartsProps = {
   /** Portfolio group slug (defaults to "all"). */
@@ -117,14 +114,21 @@ export function AllocationCharts({ slug = "all" }: AllocationChartsProps) {
         const originalMarketValue = h.market_value_gbp as unknown;
         const mv = toFiniteNumber(originalMarketValue);
         const originalInvalid = isInvalidNumericInput(originalMarketValue);
-        // Visualization safeguard only: exclude zero values (no drawable slice) and negatives (short positions/invalid data).
-        if (mv <= 0) {
-          if (allocationChartRuntime.isDev) {
+        const dropReason = originalInvalid
+          ? "invalid-numeric-input"
+          : mv <= 0
+            ? "non-positive-market-value"
+            : null;
+        // Visualization safeguard only: exclude invalid numeric input and non-positive values.
+        // Zero creates no drawable slice and negatives typically represent short/invalid values; this does not fix upstream data quality.
+        if (dropReason) {
+          if (isDevEnvironment()) {
             console.warn("Dropped invalid holding value", {
               ticker: h.ticker,
               originalValue: originalMarketValue,
               coercedValue: mv,
               originalInvalid,
+              dropReason,
             });
           }
           continue;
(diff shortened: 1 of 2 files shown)


leonarduk:
@codex PR #2545 — fix allocation chart empty renders without masking data issues or weakening test isolation

**Real blockers**

1. **The current warning/testability design is fragile.** `allocationChartRuntime` is a mutable exported singleton used by tests to flip dev/prod behavior. That is unnecessary shared state and a bad pattern for merge-ready frontend code.
2. **The filtering guard is still less explicit than it should be.** The logic currently relies on `toFiniteNumber(...)` collapsing invalid inputs to `0`, then `mv <= 0` filtering them out. That works, but it hides the real decision path. The guard should explicitly reflect both cases: invalid numeric input and non-drawable values.
3. **Coverage is better, but still needs one missing regression case:** `null` / `undefined` handling, because the old code used `?? 0` and this PR changed that behavior path.

**Required code fixes**

* Remove the mutable exported runtime object. Do not keep `allocationChartRuntime` as shared mutable module state just for tests.
* Replace that with a small internal helper that checks dev mode directly at use time, using Vite semantics (`import.meta.env.DEV`). Tests should mock/stub the warning path cleanly rather than mutating exported state.
* Make the drop condition explicit in the aggregation loop. Keep conversion and classification separate:

  * conversion helper: numeric coercion only
  * invalidity helper: identify NaN / Infinity / non-numeric input from the original value
  * filter guard: explicitly drop when `originalInvalid || mv <= 0`
* Keep the current chart semantics: **do not reintroduce zero, negative, NaN, or Infinity values into chart slices.**
* Keep the dev-only warning, but ensure it is driven by the **original input** and drop reason, not only by the derived `mv`.

**Required test/validation work**

* Add a blocking unit test for `market_value_gbp: null`.
* Add a blocking unit test for `market_value_gbp: undefined`.
* Keep the existing positive assertions across type/sector/region; those are required.
* Keep the all-invalid empty-state assertion.
* Keep production warning suppression coverage.
* Run and report:

  * `npm --prefix frontend run test -- AllocationCharts.test.tsx --run`
  * `npm --prefix frontend run test`
* For the full frontend test run, distinguish clearly between:

  * failures introduced by this PR
  * pre-existing unrelated failures
* Manual validation still required on `/allocation`:

  * mixed valid/invalid holdings => valid slices render on all tabs
  * all-invalid holdings => explicit empty state, no blank crash
  * dev console shows warnings only in dev, not production

**Scope constraints**

* Out of scope: backend validation, repo-wide lint failures, unrelated frontend suite failures, `security/snyk` quota failure.
* Do not widen this into a data-model redesign.
* Do not change the API contract.
* Do not add generic cleanup unrelated to allocation chart correctness.

**Reply format**
Reply with exactly these sections:

1. `Blockers fixed`
2. `Code changes`
3. `Tests added/updated`
4. `Validation run`
5. `Manual /allocation behaviour`
6. `Remaining out-of-scope issues`

In `Validation run`, list the exact commands and whether each passed or failed, and for any failure state whether it is unrelated to this PR.

PR #2561, opened 2026-03-27
Add TypeScript no-emit check to frontend CI workflow

### LINKED ISSUE #2556: CI: Add tsc --noEmit type-checking step to frontend workflow
## Problem

`frontend-tests.yml` runs `pnpm test` (Vitest) but does not run `tsc --noEmit`. TypeScript compilation errors therefore do not block merges.

This matters because `types.ts` is the contract boundary between frontend and backend. PR #2548 made `IndexPerformance.change` non-optional (`change?: number | null` → `change: number`). If any component was still passing `null` for `change`, TypeScript would catch it — but only if `tsc` runs in CI.

## Proposed change

Add a `tsc` step to `frontend-tests.yml`:

```yaml
- name: TypeScript type-check
  working-directory: frontend
  run: pnpm tsc --noEmit
```

This should run before the test step so type errors surface with a clean signal rather than being obscured by test failures.

## Acceptance criteria

- [ ] `tsc --noEmit` runs as a required step in `frontend-tests.yml`
- [ ] The step fails the workflow on TypeScript compilation errors
- [ ] The step passes cleanly on the current `main` branch (fix any pre-existing errors first, or add them to `tsconfig` excludes with a comment)
- [ ] A follow-up task is created if pre-existing errors are suppressed rather than fixed

## Notes

- `tsconfig.json` must include all source files that are compiled for production; check that test files are excluded (Vitest typically handles its own type-checking separately).
- This is distinct from the linting step — `tsc` checks type correctness, ESLint/ruff checks style.

leonarduk:
### Motivation
- The frontend CI ran Vitest without a dedicated `tsc --noEmit` step, so TypeScript compilation errors could pass CI and hide cross-boundary contract regressions.
Closes #2556 

### Description
- Added a `TypeScript type-check` step to `.github/workflows/frontend-tests.yml` that runs `pnpm tsc --noEmit` in the `frontend` directory after dependencies are installed and before the test/coverage step.

### Testing
- Executed the equivalent local check with `cd frontend && npx tsc --noEmit`, which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c7069b82e88327b15ca5c100f3ec29)

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -32,9 +32,9 @@ jobs:
         working-directory: frontend
         run: pnpm install --no-frozen-lockfile
 
-      - name: TypeScript type-check
+      - name: TypeScript type-check (tsconfig.app.json)
         working-directory: frontend
-        run: pnpm tsc --noEmit
+        run: pnpm run type-check
 
       - name: Run tests with coverage
         working-directory: frontend


leonarduk:
@codex

**Context**
PR #2561 – add a TypeScript no-emit check to frontend CI to catch type errors before tests.

---

**Real blocker(s)**

1. **Incorrect / unsafe invocation of TypeScript compiler**
   The workflow uses `pnpm tsc --noEmit`, which may:

   * pick up the wrong `tsconfig` (default vs app-specific)
   * ignore project references / composite configs
   * diverge from how developers run type checks locally

   This creates a false sense of safety: CI may pass while real type errors still exist.

2. **No proof this step actually validates the intended project scope**
   The PR only tested `npx tsc --noEmit` locally, which has the same flaw.
   There is no evidence this matches the repo’s intended type-check entrypoint.

---

**Required code fixes**

1. **Use the repo’s canonical type-check command**

   * Inspect `frontend/package.json`
   * If a script exists (e.g. `"type-check": "tsc --noEmit -p tsconfig.app.json"`), update workflow to:

     ```yaml
     - name: TypeScript type-check
       run: pnpm run type-check
       working-directory: frontend
     ```
   * If no script exists:

     * Add one explicitly targeting the correct config:

       ```json
       "type-check": "tsc --noEmit -p tsconfig.json"
       ```
     * OR explicitly pass `-p <correct tsconfig>` in CI (do not rely on implicit resolution)

2. **Ensure local/CI parity**

   * The CI command must be identical to what devs run locally
   * Do not call `tsc` directly in CI if the repo defines a script

3. **(Optional but correct) use `pnpm exec` if calling binary directly**
   Only if no script is used:

   ```yaml
   run: pnpm exec tsc --noEmit -p <config>
   ```

---

**Required test / validation**

You must prove this step is actually effective:

1. **Positive validation (baseline)**

   * Run the CI command locally from repo root:

     ```
     cd frontend && pnpm run type-check
     ```
   * Confirm it passes

2. **Negative validation (critical)**

   * Introduce a deliberate type error (e.g. wrong prop type in a component)
   * Confirm:

     * the type-check step fails
     * CI fails before tests run

3. **Config coverage validation**

   * If multiple tsconfigs exist (common in Vite setups), confirm:

     * the selected config covers the actual app code (not just a subset)

---

**Scope constraints**

* Ignore `security/snyk` failure (quota-related, not caused by this PR)
* Do not modify test workflows, coverage, or unrelated CI steps
* Do not introduce new tooling (no eslint/tsc integration changes)
* Focus only on making type-checking **correct and authoritative**

---

**Required reply format**

Reply with:

1. The exact workflow diff (before → after)
2. The `package.json` script (if added/updated)
3. Which `tsconfig` is used and why
4. Evidence of:

   * passing run
   * failing run with an injected type error

Do not claim “works locally” without showing the exact command used.

PR #2607, opened 2026-03-29
Scope s3:ListBucket to bucket prefix and require list_prefix

### LINKED ISSUE #2586: Scope s3:ListBucket with s3:prefix condition in Lambda IAM grants
## Context

PR #2574 tightened Lambda S3 IAM permissions, replacing `grant_read_write` with per-function scoped grants via `_grant_bucket_access`. The `s3:ListBucket` grant is applied to the bucket ARN without a `StringLike` condition on `s3:prefix`.

Without a prefix condition, any Lambda with `s3:ListBucket` can enumerate **all objects** in the data bucket, not just the objects relevant to its function. For a shared-bucket architecture holding financial portfolio data, this is a least-privilege gap.

Related to #2570. Identified during review of #2574.

## Work required

1. Define a per-function S3 prefix scheme (e.g. `prices/`, `portfolios/`, `agent/`).
2. Update `_grant_bucket_access` in `cdk/stacks/backend_lambda_stack.py` to accept an optional `list_prefix` parameter.
3. When `list_prefix` is provided, emit the `s3:ListBucket` statement with a condition:
   ```python
   iam.PolicyStatement(
       actions=["s3:ListBucket"],
       resources=[f"arn:aws:s3:::{bucket_name}"],
       conditions={
           "StringLike": {"s3:prefix": [f"{list_prefix}/*"]}
       },
   )
   ```
4. Update each `_grant_bucket_access` call site with the appropriate prefix.
5. Update the CDK assertion tests to verify the prefix condition is present on `ListBucket` statements.

## Acceptance criteria

- `s3:ListBucket` for each Lambda role is conditioned on `s3:prefix` matching only that function's namespace in the bucket.
- CDK tests assert the condition is present and correctly scoped.
- No Lambda can enumerate objects outside its designated prefix.

## Notes

This work requires agreeing on a bucket prefix scheme first. If the codebase does not yet organise S3 objects under per-function prefixes, that naming convention needs to be established as part of this issue before the IAM change can be applied meaningfully.

leonarduk:
### Motivation

- Limit Lambda IAM permissions to least-privilege by scoping `s3:ListBucket` to a specific prefix instead of granting broad bucket access.
- Ensure code explicitly requires a `list_prefix` when `allow_list=True` to avoid accidental overly-broad list grants.
Closes #2586 
### Description

- Added a `list_prefix: str | None` parameter to `BackendLambdaStack._grant_bucket_access` and normalize/validate it when `allow_list=True` by raising `ValueError` for empty prefixes.
- Conditioned the `s3:ListBucket` policy statement with `{"StringLike": {"s3:prefix": ["{normalized_prefix}/*"]}}` so list operations are restricted to the specified prefix.
- Introduced `backend_list_prefix = "portfolio"` and passed it to the backend Lambda call to `_grant_bucket_access` so backend list access is limited to `portfolio/*`.
- Removed the previous blanket `data_bucket.grant_read_write(...)` loop to avoid broad role grants and rely on the fine-grained policies instead.
- Updated unit tests by adding `_conditions_for_s3_action` and asserting that `s3:ListBucket` has an IAM `Condition` with `portfolio/*`.

### Testing

- Ran the stack unit tests in `cdk/tests/test_backend_lambda_stack_permissions.py` which include `test_s3_permissions_are_scoped_per_lambda` and `test_lambda_roles_do_not_have_s3_delete_permissions`, and they passed. 
- Executed `pytest -q` against the CDK tests and observed no failures.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c94e2479548327ab1d2959ee01654e)

### DIFF
--- a/cdk/stacks/backend_lambda_stack.py
+++ b/cdk/stacks/backend_lambda_stack.py
@@ -1,5 +1,6 @@
 import os
 from pathlib import Path
+from typing import Sequence
 
 from aws_cdk import CfnOutput, Duration, Stack
 from aws_cdk import aws_apigatewayv2 as apigwv2
@@ -27,7 +28,7 @@ class BackendLambdaStack(Stack):
         allow_read: bool,
         allow_put: bool,
         allow_list: bool,
-        list_prefix: str | None = None,
+        list_prefix: str | Sequence[str] | None = None,
     ) -> None:
         """Grant the minimum required S3 actions for a Lambda function.
 
@@ -50,10 +51,23 @@ class BackendLambdaStack(Stack):
             )
 
         if allow_list:
-            normalized_prefix = (list_prefix or "").strip().strip("/")
-            if not normalized_prefix:
+            raw_prefixes: list[str]
+            if isinstance(list_prefix, str):
+                raw_prefixes = [list_prefix]
+            elif list_prefix is None:
+                raw_prefixes = []
+            else:
+                raw_prefixes = list(list_prefix)
+
+            normalized_prefixes = [prefix.strip().strip("/") for prefix in raw_prefixes]
+            normalized_prefixes = [prefix for prefix in normalized_prefixes if prefix]
+            if not normalized_prefixes:
                 raise ValueError("list_prefix is required when allow_list=True")
-            prefix_conditions = [normalized_prefix, f"{normalized_prefix}/*"]
+
+            prefix_conditions: list[str] = []
+            for prefix in normalized_prefixes:
+                prefix_conditions.append(prefix)
+                prefix_conditions.append(f"{prefix}/*")
             fn.add_to_role_policy(
                 iam.PolicyStatement(
                     actions=["s3:ListBucket"],
@@ -104,9 +118,9 @@ class BackendLambdaStack(Stack):
 
         bucket_name = data_bucket.bucket_name
         lambda_list_prefixes = {
-            "backend": "portfolio",
-            "price_refresh": None,
-            "trading_agent": None,
+            "backend": ("accounts", "queries", "timeseries/meta", "transactions"),
+            "price_refresh": (),
+            "trading_agent": (),
         }
 
         seed_data_bucket = (
@@ -165,8 +179,11 @@ class BackendLambdaStack(Stack):
         app_secret.grant_read(backend_fn)
 
         # BackendLambda: read + put + list
-        # Audited: serves API requests that read and write portfolio/price data to S3.
-        # s3:ListBucket required: backend serves listing endpoints (e.g. portfolio enumeration).
+        # Audited S3 list prefixes used by backend code paths:
+        # - accounts/        (auth + portfolio enumeration)
+        # - queries/         (saved query listing)
+        # - timeseries/meta/ (timeseries admin listing)
+        # - transactions/    (report transaction exports)
         self._grant_bucket_access(
             backend_fn,
             bucket_name=bucket_name,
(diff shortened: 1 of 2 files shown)


leonarduk:
@codex PR #2607 is intended to make S3 `ListBucket` least-privilege by scoping listing to explicit prefixes and requiring `list_prefix` when `allow_list=True`.

**Real blockers**

1. The current backend list scope is still wrong unless it covers every prefix the backend actually enumerates. The Codex connector review identified concrete existing list paths under `accounts/`, `queries/`, and `timeseries/meta/`. If the backend role is limited to `portfolio` only, this is a runtime regression, not a security improvement.
2. The PR needs to prove that removing the old broad `grant_read_write(...)` loop did not break object read/write access for any Lambda that still needs it. Claude’s first review was right to flag this risk. The latest diff may have fixed it, but the PR description and tests still do not prove it cleanly enough.
3. The prefix condition shape must match real S3 list usage. Do not guess. If the code lists with prefixes like `portfolio/`, `accounts/`, `queries/`, etc., the IAM condition must allow exactly those request prefixes. A superficially “scoped” condition that rejects legitimate list calls is still broken.

**Required code fixes**

1. Audit all Lambda code paths that call S3 list APIs and map them to required prefixes per function.

   * For `backend_fn`, include every actually-used listed prefix.
   * For `refresh_fn` and `agent_fn`, either:

     * keep `allow_list=False` and prove they never list, or
     * assign explicit prefixes and grant only those.
2. Encode that mapping explicitly in the stack code. Do not leave `None` placeholders that imply an unfinished scheme.
3. Keep `list_prefix` mandatory when `allow_list=True`, but make the implementation reflect the real contract clearly:

   * either a single prefix string if one prefix is sufficient,
   * or a list/tuple of prefixes if the backend genuinely needs multiple prefixes.
     Do not hide multiple required prefixes behind one misleading `"portfolio"` value.
4. Preserve least privilege for object actions as well. If the old broad grant was removed, ensure equivalent read/write permissions still exist only where needed. Do not reintroduce blanket bucket-wide grants as a shortcut.
5. Keep out of scope any unrelated security tooling noise. `security/snyk` quota/limit failures are irrelevant here unless this PR changed something genuinely security-sensitive beyond IAM policy shaping.

**Required tests / validation**

1. Add/adjust CDK assertions so they prove the actual intended policy, not just that “some condition exists”:

   * exact `s3:prefix` conditions for every Lambda with list access,
   * explicit absence of `s3:ListBucket` for Lambdas that should not list,
   * explicit object action assertions for roles that still need read/write.
2. Add a unit test that `allow_list=True` with missing/empty `list_prefix` raises `ValueError`.
3. Add a test that would have caught the backend regression:

   * assert backend list permissions cover all real backend-enumerated prefixes identified from the codebase.
4. Run the full relevant CDK test file and include the command/output summary in the reply.
5. Run a quick repo grep or equivalent validation over S3 list call sites and include the concrete prefixes found. Do not claim this is fixed without showing that mapping.

**Scope constraints**

* Do not broaden permissions “just to make tests pass.”
* Do not change application behavior, prefix layout, or add new storage conventions unless required to match existing code paths.
* Do not spend time on naming micro-cleanups or comment polish unless they clarify the IAM contract materially.
* Ignore weak bot feedback that merely says the PR “looks fine.” The only review points that matter are the concrete runtime-permission risks.

**Reply format**
Reply with:

1. `Blockers addressed:` bullet list
2. `Code changes:` bullet list with exact Lambda→prefix mapping
3. `Tests added/updated:` bullet list
4. `Validation run:` exact commands + concise results
5. `Remaining risks:` bullet list or `None`

PR #2622, opened 2026-03-29
Relax key findings length validation and escalate oversized entries (fix #2614)

### LINKED ISSUE #2614: Bug: key findings silently dropped if > 240 chars — undocumented limit truncates valid content
## Gap vs MONTH1_PLAN.md

Issue #2580 AC: *"findings render as readable paragraphs, not table rows"*

The `_parse_key_findings_text` function in `backend/reports.py` silently drops any finding string where `len(value) < 20 or len(value) > 240`:

```python
if len(value) < 20 or len(value) > 240:
    logger.warning("Skipping invalid key finding from %s: %s", source_name, value)
    continue
```

The 240-character upper bound is undocumented and will silently truncate legitimate findings. A well-written finding like "Your portfolio holds £127k in US equities via VWRL and VUSA, giving you 58% unintentional geographic concentration — consider switching one to a global-ex-US fund." is 168 characters, fine. But a finding that includes a percentage breakdown sentence could easily hit 240. There is no warning to the report writer at authoring time — the finding just disappears from the PDF.

The 20-character lower bound will silently drop short but valid findings like "No ISA allowance used."

## Impact

A hand-written key finding disappears from the customer PDF with no indication to the report author that it was dropped. This directly undermines the product's value proposition.

## Fix

1. Raise the upper bound to at least 500 characters, or remove it entirely — these are editorial findings, not database fields.
2. Log at `ERROR` not `WARNING` when a finding is dropped so the report author notices.
3. Document the constraints in `docs/REPORT_WORKFLOW.md`.

## Acceptance criteria

- [ ] Upper bound raised to ≥500 chars or removed
- [ ] Dropped findings log at `ERROR` level
- [ ] `REPORT_WORKFLOW.md` documents any length constraints
- [ ] Test covers the boundary conditions

leonarduk:
### Motivation
- Key findings parsing previously dropped short lines and silently rejected entries outside an undocumented 20–240 character window. 
- This caused valid short findings to disappear and made long offending entries hard to notice. 
- The change targets clearer, less-surprising behavior when authors provide key findings files.

Closes #2614 

### Description
- Updated `backend/reports.py` to remove the 20-character minimum and enforce a single maximum length via `max_length = 500`. 
- Oversized findings now emit an `ERROR` level log with an explicit message including the maximum and source, instead of a warning. 
- Updated `tests/test_reports.py` to assert that short findings are preserved, a 501-character finding is dropped, and the drop is logged at `ERROR` level. 
- Updated `docs/REPORT_WORKFLOW.md` to document the new constraint (max 500 characters, no minimum beyond non-empty text) and the logging behavior.

### Testing
- Ran the targeted test subset `pytest -q tests/test_reports.py -k key_findings`, which completed successfully with `7 passed, 44 deselected`.
- The modified unit tests validate acceptance of short findings, rejection of findings >500 characters, and the `ERROR`-level log message for dropped items.
- No other automated test suites were run as part of this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c99f1635ec8327a68c8d7a790f930c)

### DIFF
--- a/backend/reports.py
+++ b/backend/reports.py
@@ -771,6 +771,10 @@ def _parse_key_findings_text(content: str, *, source_name: str = "key findings")
             numbered = re.match(r"^([1-9][0-9]?)\.\s+(.*)$", value)
             if numbered:
                 value = numbered.group(2).strip()
+        if value in {"-", "*", "\u2022"}:
+            continue
+        if re.match(r"^([1-9][0-9]?)\.$", value):
+            continue
         if not value:
             continue
         if len(value) > max_length:
--- a/docs/REPORT_WORKFLOW.md
+++ b/docs/REPORT_WORKFLOW.md
@@ -80,8 +80,8 @@ EOF2
 Writing guidance:
 - Findings are **not auto-generated**.
 - Use one bullet per finding with concrete numbers.
-- Keep each finding short and specific (recommended 20-240 chars).
-- Invalid lines are skipped with warnings instead of failing report generation.
+- Keep each finding short and specific (maximum 500 chars; no minimum beyond non-empty text).
+- Invalid lines are skipped with error logs instead of failing report generation.
 
 ## 4) Generate and download the PDF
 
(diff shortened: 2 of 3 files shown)


leonarduk:
@codex fix issues raised in comments

PR #2624, opened 2026-03-29
Unify portfolio VaR output and remove duplicate risk alias

### LINKED ISSUE #2612: Bug: _build_portfolio_var_section uses wrong import alias in audit-report path
## Bug

In `backend/reports.py`, `_build_portfolio_var_section` has two code paths gated by `_is_audit_var_section()`. The audit-report path (the `if _is_audit_var_section(section):` branch) calls functions on the module-level `risk` alias:

```python
payload = risk.compute_portfolio_var(context.owner, confidence=confidence)
sharpe = risk.compute_sharpe_ratio(context.owner)
```

The non-audit path uses `risk_mod`:

```python
var_data = risk_mod.compute_portfolio_var(...)
sharpe_ratio = risk_mod.compute_sharpe_ratio(...)
```

At the top of the file, `risk` and `risk_mod` are both imported from `backend.common.risk` in separate `try/except` blocks:

```python
try:
    from backend.common import risk as risk_mod
except ModuleNotFoundError:
    risk_mod = None

try:
    from backend.common import risk
except ModuleNotFoundError:
    risk = None
```

This is redundant and potentially confusing, but not itself a bug. The actual bug is that the guard in the audit path checks `if risk is None: return []` — so it correctly guards against `None`. However, if `risk` is set but `risk_mod` is `None` (or vice versa) due to import ordering or test monkeypatching, the two paths could behave inconsistently.

More concretely: `_is_audit_var_section(section)` checks for `{"metric", "value", "units"}` but **not** `"confidence"`. The `audit-report` template's `risk-assessment` section uses exactly `{"metric", "value", "units"}`, so it takes the audit path. The `portfolio-var` standalone section uses `{"metric", "confidence", "horizon_days", "value", "units"}`, so it takes the non-audit path. This dual-path logic means a user-defined template that includes `confidence` in its column schema will silently get a different (non-audit) response shape even if it is logically equivalent.

## Impact

Medium — the audit-report path is the one that matters for Month 1. It is currently correct. The non-audit path is unused by any built-in template today. Risk of divergence if `PORTFOLIO_VAR_SECTION` is ever wired into a template.

## Fix

Consolidate `risk` and `risk_mod` into a single import alias. Remove the dead `_is_audit_var_section` branching and produce one consistent row shape regardless of which columns the section schema declares. The section schema's column list should filter/rename, not change the underlying data fetched.

## Acceptance criteria

- [ ] `risk` and `risk_mod` collapsed to a single `risk` alias in `reports.py`
- [ ] `_build_portfolio_var_section` has one data-fetch code path, not two
- [ ] Existing tests for the VaR section still pass
- [ ] No behaviour change to the audit-report template output

leonarduk:
### Motivation
- Fix inconsistent VaR section behavior and duplicate `risk`/`risk_mod` aliasing that made report output and tests fragile (Closes #2612). 
- Ensure built-in and audit-style report templates produce a single, deterministic VaR row shape.

### Description
- Consolidated optional risk import to a single `risk` alias and removed the unused `risk_mod` branching in `backend/reports.py`.
- Simplified `_build_portfolio_var_section` to always fetch via `risk.compute_portfolio_var` and emit rows with `metric`, `value`, and `units` (e.g. `VaR (95%)`, `VaR (99%)`, optional `Sharpe ratio`), and aligned `PORTFOLIO_VAR_SECTION` schema accordingly.
- Updated `_portfolio_snapshot` to call `portfolio_mod.build_owner_portfolio` directly instead of a duplicate alias.
- Adjusted tests to patch the consolidated `risk` and `portfolio_mod.build_owner_portfolio` paths and to assert the unified VaR output shape; modified `tests/test_reports.py`, `tests/test_reports_additional.py`, and `tests/test_reports_pdf.py` to match the new behavior.

### Testing
- Ran `pytest -q tests/test_reports.py tests/test_reports_additional.py tests/test_reports_pdf.py` and all tests passed: `94 passed`.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69c99efd644c832790a51a19cb83f339)

### DIFF
--- a/tests/test_reports_additional.py
+++ b/tests/test_reports_additional.py
@@ -1113,6 +1113,10 @@ def test_build_report_document_audit_template_dispatches_real_builders(monkeypat
         summary=lambda: summary,
         transactions=lambda: [],
         allocation=lambda: [],
+        owner_portfolio=lambda: {
+            "total_value_estimate_gbp": 200.0,
+            "accounts": [{"holdings": [{"ticker": "AAA"}, {"ticker": "BBB"}]}],
+        },
         portfolio=lambda: {
             "total_value_estimate_gbp": 200.0,
             "accounts": [{"holdings": [{"ticker": "AAA"}, {"ticker": "BBB"}]}],


leonarduk:
@codex fix issues raised in comments

PR #2794, opened 2026-05-03
fix: replace build-time S3 sync with COPY of pre-synced data in Dockerfile.lambda

### LINKED ISSUE #2795: bug: Deploy Infrastructure fails with "Unable to locate credentials" during Docker build
## Description

The `Deploy Infrastructure` workflow fails during the `cdk deploy` step when building the Lambda Docker image. The Dockerfile attempts to run `sync_data.sh` at image construction time, which calls `aws s3 sync` — but AWS credentials are not available inside the Docker build context (they only exist in the runner environment).

## Error

```
#11 0.569 fatal error: Unable to locate credentials
ERROR: failed to build: failed to solve: process "/bin/sh -c chmod +x scripts/sync_data.sh && if [ -n "$DATA_REPO" ] || [ -n "$DATA_BUCKET" ]; then ..." did not complete successfully: exit code: 1
```

Failing run: https://github.com/leonarduk/allotmint/actions/runs/25282420759/job/74121634180

## Root cause

`BackendLambdaStack` passes `DATA_BUCKET` as a Docker build arg. The `Dockerfile.lambda` condition `[ -n "$DATA_BUCKET" ]` evaluates to true, so `sync_data.sh` runs `aws s3 sync` — which fails because Docker build containers don't inherit the runner's AWS credentials.

A secondary issue: `pip install awscli` in the Dockerfile installs a version of `botocore` that conflicts with the `boto3` version pinned in `requirements.txt`, producing a dependency resolver warning.

## Expected behaviour

The Docker image builds successfully. Data is already synced to `data/` by the workflow's **Sync data from S3** step before `cdk deploy` runs, so the Dockerfile should simply `COPY data/` rather than re-syncing from S3.

## Fix

See PR #2794.

leonarduk:
## Summary

Closes #2795 
- The `Deploy Infrastructure` workflow was failing because `Dockerfile.lambda` ran `sync_data.sh` during Docker image build, which tried to pull from S3 using `awscli` — but AWS credentials are not available in the Docker build context (only in the runner environment).
- The workflow's **Sync data from S3** step already downloads `data/` to the runner before `cdk deploy` runs, so the Dockerfile can simply `COPY data/ /var/task/data/` instead.
- Removing `pip install awscli` from the Dockerfile also eliminates a `botocore`/`s3transfer` version conflict with `boto3` that appeared as a warning in the build log.
- Removes the `DATA_BUCKET`/`DATA_REPO`/`DATA_BRANCH` build args from `backend_lambda_stack.py` since they were only used to drive the now-removed sync step.

## Root cause

`BackendLambdaStack` passed `DATA_BUCKET` as a Docker build arg → `Dockerfile.lambda` condition `[ -n "$DATA_BUCKET" ]` was true → `sync_data.sh` ran `aws s3 sync` → **fatal error: Unable to locate credentials**.

## Test plan

- [ ] CI `Deploy Infrastructure` workflow completes without the credential error
- [ ] Lambda image contains the expected `data/` directory (same content as before, sourced from the S3 sync step that already runs in the runner)
- [ ] CDK tests (`pytest cdk/`) still pass — no tests reference `build_args`

### DIFF
--- a/backend/Dockerfile.lambda
+++ b/backend/Dockerfile.lambda
@@ -1,26 +1,15 @@
 FROM public.ecr.aws/lambda/python:3.12
 
-ARG DATA_BUCKET
-ARG DATA_REPO
-ARG DATA_BRANCH
-
-# Install application dependencies and sync prerequisites
+# Install application dependencies
 COPY backend/requirements.txt ./requirements.txt
 RUN pip install --no-cache-dir -r requirements.txt \
-    && yum install -y git \
-    && pip install awscli
+    && yum install -y git
 
-# Copy application code and sync script into the Lambda task root
+# Copy application code and pre-synced data into the Lambda task root
 COPY backend/ /var/task/
-COPY scripts/bash/sync_data.sh /var/task/scripts/sync_data.sh
+COPY data/ /var/task/data/
 
 WORKDIR /var/task
-RUN chmod +x scripts/sync_data.sh \
-    && if [ -n "$DATA_REPO" ] || [ -n "$DATA_BUCKET" ]; then \
-         DATA_BUCKET=$DATA_BUCKET DATA_REPO=$DATA_REPO DATA_BRANCH=$DATA_BRANCH scripts/sync_data.sh; \
-       else \
-         echo "No build-time data source configured; skipping sync_data.sh"; \
-       fi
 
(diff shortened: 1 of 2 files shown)


leonarduk:
Closes #2795

PR #2802, opened 2026-05-03
fix: derive default template selection synchronously to fix checked radio test

### LINKED ISSUE #2801: fix: Reports page test failing — "Performance summary" radio button not checked
## Summary

CI is failing on the frontend test suite after merging PR #2799. One test in `Reports.test.tsx` asserts that the "Performance summary" radio button is pre-selected (checked) after the page renders, but the element is found in the DOM unchecked.

## Failing test

**File:** `tests/unit/pages/Reports.test.tsx`  
**Test:** `Reports page > renders the reports catalog with template metadata`  
**CI run:** https://github.com/leonarduk/allotmint/actions/runs/25287288615/job/74133627580

```
Error: expect(element).toBeChecked()

Received element is not checked:
  <input
    aria-label="Select Performance summary template"
    class="mt-1 h-4 w-4 text-indigo-600 focus:ring-indigo-500"
    name="report-template"
    type="radio"
    value="performance-summary"
  />
 ❯ tests/unit/pages/Reports.test.tsx:137:7
```

## Likely causes

- The default selected template in the `Reports` page component may have changed (e.g. first template in the list is no longer `performance-summary`, or the initial state no longer pre-selects any template).
- A recent refactor altered how the default radio value is set or which template appears first.

## Acceptance criteria

- [ ] `tests/unit/pages/Reports.test.tsx` passes cleanly with `npm --prefix frontend run test -- --run`
- [ ] The fix either corrects the component default selection to match the test's expectation, or updates the test to reflect the correct intended default — whichever reflects the actual desired UX.
- [ ] `make lint` and `npm --prefix frontend run lint` pass clean.

leonarduk:
## Summary

- Replaces the post-render `useEffect` that set `selectedTemplateId` with a `useMemo` that derives the effective selection in the **same render cycle** as templates arriving from the API.
- The old approach caused a render where templates were visible but `selectedTemplateId` was still `null`; `findByRole` resolved at that instant, so `toBeChecked()` failed before the state update landed.
- The new approach never shows templates in an unselected state: if the user hasn't explicitly chosen a template, the memo falls back to the first builtin, then the first template, then `null`.

## Test plan

- [x] `npm --prefix frontend run test -- --run` — 79 files, 434 tests all pass (was 1 failing)

Closes #2801

### DIFF
--- a/frontend/src/pages/Reports.tsx
+++ b/frontend/src/pages/Reports.tsx
@@ -149,7 +149,7 @@ export default function Reports() {
   const [owner, setOwner] = useState("");
   const [start, setStart] = useState("");
   const [end, setEnd] = useState("");
-  const [selectedTemplateId, setSelectedTemplateId] = useState<string | null>(
+  const [explicitTemplateId, setExplicitTemplateId] = useState<string | null>(
     null,
   );
 
@@ -162,12 +162,14 @@ export default function Reports() {
       .finally(() => setOwnersLoaded(true));
   }, []);
 
-  useEffect(() => {
-    if (templates.length === 0 || selectedTemplateId) return;
-    const preferred =
-      templates.find((template) => template.builtin) ?? templates[0];
-    setSelectedTemplateId(preferred.template_id);
-  }, [templates, selectedTemplateId]);
+  const selectedTemplateId = useMemo(() => {
+    if (explicitTemplateId) return explicitTemplateId;
+    return (
+      templates.find((template) => template.builtin)?.template_id ??
+      templates[0]?.template_id ??
+      null
+    );
+  }, [explicitTemplateId, templates]);
 
   const selectedTemplate = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


leonarduk:
A second commit has been added to this branch to also fix #2803.

**Second commit: `fix: don't re-raise prime_latest_prices failure during ASGI lifespan startup`**

Root cause: `backend/bootstrap/startup.py:_warm_snapshot` had an explicit `raise` after logging the exception from `prime_latest_prices`. Price priming is a startup cache-warmup optimisation — when it fails (e.g. Lambda cold-start can't reach external price APIs like Stooq/Yahoo Finance), the `raise` propagated through the FastAPI lifespan context manager. Mangum treats a lifespan startup failure as fatal and returns HTTP 500 for every subsequent request, including the `/docs` smoke-test curl.

Fix: remove the `raise`. A regression test has been added that asserts `startup()` completes without raising when `prime_latest_prices` throws.

This PR now closes both #2801 and #2803.

PR #2819, opened 2026-05-04
Fix /prices/refresh event loop blocking

### LINKED ISSUE #2818: perf: refresh_prices blocks the asyncio event loop during network and file I/O
## Summary

The `/prices/refresh` route handlers (`refresh_prices_get` and `refresh_prices_post` in `backend/routes/portfolio.py`) call the synchronous helper `_do_refresh_prices()`, which in turn calls `backend.common.prices.refresh_prices()` directly from the async handler body. This blocks the uvicorn asyncio event loop for the full duration of the operation.

## Why this matters

`prices.refresh_prices()` (`backend/common/prices.py:225`) performs significant blocking work:

1. **External network I/O** — `load_live_prices()` fetches current prices from external market-data sources
2. **File I/O** — reads timeseries data via `_load_latest_prices()` and writes the resulting snapshot to `config.prices_json` via `Path.write_text()`
3. **In-memory cache and alert fan-out** — `refresh_snapshot_in_memory()` and `check_price_alerts()`

While the route handlers are declared `async`, calling a long-running synchronous function directly inside them ties up the single event loop thread. During a price refresh, all other in-flight requests (health checks, portfolio loads, etc.) are stalled until the sync call returns.

## Affected code

```python
# backend/routes/portfolio.py
def _do_refresh_prices() -> dict:
    log.info("Refreshing prices via /prices/refresh")
    result = prices.refresh_prices()        # <-- blocks the event loop
    return {"status": "ok", **result}

@router.get("/prices/refresh", operation_id="refresh_prices_get")
async def refresh_prices_get():
    return _do_refresh_prices()             # <-- called from async context

@router.post("/prices/refresh", operation_id="refresh_prices_post")
async def refresh_prices_post():
    return _do_refresh_prices()             # <-- called from async context
```

## Proposed fix

Offload the blocking call to a thread pool using `asyncio.to_thread` (Python 3.9+):

```python
import asyncio

async def _do_refresh_prices() -> dict:
    log.info("Refreshing prices via /prices/refresh")
    result = await asyncio.to_thread(prices.refresh_prices)
    return {"status": "ok", **result}
```

This runs `prices.refresh_prices()` in a worker thread, freeing the event loop to handle other requests during the refresh.

Alternatively, if the refresh is always triggered as a background/admin operation and latency is not a concern, a `BackgroundTasks` approach could be used to return immediately and run the refresh asynchronously.

## Notes

- This is a pre-existing issue: the original `@router.api_route` handler (removed in PR #2816) had the same blocking behaviour. PR #2816 did not introduce the regression, but it was a natural point to flag it.
- `asyncio.to_thread` is thread-safe here as long as `prices.refresh_prices()` does not mutate shared async primitives — the current implementation uses a plain `dict` cache (`_price_cache`) which would need a lock under concurrent refreshes, but that is a separate concern.

leonarduk:
## Summary
- move `/prices/refresh` off the async event loop with `asyncio.to_thread`
- await the shared refresh helper from both GET and POST handlers
- add a regression test proving the route helper uses `asyncio.to_thread`

## Testing
- `../allotmint/.venv/Scripts/python.exe -m pytest -o addopts= tests/backend/routes/test_portfolio_helpers.py -k refresh_prices -q`
- `../allotmint/.venv/Scripts/python.exe -m pytest -o addopts= tests/test_backend_api.py -k 'prices_refresh_get or prices_refresh_post or alerts_endpoint' -q`
- `../allotmint/.venv/Scripts/python.exe -m pytest -o addopts= tests/test_main.py::test_prices_refresh_get tests/test_main.py::test_prices_refresh_post tests/test_main.py::test_openapi_no_duplicate_operation_ids -q`

Closes #2818

### DIFF
--- a/backend/routes/portfolio.py
+++ b/backend/routes/portfolio.py
@@ -10,6 +10,7 @@ Owners / groups / portfolio endpoints (shared).
 
 from __future__ import annotations
 
+import asyncio
 import datetime as dt
 import inspect
 import logging
@@ -924,18 +925,17 @@ async def instrument_detail(slug: str, ticker: str):
     return {"prices": prices_list, "mini": series.get("mini", {}), "positions": positions_list}
 
 
-def _do_refresh_prices() -> dict:
-    # Shared by GET and POST handlers; see issue #2818 to make this non-blocking.
+async def _do_refresh_prices() -> dict:
     log.info("Refreshing prices via /prices/refresh")
-    result = prices.refresh_prices()
+    result = await asyncio.to_thread(prices.refresh_prices)
     return {"status": "ok", **result}
 
 
 @router.get("/prices/refresh", operation_id="refresh_prices_get")
 async def refresh_prices_get():
-    return _do_refresh_prices()
+    return await _do_refresh_prices()
 
 
 @router.post("/prices/refresh", operation_id="refresh_prices_post")
 async def refresh_prices_post():
-    return _do_refresh_prices()
+    return await _do_refresh_prices()
--- a/tests/backend/routes/test_portfolio_helpers.py
+++ b/tests/backend/routes/test_portfolio_helpers.py
@@ -1,3 +1,4 @@
+import asyncio
 import types
 from pathlib import Path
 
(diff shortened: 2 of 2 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #2818` present in PR body |
| Issue coverage | ✅ | All acceptance criteria met: `_do_refresh_prices` is now async, `asyncio.to_thread` offloads the blocking call, both route handlers await it, regression test added |
| Review comments | ✅ | No open inline review comments |
| CI | ✅ | All 10 checks passing (backend integration, unit tests, CodeQL, Snyk, GPT/Claude reviews, issue-ref check) |
| Logic correctness | ✅ | Diff exactly matches the proposed fix in #2818; return shape unchanged; no frontend contract breakage |

### Actions taken
- No code changes required — the PR was already correct and complete.

### Remaining blockers
None. Two advisory notes from the automated reviews are worth being aware of but are not blockers:
1. **`asyncio.to_thread` requires Python ≥ 3.9** — the issue itself calls this out and CI is green, so the runtime is confirmed compatible.
2. **`_price_cache` dict is not thread-safe under concurrent refreshes** — acknowledged as out-of-scope in the issue. Under Lambda single-instance deployment this is benign; if multi-worker uvicorn is ever used, a `threading.Lock` will be needed. The issue is self-documenting on this point.

### Verdict
✅ **READY TO MERGE** — minimal, correct fix that exactly satisfies the issue acceptance criteria, all CI green, no open comments or blocking reviews.

PR #2820, opened 2026-05-04
fix: harden startup lifecycle to prevent 500 on /docs after Lambda cold start

leonarduk:
## Summary

Closes #2808

The deploy pipeline was failing at the "Validate backend and frontend endpoints" step because the deployed Lambda returned HTTP 500 on `/docs`. This PR addresses two remaining gaps after #2809 fixed the duplicate `refresh_prices` operation ID:

- **Harden `_warm_snapshot`**: `update_latest_prices_from_snapshot` was not wrapped in a try/except. If it raised (e.g. due to a cold-start network error or stale data), the exception propagated through the ASGI lifespan context manager, causing every subsequent request — including `/docs` — to return 500.
- **Add `/docs` integration tests**: Three new tests call `/docs` through the full `TestClient` lifecycle (i.e. they exercise the real ASGI lifespan, not just the service in isolation) to verify the endpoint returns 200 under normal conditions and when startup helpers raise exceptions simulating Lambda cold-start failures.

### `backend/bootstrap/startup.py`
- Wrapped `instrument_api.update_latest_prices_from_snapshot(snapshot)` in a `try/except Exception` that logs and continues, matching the existing pattern for `prime_latest_prices`.

### `tests/test_app.py`
- `test_docs_endpoint_returns_200` — verifies `/docs` returns 200 under normal startup.
- `test_docs_returns_200_when_prime_latest_prices_fails` — verifies `/docs` returns 200 when `prime_latest_prices` raises during startup (Lambda cold-start network failure simulation).
- `test_docs_returns_200_when_update_latest_prices_fails` — verifies `/docs` returns 200 when `update_latest_prices_from_snapshot` raises during startup.

## Test plan
- [x] `python -m pytest tests/test_app.py` — 7 passed
- [x] `python -m pytest tests/backend/test_app_bootstrap.py tests/test_lambda_handler.py` — 7 passed
- [x] `python -m ruff check backend/bootstrap/startup.py tests/test_app.py` — clean

PR #2836, opened 2026-05-05
fix: load JWT_SECRET from Secrets Manager before app import to fix 500 on /docs

leonarduk:
## Root cause

`backend/auth.py` raises `RuntimeError("JWT_SECRET environment variable is required")` at **module import time** when `APP_ENV=aws` and `JWT_SECRET` is not set. The CDK stack never puts `JWT_SECRET` into the Lambda environment — it lives in the `allotmint/app` Secrets Manager secret, which the Lambda has permission to read but no code ever fetched.

Every Lambda cold-start therefore failed to import `backend.auth`, causing Mangum to return 500 for every request including `GET /docs`.

This is the actual root cause of the post-deploy health-check failure that has been blocking CI:
```
curl: (22) The requested URL returned error: 500
##[error]Process completed with exit code 22.
```

The previous fix (PR #2820) hardened the startup lifecycle but didn't address this import-time failure.

### `backend/bootstrap/aws_secrets.py` (new)
Reads the named Secrets Manager secret and injects `JWT_SECRET` and `GOOGLE_CLIENT_ID` into `os.environ`. Runs only when `APP_ENV` is `aws` or `production`. Existing env vars are never overwritten. All errors are logged and swallowed so the app can produce its own validation messages rather than an opaque crash.

### `backend/lambda_api/handler.py`
Calls `load_aws_secrets_to_env()` **before** `from backend.app import create_app`. That import chain triggers `backend.auth` which reads `JWT_SECRET` at module level — the secret must be in `os.environ` before that line executes.

### `cdk/stacks/backend_lambda_stack.py`
- Removes `"GOOGLE_CLIENT_ID": "${GOOGLE_CLIENT_ID}"` — this was a Python dict literal (not shell interpolation), so the Lambda was receiving the string `"${GOOGLE_CLIENT_ID}"` which silently overwrote the real value loaded from the secret.
- Adds `"APP_SECRET_NAME": secret_name` so `load_aws_secrets_to_env()` knows which secret to fetch without relying solely on the `"allotmint/app"` default.
- Fixes pre-existing `typing.Sequence` → `collections.abc.Sequence` lint error.

### `tests/test_aws_secrets.py` (new)
8 tests covering: no-op outside AWS, injection of missing vars, non-overwrite of existing vars, boto3 errors, invalid JSON, missing secret keys.

## Test plan
- [x] `pytest tests/test_aws_secrets.py tests/test_lambda_handler.py` — 10/10 pass
- [x] `ruff check` on all changed files — clean
- [ ] Deploy and verify `curl /docs` returns 200

Closes #2836

### DIFF
--- a/backend/bootstrap/aws_secrets.py
+++ b/backend/bootstrap/aws_secrets.py
@@ -1,0 +1,66 @@
+"""Load secrets from AWS Secrets Manager into environment variables.
+
+This module must be imported (and ``load_aws_secrets_to_env`` called) before
+any backend module that reads ``os.environ`` at import time, such as
+``backend.auth``.  The handler entry-point calls this first so that
+``JWT_SECRET`` and ``GOOGLE_CLIENT_ID`` are in ``os.environ`` before
+``backend.app`` is imported.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+
+logger = logging.getLogger(__name__)
+
+_ENV_MAPPING: dict[str, str] = {
+    "jwt_secret": "JWT_SECRET",
+    "google_client_id": "GOOGLE_CLIENT_ID",
+}
+
+
+def load_aws_secrets_to_env(
+    secret_name: str | None = None,
+) -> None:
+    """Fetch the named Secrets Manager secret and inject values as env vars.
+
+    Only runs when ``APP_ENV`` is ``aws`` or ``production``.  Values that are
+    already present in ``os.environ`` are not overwritten so that explicit env
+    vars always win.
+
+    Missing or non-JSON secrets are logged and silently skipped; the app will
+    fail with its own validation errors later if required values are absent.
+    """
+
+    app_env = os.getenv("APP_ENV", "")
+    if app_env not in {"aws", "production"}:
+        return
+
+    if secret_name is None:
+        secret_name = os.getenv("APP_SECRET_NAME", "allotmint/app")
+
+    try:
+        import boto3  # type: ignore[import-untyped]
+
+        client = boto3.client("secretsmanager", region_name=os.getenv("APP_REGION") or os.getenv("AWS_REGION"))
+        response = client.get_secret_value(SecretId=secret_name)
+        secret_string = response.get("SecretString") or ""
+        secret_data: dict = json.loads(secret_string)
+    except ImportError:
+        logger.warning("boto3 not available; cannot load secrets from AWS Secrets Manager")
+        return
+    except json.JSONDecodeError as exc:
+        logger.error("Secret %s is not valid JSON: %s", secret_name, exc)
+        return
+    except Exception as exc:
+        logger.error("Failed to load secret %s from Secrets Manager: %s", secret_name, exc)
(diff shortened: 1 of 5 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ⚠️ | `Closes #2836` is self-referential (PR and issue share the same number). No separate issue exists. Acceptable for solo-maintainer flow. |
| Issue coverage | ✅ | Root cause (import-time `RuntimeError` in `backend.auth`) is addressed. CDK `${GOOGLE_CLIENT_ID}` literal bug fixed. `APP_SECRET_NAME` wired through CDK. Tests cover the main paths. |
| Review comments | ⚠️ | 3 open inline comments (see below). Owner dismissed both CodeQL alerts in the UI. |
| CI | ✅ | All checks passing (Snyk + 10 CI checks reported green). |
| Logic correctness | ✅ | Import ordering in `handler.py` is correct: `load_aws_secrets_to_env()` runs before `from backend.app import create_app`, which triggers `backend.auth`'s module-level `SECRET_KEY` resolution. |

### Open review comments

**CodeQL (×2) — "Clear-text logging of sensitive information"** (dismissed by owner)
These are false positives. The logger lines in `aws_secrets.py` log `secret_name` (the identifier, e.g. `"allotmint/app"`) and an exception message — not the secret value itself. No actual secret content reaches the log. The dismissals are justified.

**Codex P2 — Normalize `APP_ENV` before comparison**
The check `app_env not in {"aws", "production"}` is case-sensitive, while `backend/auth.py` normalises with `.lower()` before the same comparison. If `APP_ENV` is ever set to `"AWS"` or `"Production"` (e.g. by a CDK context override), `aws_secrets.py` would skip loading the secret but `backend.auth` would still raise `RuntimeError`, producing the same 500. The CDK stack currently hardcodes `"aws"` (lowercase), so this is low risk today, but worth a one-line fix:

```python
app_env = os.getenv("APP_ENV", "").lower()
```

### Actions taken
- No code changes committed (all remaining issues are low-risk or require human judgement).
- This summary comment posted.

### Remaining items before merge
1. **Deploy verification**: The test plan item `[ ] Deploy and verify curl /docs returns 200` is unchecked. This is the specific symptom the PR fixes; confirming it in staging/prod before merge is strongly recommended.
2. **Optional follow-up**: Add `app_env = os.getenv("APP_ENV", "").lower()` in `aws_secrets.py` to match `backend.auth`'s normalisation and close the Codex P2 gap. Could be done in this PR or a quick follow-up.
3. **Minor**: `secret_data: dict = json.loads(secret_string)` is not validated as a dict before the loop. A non-dict JSON value (extremely unlikely in practice) would raise `TypeError` outside the try/except. Low priority edge case.

### Verdict
**READY TO MERGE** (with the caveat that deploy verification is highly recommended first) — the core fix is logically correct, import ordering is sound, CI is green, the CDK literal-string bug is fixed, and all blocking items are resolved or acceptable.

PR #2845, opened 2026-05-06
fix: correct Dockerfile COPY and add Lambda timeout to fix /docs 500

### LINKED ISSUE #2842: fix: backend Lambda returns HTTP 500 on /docs post-deploy — build fails at health check
## Summary

The CI deploy job fails at the post-deploy validation step because `curl` to `${backend_url}/docs` returns **HTTP 500**, causing the workflow to exit with code 22. The smoke-test job is then skipped because the deploy step did not succeed.

## Evidence from logs (2026-05-05 run)

```
curl --fail --show-error --silent "${backend_url%/}/docs" >/dev/null
curl --fail --show-error --silent "https://${frontend_domain}" >/dev/null
...
curl: (22) The requested URL returned error: 500
##[error]Process completed with exit code 22.
```

Both CDK stacks (`BackendLambdaStack`, `StaticSiteStack`) deployed successfully before the curl check. The 500 originates from the Lambda handler — FastAPI's `/docs` endpoint should never return 500 unless the application itself fails to start or crashes on the first request.

## Likely root causes (investigate in order)

1. **Missing environment variable at Lambda runtime** — The recent fix in PR #2840 added `JWT_SECRET` and `GOOGLE_CLIENT_ID` to the *StaticSiteStack* deploy step, but the `BackendLambdaStack` CDK definition or its Lambda function environment may still be missing one or both of these at runtime. If the app reads them at import-time or startup, Lambda will 500 on cold start.
2. **Startup/import error in Lambda handler** — Any uncaught exception during module import or app factory setup will cause every request to return 500 via Mangum. Check CloudWatch Logs for the Lambda function for the exact traceback.
3. **Dependency mismatch in Docker image** — The Lambda Docker image installs `curl_cffi-0.15.0` while the CI test environment installs `curl_cffi-0.12.0`. A version-sensitive import could fail inside the container.

## How to diagnose

```bash
# Fetch CloudWatch logs for the Lambda function around the time of the failed deploy
aws logs tail /aws/lambda/<function-name> --since 1h --format short
```

Or check the Lambda test console for the `GET /docs` invocation error.

## Acceptance criteria

- `curl --fail ... /docs` returns HTTP 200 after CDK deploy in CI.
- Smoke-test job runs and passes.
- CloudWatch shows no startup errors for the Lambda function.

## Priority

**Critical** — this is the only step that causes the build to go red. All unit tests (79 frontend, 2115 backend) pass before this point.

leonarduk:
## Summary

- **Root cause**: `COPY backend/ /var/task/` in `Dockerfile.lambda` placed the package *contents* directly into `/var/task/` with no `backend/` subdirectory. The Lambda CMD `backend.lambda_api.handler.lambda_handler` requires `import backend` to succeed, but with no `/var/task/backend/` Python raises `ModuleNotFoundError` — every invocation returned HTTP 500 before the app ever ran.
- **Fix 1 (critical)**: Change to `COPY backend /var/task/backend` so the package is importable and all `from backend.X import Y` imports inside the handler work correctly.
- **Fix 2 (safety)**: Add `timeout=Duration.seconds(30)` to `BackendLambda`; the 3-second default is too short for a Docker Lambda cold start (FastAPI init + potential S3 fallback reads).
- **Fix 3 (guard)**: Add a CDK test asserting `BackendLambda` timeout ≥ 30 s to prevent silent regression.

## Test plan

- [x] All 18 CDK unit tests pass (`pytest cdk/tests/test_backend_lambda_stack.py`)
- [ ] After deploy, `curl --fail .../docs` returns HTTP 200
- [ ] CloudWatch shows no startup errors for BackendLambda

Closes #2842

### DIFF
--- a/backend/Dockerfile.lambda
+++ b/backend/Dockerfile.lambda
@@ -5,8 +5,9 @@ FROM public.ecr.aws/lambda/python:3.12
 COPY backend/requirements.txt ./requirements.txt
 RUN pip install --no-cache-dir -r requirements.txt
 
-# Copy application code and pre-synced data into the Lambda task root
-COPY backend/ /var/task/
+# Copy the backend package preserving its directory name so that
+# "backend.lambda_api.handler.lambda_handler" is importable from /var/task.
+COPY backend /var/task/backend
 COPY data/ /var/task/data/
 
 WORKDIR /var/task
--- a/backend/routes/support.py
+++ b/backend/routes/support.py
@@ -61,6 +61,8 @@ def _cache_is_fresh(cache: _PortfolioHealthSnapshot, threshold: float) -> bool:
 
 async def _compute_portfolio_health(threshold: float) -> _PortfolioHealthSnapshot:
     try:
+        # Deferred: scripts/ is outside the backend package and not present in the
+        # Lambda Docker image, so a module-level import would block Lambda startup.
         from scripts.check_portfolio_health import run_check  # noqa: PLC0415
         findings = await asyncio.to_thread(run_check, threshold)
     except Exception:  # pragma: no cover - defensive logging
--- a/backend/routes/support.py
+++ b/backend/routes/support.py
@@ -12,7 +12,6 @@ from fastapi import APIRouter, HTTPException
 from pydantic import BaseModel
 
 from backend.utils.telegram_utils import send_message
-from scripts.check_portfolio_health import run_check
 
 router = APIRouter(prefix="/support", tags=["support"])
 
@@ -62,6 +61,7 @@ def _cache_is_fresh(cache: _PortfolioHealthSnapshot, threshold: float) -> bool:
 
 async def _compute_portfolio_health(threshold: float) -> _PortfolioHealthSnapshot:
     try:
+        from scripts.check_portfolio_health import run_check  # noqa: PLC0415
         findings = await asyncio.to_thread(run_check, threshold)
     except Exception:  # pragma: no cover - defensive logging
         logger.exception(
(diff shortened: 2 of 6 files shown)


leonarduk:
Addressing the bot review concerns:

**"Dockerfile change absent from diff"** — the Dockerfile change *is* present in commit `85f7b8c5` (`backend/Dockerfile.lambda` is one of the 3 files in that commit). The review bots appear to have processed only the second commit. The actual change is:
```
-COPY backend/ /var/task/
+COPY backend /var/task/backend
```

**"JWT_SECRET / GOOGLE_CLIENT_ID missing"** — these are already in `backend_env` in the current `backend_lambda_stack.py` (added in prior work). The existing `test_backend_lambda_has_jwt_and_google_env_vars` CDK test verifies this and passes.

**Comment on deferred import** — added in the third commit (`effaa6d0`). The `# noqa: PLC0415` suppressor now has an inline explanation of *why* the import is deferred.

**Summary of all three root causes and their fixes:**

| Root cause | Fix |
|---|---|
| `COPY backend/ /var/task/` → no `backend` package → `ModuleNotFoundError` on all requests | `COPY backend /var/task/backend` (commit `85f7b8c5`) |
| `from scripts.check_portfolio_health import run_check` at module level → second `ModuleNotFoundError` on startup | Deferred import inside the one callsite (commit `dd2a9adb`) |
| Default 3-second Lambda timeout too short for cold start | `timeout=Duration.seconds(30)` (commit `85f7b8c5`) |

PR #2880, opened 2026-05-08
Check S3 meta timeseries cache existence using boto3 head_object

### LINKED ISSUE #2874: fix: has_cached_meta_timeseries always returns True for S3 paths
## Problem

`has_cached_meta_timeseries` in `backend/timeseries/cache.py` currently short-circuits for any `s3://` path:

```python
if cache.startswith("s3://"):
    return True  # S3 presence is checked lazily on read
```

This means callers that gate logic on this function (e.g. "only serve cached data if it exists") will always proceed for S3 paths, even when the object has never been written. The subsequent `pd.read_parquet(cache)` will then raise a `NoSuchKey` / `FileNotFoundError` rather than returning a clean "not found" result.

## Root cause

Introduced in PR #2872 as a pragmatic stub to avoid `Path.exists()` being called on an S3 URI. At the time, the S3 cache path was never reachable (env var wasn't set), so the behaviour was acceptable to unblock the 503 fix.

## Fix

Replace the stub with a real S3 existence check using `boto3`:

```python
import boto3

def has_cached_meta_timeseries(ticker: str, exchange: str) -> bool:
    cache = meta_timeseries_cache_path(ticker, exchange)
    if cache.startswith("s3://"):
        # e.g. s3://bucket/timeseries/meta/AAPL_US.parquet
        without_scheme = cache[len("s3://"):]
        bucket, _, key = without_scheme.partition("/")
        try:
            boto3.client("s3").head_object(Bucket=bucket, Key=key)
            return True
        except boto3.client("s3").exceptions.ClientError:
            return False
    p = Path(cache)
    return p.exists() and p.stat().st_size > 0
```

Or, if `pyarrow.fs` is preferred (already a dependency), use `pyarrow.fs.S3FileSystem` to stat the path.

## Acceptance criteria

- `has_cached_meta_timeseries` returns `False` for an S3 key that does not exist
- `has_cached_meta_timeseries` returns `True` for an S3 key that does exist
- Unit tests cover both cases with a mocked S3 client
- Existing local-path behaviour is unchanged

leonarduk:
### Motivation
- The existing `has_cached_meta_timeseries` returned `True` for any `s3://` cache path which caused callers to attempt parquet reads for S3 objects that might not exist.
- Implement an explicit S3 existence check so the cache layer can avoid unnecessary reads and behave consistently with local-path checks.

### Description
- Added `boto3` import and `from botocore.exceptions import ClientError` and implemented `_s3_cache_object_exists(cache: str) -> bool` which parses `s3://` URIs and calls `boto3.client('s3').head_object(...)` to verify existence, returning `False` on 404/NoSuchKey/NotFound and logging other errors.
- Updated `has_cached_meta_timeseries` to call `_s3_cache_object_exists` for `s3://` paths and preserved the existing local `Path(...).exists()` and `stat().st_size` behavior.
- Added unit tests in `tests/test_timeseries_cache_base.py` to cover an existing S3 key, a missing S3 key (404), and unchanged local-file behavior.

### Testing
- Ran `PYENV_VERSION=3.11.15 python -m pytest tests/test_timeseries_cache_base.py` and the test file passed (`6 passed`).
- Ran `PYENV_VERSION=3.11.15 ruff check --config backend/pyproject.toml backend/timeseries/cache.py tests/test_timeseries_cache_base.py` and the targeted checks passed.
- Ran `PYENV_VERSION=3.11.15 black --config backend/pyproject.toml` to format the changed files and then re-checked with `--check`, which passed for the modified files.
- `make lint` reported repo-wide lint failures and a local `pyenv` version mismatch that are pre-existing and outside the scope of this change (these did not block the targeted file checks or the new tests).

Closes #2874

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fe2fcf57288327a2d9aef2a26b5242)

### DIFF
--- a/backend/timeseries/cache.py
+++ b/backend/timeseries/cache.py
@@ -18,8 +18,10 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Callable, Dict
 
+import boto3
 import pandas as pd
 import requests
+from botocore.exceptions import ClientError
 
 from backend.common.instruments import get_instrument_meta
 from backend.config import config
@@ -446,7 +448,6 @@ def _convert_to_base_currency(
                 return pd.DataFrame()
             fx["Date"] = pd.to_datetime(fx["Date"])
 
-
         fx["Rate"] = pd.to_numeric(fx["Rate"], errors="coerce")
         return fx
 
@@ -520,10 +521,28 @@ def load_meta_timeseries_range(
     return _empty_ts()
 
 
+def _s3_cache_object_exists(cache: str) -> bool:
+    without_scheme = cache.removeprefix("s3://")
+    bucket, _, key = without_scheme.partition("/")
+    if not bucket or not key:
+        logger.warning("Invalid S3 timeseries cache path: %s", cache)
+        return False
+
+    try:
+        boto3.client("s3").head_object(Bucket=bucket, Key=key)
+    except ClientError as exc:
+        error_code = exc.response.get("Error", {}).get("Code")
+        if error_code in {"404", "NoSuchKey", "NotFound"}:
+            return False
+        logger.warning("Unable to check S3 timeseries cache object %s: %s", cache, exc)
+        return False
+    return True
+
+
 def has_cached_meta_timeseries(ticker: str, exchange: str) -> bool:
     cache = meta_timeseries_cache_path(ticker, exchange)
     if cache.startswith("s3://"):
-        return True  # S3 presence is checked lazily on read
+        return _s3_cache_object_exists(cache)
     p = Path(cache)
     return p.exists() and p.stat().st_size > 0
 
--- a/backend/timeseries/cache.py
+++ b/backend/timeseries/cache.py
@@ -11,7 +11,6 @@ where the parquet files live, e.g.
 
 from __future__ import annotations
 
-import importlib
 import logging
 import os
 from datetime import date, datetime, timedelta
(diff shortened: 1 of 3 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #2874` present in PR body |
| Issue coverage | ✅ | All 4 ACs met: False for missing key, True for existing key, mocked S3 unit tests, local-path behaviour preserved |
| Review comments | ✅ | 1 open Codex bot comment (P2: unhandled `BotoCoreError`) — fixed in `31739a4` |
| CI | ✅ | All checks passing (Snyk security: success) |
| Logic correctness | ✅ | Core logic correct; `BotoCoreError` gap fixed |

### Actions taken
- **`31739a4`** — `fix: catch BotoCoreError in _s3_cache_object_exists and log non-404 at error level`
  - Added `BotoCoreError` to the `from botocore.exceptions import ...` line
  - Added `except BotoCoreError` clause so credential/DNS/network failures return `False` gracefully instead of propagating an uncaught exception
  - Upgraded `logger.warning` → `logger.error` for non-404 `ClientError` (e.g. 403 Forbidden) so IAM misconfigurations are visibly alarming in logs
- **`bc754d8`** — `test: add BotoCoreError, non-404 ClientError, and invalid S3 path coverage`
  - `test_has_cached_meta_timeseries_returns_false_for_non_404_client_error` — 403 AccessDenied returns False
  - `test_has_cached_meta_timeseries_returns_false_for_botocore_error` — credential/network failure returns False
  - `test_s3_cache_object_exists_returns_false_for_invalid_path` — malformed `s3://` URIs return False

### Remaining blockers
- **No CDK/IAM change**: `_s3_cache_object_exists` requires `s3:HeadObject` on the cache bucket. The Lambda execution role must have this permission or every S3 path will silently return `False`. This is a deployment concern, not a blocker for merge if the role already has broad S3 access — but worth verifying. Tracked separately if needed.
- **`boto3.client("s3")` instantiated per call**: creates a new TCP connection on every `has_cached_meta_timeseries` call. Not a correctness issue but a performance one for hot paths. Out of scope for this PR; can be a follow-up.

### Verdict
**READY TO MERGE** — all correctness issues addressed. The two remaining items (IAM verification and client caching) are operational/performance concerns that don't block the fix.

leonarduk:
@codex address PR comments

PR #2881, opened 2026-05-08
Fix S3 timeseries cache invalidation

### LINKED ISSUE #2875: fix: S3-backed timeseries cache never invalidates within a warm Lambda
## Problem

`load_meta_timeseries` in `backend/timeseries/cache.py` uses file mtime to detect when the on-disk parquet has been updated and clear the in-process LRU cache (`_load_meta_timeseries_cached`). Since S3 objects have no mtime accessible without an API call, the current code stubs this out:

```python
if cache.startswith("s3://"):
    mtime = 0.0
```

`mtime` is always `0.0` for S3 paths, so `prev != mtime` is never `True` after the first call (both `prev` and `mtime` are `0.0`). The LRU cache therefore never clears for S3-backed data, meaning a warm Lambda will serve stale prices indefinitely until the process restarts — even after `PriceRefreshLambda` has written a fresh parquet file to S3.

## Root cause

Introduced in PR #2872 as a minimal fix to avoid calling `Path.stat()` on an S3 URI. The local-file invalidation strategy does not translate to S3.

## Fix

Use the S3 object's `LastModified` timestamp (from `head_object`) as the equivalent of mtime:

```python
if cache.startswith("s3://"):
    without_scheme = cache[len("s3://"):]
    bucket, _, key = without_scheme.partition("/")
    try:
        resp = boto3.client("s3").head_object(Bucket=bucket, Key=key)
        mtime = resp["LastModified"].timestamp()
    except Exception:
        mtime = 0.0
```

Alternatively, use the S3 `ETag` as a cache key instead of a float timestamp — this avoids clock-skew issues.

Consider also whether the LRU should have a TTL-based expiry (e.g. via `cachetools.TTLCache`) so that S3 changes are picked up within a bounded window even without an explicit invalidation signal.

## Acceptance criteria

- After `PriceRefreshLambda` writes a new parquet to S3, a subsequent call to `load_meta_timeseries` in a warm `BackendLambda` returns the updated data within a reasonable window (configurable TTL or on next request after the S3 object's `LastModified` changes)
- Unit tests cover the S3 invalidation path with a mocked `head_object` response
- Existing local mtime invalidation behaviour is unchanged

leonarduk:
### Motivation
- Warm Lambda / long-running processes were serving stale Meta timeseries from an in-memory LRU because S3-updated parquet files did not change the cached mtime used to invalidate the LRU.

### Description
- Added `_s3_object_mtime(cache: str) -> float` which uses `boto3.client('s3').head_object(...)["LastModified"]` to obtain an S3 object's timestamp for cache invalidation and logs/wraps failures defensively.
- Use `_s3_object_mtime` in `load_meta_timeseries` so that S3-backed `meta` parquet files trigger in-process LRU invalidation when their `LastModified` changes, while local files continue to use filesystem `stat().st_mtime`.
- Import `boto3` lazily via `importlib.import_module` and fall back to a `0.0` mtime on errors to preserve previous best-effort behavior.
- Added unit tests to `tests/test_timeseries_cache_base.py` that mock S3 `head_object` metadata changing and assert the LRU is cleared, and a test that verifies local file mtime invalidation remains unchanged.

### Testing
- Ran `PYENV_VERSION=3.11.15 python -m pytest tests/test_timeseries_cache_base.py` and the new/updated tests passed (`5 passed`).
- Ran `isort` / `black` / `ruff` on the touched files and formatting/lint checks for the changed files passed. 
- Attempted full `make lint`; it failed due to existing repository-wide lint issues unrelated to these changes, so only the focused checks and unit tests for this change were used for validation.

Closes #2875

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69fe2f92bc308327b6c0b0943392a115)

### DIFF
--- a/backend/timeseries/cache.py
+++ b/backend/timeseries/cache.py
@@ -310,6 +310,21 @@ def _s3_object_mtime(cache: str) -> float:
     return float(last_modified.timestamp())
 
 
+def _invalidate_meta_caches_if_stale(ticker: str, exchange: str) -> None:
+    """Clear both meta LRUs when the backing file's mtime has changed."""
+    cache = meta_timeseries_cache_path(ticker, exchange)
+    if cache.startswith("s3://"):
+        mtime = _s3_object_mtime(cache)
+    else:
+        p = Path(cache)
+        mtime = p.stat().st_mtime if p.exists() else 0.0
+    prev = _CACHE_FILE_MTIMES.get(cache)
+    if prev is not None and prev != mtime:
+        _load_meta_timeseries_cached.cache_clear()
+        _memoized_range_cached.cache_clear()
+    _CACHE_FILE_MTIMES[cache] = mtime
+
+
 @lru_cache(maxsize=512)
 def _load_meta_timeseries_cached(ticker: str, exchange: str, days: int) -> pd.DataFrame:
     """LRU-backed loader for Meta timeseries."""
@@ -332,19 +347,10 @@ def load_meta_timeseries(ticker: str, exchange: str, days: int) -> pd.DataFrame:
     if OFFLINE_MODE != config.offline_mode:
         OFFLINE_MODE = config.offline_mode
         _load_meta_timeseries_cached.cache_clear()
+        _memoized_range_cached.cache_clear()
         _CACHE_FILE_MTIMES.clear()
 
-    cache = meta_timeseries_cache_path(ticker, exchange)
-    if cache.startswith("s3://"):
-        mtime = _s3_object_mtime(cache)
-    else:
-        p = Path(cache)
-        mtime = p.stat().st_mtime if p.exists() else 0.0
-    prev = _CACHE_FILE_MTIMES.get(cache)
-    if prev is not None and prev != mtime:
-        _load_meta_timeseries_cached.cache_clear()
-    _CACHE_FILE_MTIMES[cache] = mtime
-
+    _invalidate_meta_caches_if_stale(ticker, exchange)
     return _load_meta_timeseries_cached(ticker, exchange, days).copy()
 
 
@@ -509,6 +515,7 @@ def load_meta_timeseries_range(
     base_currency: str = "GBP",
 ) -> pd.DataFrame:
     global OFFLINE_MODE
+    _invalidate_meta_caches_if_stale(ticker, exchange)
     for offset in range(0, 5):  # try same day, 1-day back, 2-day back...
         s = start_date - timedelta(days=offset)
         e = end_date - timedelta(days=offset)
--- a/backend/timeseries/cache.py
+++ b/backend/timeseries/cache.py
@@ -11,6 +11,7 @@ where the parquet files live, e.g.
 
 from __future__ import annotations
 
+import importlib
 import logging
 import os
 from datetime import date, datetime, timedelta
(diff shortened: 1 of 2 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness

| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | PR body contains `Closes #2875` |
| Issue coverage | ⚠️ | AC1/AC2/AC3 partially met — see actions taken below |
| Review comments | ⚠️ | Codex P1 (range LRU) — fixed in commit `7a2c02e` |
| CI | ✅ | All checks passing |
| Logic correctness | ✅ | Fixed — see below |

### Actions taken

**Bug fixes committed to this branch:**

1. **`7a2c02e` — fix: clear range LRU on S3 mtime change** (`backend/timeseries/cache.py`)
   - Added `_invalidate_meta_caches_if_stale(ticker, exchange)` which clears **both** `_load_meta_timeseries_cached` and `_memoized_range_cached` when the S3 `LastModified` changes.
   - Replaced the inline mtime block in `load_meta_timeseries` with a call to this helper.
   - Added call to `_invalidate_meta_caches_if_stale` at the top of `load_meta_timeseries_range` so that `get_price_for_date` / price lookups are also invalidated when the backing parquet changes.
   - Also clears `_memoized_range_cached` on offline-mode toggle (previously only `_load_meta_timeseries_cached` was cleared).

2. **`1d2e9d9` — test: verify `_memoized_range_cached` is cleared when S3 mtime changes** (`tests/test_timeseries_cache_base.py`)
   - Added `test_s3_range_cache_invalidates_when_last_modified_changes` which asserts that a repeated call to `load_meta_timeseries_range` with the same date range re-fetches data after the S3 `LastModified` changes.

**Follow-up issue created:**

3. **#2894 — fix: grant `s3:HeadObject` to Lambda timeseries cache IAM policy**
   - `_s3_object_mtime` calls `head_object`, but `_grant_timeseries_cache_access` only grants `s3:GetObject`. Without `s3:HeadObject` the call silently 403s and falls back to `0.0`, making the entire invalidation a no-op in production. This CDK change could not be included here because the PR branch predates PR #2882 which restructured the CDK stack — patching it here would produce a merge conflict or overwrite #2882's changes.

### Remaining blockers

- **#2894 must be merged before this PR's fix is effective in production.** Without `s3:HeadObject` granted to the Lambda roles, `_s3_object_mtime` will always return `0.0` silently, and no LRU invalidation will occur. The unit tests pass because they mock boto3, but the live Lambda will continue serving stale prices until the IAM permission is added.

### Verdict

**NOT READY TO MERGE** — the cache invalidation logic is correct after the commits above, but the fix is a no-op in production until [#2894](https://github.com/leonarduk/allotmint/issues/2894) is resolved and deployed. Merge #2894 first (or as a follow-on immediately after), then this PR is safe to merge.

---
*Reviewed by Claude PR Merge Checker. Advisory only.*

leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #2875` in PR body |
| Issue coverage | ✅ | All 3 ACs met: S3 LastModified invalidation, mocked unit tests, local mtime unchanged. TTL-based expiry was "consider whether" — not mandatory |
| Review comments | ✅ | Codex P1 (`_memoized_range_cached` not cleared) addressed in commit `1d2e9d99` — extracted `_invalidate_meta_caches_if_stale()` and called it from both `load_meta_timeseries` and `load_meta_timeseries_range` |
| CI | ✅ | Snyk passes on latest commit (`1d2e9d99`). No other required checks configured |
| Logic correctness | ✅ | Fix is correct: both LRUs are cleared before the cached lookup path in each public function |

### Actions taken
- Reviewed all commits including the additional fix commit `1d2e9d99` ("test: verify _memoized_range_cached is cleared when S3 mtime changes") — confirmed it also fixes `cache.py`, not just adds a test
- Verified no merge conflicts with `main` (the 4 commits ahead of the PR base all touch CDK/Lambda infrastructure, not `cache.py`)

### Remaining (non-blocking, out of scope for this PR)
- `boto3.client("s3")` is created fresh on every `_s3_object_mtime` call — one `head_object` round-trip per warm request per ticker. A TTL guard (e.g. only re-check every 60 s) would reduce cost and latency; tracked by the issue's "consider TTL" note
- `_CACHE_FILE_MTIMES` dict mutation is not thread-safe under concurrent async workers — pre-existing issue, not introduced here
- `# pragma: no cover` on the AWS exception path — the fallback silently serves stale data; worth a follow-up test in a future PR

### Verdict
**READY TO MERGE** — all acceptance criteria satisfied, the Codex P1 review concern is fixed, CI is green, and no merge conflicts exist.

Reviewed by Claude via pr-merge-checker skill.

PR #2944, opened 2026-05-11
fix: align @capacitor/core and @capacitor/ios to v8 to match android/cli

### LINKED ISSUE #2943: fix: @capacitor/core and @capacitor/ios pinned to v5 but @capacitor/android requires v8 — npm ci fails
## Bug

The deploy CI job fails at `npm ci` with an `ERESOLVE` peer dependency conflict:

```
npm error While resolving: @capacitor/android@8.3.3
npm error Found: @capacitor/core@5.7.8
npm error peer @capacitor/core@"^8.3.0" from @capacitor/android@8.3.3
```

## Root cause

`package.json` has a major-version mismatch across Capacitor packages:

| Package | Version |
|---|---|
| `@capacitor/cli` | `^8.3.3` |
| `@capacitor/android` | `^8.3.3` |
| `@capacitor/core` | `^5.7.0` ← outdated |
| `@capacitor/ios` | `^5.7.0` ← outdated |

`@capacitor/android@8.x` requires `@capacitor/core@^8.3.0`, but the project supplies v5.

## Fix

Bump `@capacitor/core` and `@capacitor/ios` from `^5.7.0` → `^8.3.0` to align with the already-upgraded `@capacitor/android` and `@capacitor/cli`.

## Evidence

Seen in CI run for tag `v0.14` — deploy job exits with code 1 before any deployment step runs, and the smoke-test job is skipped as a result.

leonarduk:
## Summary

- `@capacitor/android` and `@capacitor/cli` were already on `^8.3.3` but `@capacitor/core` and `@capacitor/ios` were still pinned to `^5.7.0`
- This caused an `ERESOLVE` peer-dependency conflict that made `npm ci` fail immediately in the deploy CI job, skipping the smoke-test job entirely
- Bumped `@capacitor/core` and `@capacitor/ios` to `^8.3.0` to align all four Capacitor packages on the same major version

## Test plan

- [ ] CI `npm ci` step should now resolve cleanly
- [ ] Deploy job should proceed past the install step
- [ ] Smoke-test job should run after a successful deploy

Closes #2943

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,8 +9,8 @@
       "version": "0.0.2",
       "dependencies": {
         "@capacitor/android": "^8.3.3",
-        "@capacitor/core": "^5.7.0",
-        "@capacitor/ios": "^5.7.0",
+        "@capacitor/core": "^8.3.0",
+        "@capacitor/ios": "^8.3.0",
         "yaml": "^2.8.4"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


leonarduk:
Addressing the two flagged concerns from the Claude AI review:

**1. `"peer": true` in the lockfile for `@capacitor/core`**

This is expected npm v7+ behaviour. When a package satisfies both a direct dependency (listed in `package.json` `dependencies`) *and* a peer dependency requirement of another installed package, npm records the single resolved entry with `"peer": true` as metadata about how it was resolved. It does **not** mean the package is omitted during `npm ci` — npm installs all entries in the lockfile regardless of that flag. This was verified by running `npm install` fresh after the fix; the lockfile produced is identical to the committed one.

**2. Missing `android/` and `ios/` native project updates for a v5→v8 migration**

Neither `android/` nor `ios/` directories exist in this repository — the native projects have never been scaffolded or committed. The `mobile:*` scripts in `package.json` are present for future use but no native build is part of any current CI workflow. There is therefore nothing to migrate. If native projects are added in the future, Capacitor v8 migration requirements (Gradle version, minSdkVersion, iOS deployment target) will need to be addressed at that point.

PR #3025, opened 2026-05-18
fix: remove non-ASCII em dashes from StaticSiteStack to fix silent cdk diff omission

### LINKED ISSUE #3014: chore: StaticSiteStack CDK diff silently omits one change due to mangled non-ASCII characters
## Problem

Every `cdk diff` run for `StaticSiteStack` prints:

```
Omitted 1 changes because they are likely mangled non-ASCII characters. Use --strict to print them.
```

This means one resource property in the `StaticSiteStack` CloudFormation template contains characters that CDK's diff renderer can't display safely, so it silently drops that diff entry.

## Steps to reproduce

```bash
npx cdk diff StaticSiteStack --all -c prod=true --strict
```

The `--strict` flag should print the suppressed change and reveal which property it affects.

## Likely cause

A non-ASCII character (em dash, smart quote, Unicode arrow, etc.) in a string property — most likely in a CSP header string, a cache policy comment, or a Cognito domain.

## Fix

1. Run `cdk diff --strict` to identify the affected property.
2. Replace the non-ASCII character with its ASCII equivalent.
3. Re-run `cdk diff` to confirm the warning is gone.

## Impact

Low severity — deploy still works. However, a silently-omitted diff entry could hide a real infrastructure change from review.

leonarduk:
## Summary

- Replaces three UTF-8 em dashes (U+2014 `—`) with ASCII hyphens (`-`) in `cdk/stacks/static_site_stack.py`
- The critical one is in the `CfnParameter` `description=` string, which CDK embeds directly in the synthesised CloudFormation template — that caused `cdk diff` to silently skip one entry with the warning `Omitted 1 changes because they are likely mangled non-ASCII characters`
- The other two were in Python comments and cleaned up for consistency

## Test plan

- [x] All 20 existing CDK assertion tests pass (`pytest cdk/tests/test_static_site_stack.py -v`)
- [ ] Run `npx cdk diff StaticSiteStack --all -c prod=true` after deploy and confirm the "Omitted N changes" warning no longer appears

Closes #3014

### DIFF
--- a/cdk/stacks/static_site_stack.py
+++ b/cdk/stacks/static_site_stack.py
@@ -44,7 +44,7 @@ class StaticSiteStack(Stack):
         ui_auth_removal_policy = _ui_auth_removal_policy(self)
 
         # BucketDeployment.Source.json_data() only accepts intra-stack tokens
-        # (Ref / Fn::GetAtt / Fn::Select) — Fn::ImportValue is explicitly rejected
+        # (Ref / Fn::GetAtt / Fn::Select) - Fn::ImportValue is explicitly rejected
         # by CDK's renderData validator. A CfnParameter produces a Ref token and
         # therefore works. The real URL is supplied at deploy time via
         # `--parameters StaticSiteStack:BackendApiUrl=<url>`.
@@ -60,7 +60,7 @@ class StaticSiteStack(Stack):
                 "HTTPS URL such as https://abc123.execute-api.us-east-1.amazonaws.com."
             ),
             description=(
-                "Backend API base URL — override at deploy time with the "
+                "Backend API base URL - override at deploy time with the "
                 "BackendLambdaStack BackendApiUrl output."
             ),
         )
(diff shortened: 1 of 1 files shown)


leonarduk:
### Clarification on the three changed lines (in response to code review)

The code review correctly noted that only one of the three changed lines is the actual bug fix:

**Line 60 — the real fix**
```python
description=(
    "Backend API base URL - override at deploy time with the "  # was: "—"
```
This string is a `CfnParameter` property. CDK serialises it verbatim into the synthesised CloudFormation template (`AWS::CloudFormation::Parameter` → `Description` field). That is the only non-ASCII string that ends up in the CFN template, and therefore the sole source of the `cdk diff` warning:
```
Omitted 1 changes because they are likely mangled non-ASCII characters.
```
Confirmation: a byte-scan of the file (`python -c "..."`) found exactly one non-ASCII sequence in a non-comment string — at the offset corresponding to this line.

**Lines 47 and 120 — cosmetic only**
The other two em dashes were in Python source comments (`# ...`). Python comments are stripped before CDK ever sees the code; they do not appear in the synthesised CloudFormation template and could not have caused the warning. They were cleaned up for consistency but have zero effect on CDK output.

The PR description's note that "The critical one is in the `CfnParameter` `description=` string" is accurate but understated — it is the *only* fix that matters for the issue.

PR #3029, opened 2026-05-18
fix: seed price snapshot and guard against empty overwrites

### LINKED ISSUE #2746: Fix missing data for positions
Some positions have missing data. Investigate and fix the root cause to ensure all positions display correctly in the portfolio view.

leonarduk:
## Summary

- **Root cause of #2746**: `data/prices/latest_prices.json` was missing (gitignored), leaving `_PRICE_SNAPSHOT` empty at startup — all non-cash positions returned `market_value_gbp=None`
- Add seed `data/prices/latest_prices.json` with demo-holding prices (VWRL.L, ERNS.L, PFE.N) and whitelist `data/prices/` in `.gitignore` so it is tracked
- Guard `refresh_prices()` and `refresh_snapshot_in_memory_from_timeseries()` against overwriting a valid snapshot when every freshly-fetched price is `None` (offline/test mode with no timeseries cache)
- Fix malformed/empty instrument metadata for `GSK.L`, `HFEL.L`, and `COMETF.L`

## Test plan

- [x] `test_enrich_holding_market_value_set_from_price_snapshot` — enrichment path uses snapshot price, `market_value_gbp` is non-None
- [x] `test_seed_prices_file_loads_demo_holdings` — seed file loads correctly and contains expected tickers
- [x] `test_refresh_prices_skips_write_when_all_prices_null` — all-null refresh does not overwrite a valid seed file
- [x] `test_refresh_prices_writes_json_and_updates_cache` — normal refresh still writes to disk
- [x] Seed file survives a full test-suite run (`data/prices/latest_prices.json` content unchanged after `pytest`)

Closes #2746

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,8 @@ data/*
 !data/accounts/**
 !data/instruments/
 !data/instruments/**
+!data/prices/
+!data/prices/**
 !data/scaling_overrides.json
 !data/events.json
 !data/trail.json
--- a/backend/common/portfolio_utils.py
+++ b/backend/common/portfolio_utils.py
@@ -1981,9 +1981,18 @@ def refresh_snapshot_in_memory_from_timeseries(days: int = 365) -> None:
 
     try:
         if _PRICES_PATH:
-            _PRICES_PATH.parent.mkdir(parents=True, exist_ok=True)
-            _PRICES_PATH.write_text(json.dumps(snapshot, indent=2))
-            logger.info("Wrote %d prices to %s", len(snapshot), _PRICES_PATH)
+            has_valid = any(
+                v.get("last_price") is not None for v in snapshot.values()
+            )
+            if has_valid:
+                _PRICES_PATH.parent.mkdir(parents=True, exist_ok=True)
+                _PRICES_PATH.write_text(json.dumps(snapshot, indent=2))
+                logger.info("Wrote %d prices to %s", len(snapshot), _PRICES_PATH)
+            else:
+                logger.info(
+                    "Skipping price snapshot write — no valid prices fetched"
+                    " (offline mode or no timeseries data)"
+                )
         else:
             logger.info(
                 "Price snapshot path not configured; skipping write (expected when config.prices_json is unset)"
--- a/backend/common/portfolio_utils.py
+++ b/backend/common/portfolio_utils.py
@@ -1980,23 +1980,24 @@ def refresh_snapshot_in_memory_from_timeseries(days: int = 365) -> None:
     refresh_snapshot_in_memory(snapshot, datetime.now(UTC))
 
     try:
-        if _PRICES_PATH:
-            has_valid = any(
-                v.get("last_price") is not None for v in snapshot.values()
-            )
-            if has_valid:
-                _PRICES_PATH.parent.mkdir(parents=True, exist_ok=True)
-                _PRICES_PATH.write_text(json.dumps(snapshot, indent=2))
-                logger.info("Wrote %d prices to %s", len(snapshot), _PRICES_PATH)
-            else:
-                logger.info(
-                    "Skipping price snapshot write — no valid prices fetched"
-                    " (offline mode or no timeseries data)"
-                )
-        else:
+        if _PRICES_PATH and snapshot:
+            existing: Dict[str, Any] = {}
+            if _PRICES_PATH.exists():
+                try:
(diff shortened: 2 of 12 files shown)


leonarduk:
All findings from the review were addressed in commit 679d6ffb. Summary of what changed:

**Critical bugs fixed:**
- **S3 upload now guarded** — `if config.app_env == "aws" and to_persist:` ensures Lambda instances are never served an all-null snapshot during a data-source outage.
- **Merge strategy replaces `any()` guard** — `refresh_prices()` now reads the existing snapshot, selects only entries with valid (`last_price is not None`) prices from the new fetch, and writes `{**existing, **to_persist}`. Partial-outage refreshes update the tickers that succeeded and preserve existing prices for tickers that returned `None`. Same approach applied to `refresh_snapshot_in_memory_from_timeseries()` in `portfolio_utils.py`.
- **Restored deleted `gain_gbp` assertion** — `assert row["gain_gbp"] == pytest.approx(0.25)` is back in `test_gbx_scaling_defaults_apply_without_override`.

**Other findings fixed:**
- `HFEL.json` sector corrected from `"Equity"` to `"Investment Trust"`.
- Seed file pretty-printed with `indent=2`.
- Dead `instrument_api` monkeypatch removed from `test_enrich_holding_market_value_set_from_price_snapshot`.
- New regression test `test_refresh_prices_partial_null_preserves_existing_prices` covers the partial-null merge case.

**Re: GSK/HFEL/COMETF absent from seed** — those instrument metadata files exist under `data/instruments/` but none of those tickers appear in any demo account file. The demo portfolios only hold `VWRL.L`, `ERNS.L`, `PFE.N` (plus `CASH.GBP`), so the seed already covers every non-cash demo position. The instrument JSON fixes were for data quality, not because those instruments are held in the demo portfolio.

**Re: `symbol` field in COMETF.json** — the original file used `"symbol": "COMETF"` which is not a key read by any backend code path (verified by grepping the backend for `instrument["symbol"]` and `get("symbol")`). The canonical key throughout the backend is `ticker`.

leonarduk:
Review findings addressed in commit cae48d67:

**`merged` scoping fixed** — moved the S3 `put_object` call inside the `if to_persist:` block so `merged` is always defined in scope when referenced. No logical change to behaviour, but the fragility is gone.

**`PFE.N price_currency: "GBP"` is correct** — the `prices.py` module docstring explicitly states all snapshot prices are GBP-normalised before storage: "load_live_prices returns GBP-normalised prices, so live snapshots emit price_currency = 'GBP'" and "_load_latest_prices also returns GBP-normalised prices". The seed file stores what a real refresh would store. The `price_currency` field records what denomination the `last_price` value is in (always GBP), not the instrument's native trading currency.

**Missing merge test for `portfolio_utils.py`** — added `test_refresh_snapshot_merges_with_existing_prices_file` to `tests/common/test_portfolio_utils_snapshot.py`, which verifies that a ticker with no timeseries data retains its existing seed price while a ticker with fresh data is updated.

leonarduk:
Review findings addressed in commit 9bf059f6:

**In-memory cache gap fixed** — `refresh_prices()` was calling `refresh_snapshot_in_memory(snapshot)` and populating `_price_cache` from `snapshot` even after merging. This meant a partial-outage refresh updated the disk correctly but left the running process with null prices in memory — defeating the seed preservation for any request served before the next cold start. Now:
- When `to_persist` is non-empty, both `_price_cache` and `_PRICE_SNAPSHOT` are updated from `merged` (which includes preserved seed prices for tickers that returned `None`)
- When `to_persist` is empty (all-null refresh), neither cache is cleared — the existing in-memory state is left intact

The test `test_refresh_prices_partial_null_preserves_existing_prices` is expanded to assert that both in-memory caches contain the preserved seed price for the null-returning ticker.

**`PFE.N price_currency: "GBP"` is correct** — all snapshot entries store GBP-normalised values. `price_currency` records the denomination of `last_price` in the snapshot (always GBP after normalisation), not the instrument's native trading currency. The `prices.py` module docstring states this explicitly: "load_live_prices returns GBP-normalised prices, so live snapshots emit price_currency = 'GBP'." The 22.0 GBP value is what a real refresh would store for PFE.N after USD→GBP conversion.

**`symbol`→`ticker` rename in COMETF.json** — verified safe. `get("symbol")` appears in two backend files: one reads from Yahoo Finance API response rows, the other from broker API responses. Neither reads from instrument JSON files. The canonical key throughout the instrument metadata layer is `ticker`.

PR #3057, opened 2026-05-19
fix: ensure frontend smoke tests run even when backend smoke fails

### LINKED ISSUE #3055: gap: frontend smoke tests (smoke:frontend) never run in CI — blocked by backend smoke test failure
## Problem

The `smoke-test` job defines a `Run frontend smoke tests` step:

```yaml
- name: Run frontend smoke tests
  run: npm --prefix frontend run smoke:frontend
```

However, this step never executes in practice because `Run backend smoke tests` (the preceding step) fails first — currently due to `SMOKE_URL` being empty (see linked issue). Even when the backend smoke test is fixed, the frontend smoke test step has **never been verified to pass** in a CI deployment run.

Observed in run: https://github.com/leonarduk/allotmint/actions/runs/26099581733/job/76748564310  
The log archive for this run contains no `9_Run frontend smoke tests.txt` file, confirming the step was never reached.

## Impact

- Frontend smoke tests provide zero coverage in CI post-deploy; regressions in the deployed frontend go undetected.
- Playwright browsers are installed (step 7, `123.7 KB` log) but never used.
- It is unknown whether `npm --prefix frontend run smoke:frontend` is correctly wired up and passes against the deployed environment.

## Fix

1. Fix the `SMOKE_URL` issue first (linked issue) so backend smoke tests pass.
2. Verify `npm --prefix frontend run smoke:frontend` exists in `frontend/package.json` and that it correctly reads `SMOKE_URL` (or a `PLAYWRIGHT_BASE_URL` equivalent) to target the deployed frontend.
3. Confirm the step passes in a full deploy run.
4. If the frontend smoke tests require auth (Cognito tokens), document or automate the credential provisioning needed in CI.

## Related

- Blocked by: `SMOKE_URL` not resolved from deploy job output (fix that issue first)

## Acceptance Criteria

- [ ] `Run frontend smoke tests` step appears in the CI log archive for a successful deploy run
- [ ] The step exits 0 against the deployed CloudFront/API Gateway environment
- [ ] Any credential/env-var requirements for the frontend smoke tests are documented or automated

leonarduk:
## Summary

- **`deploy-lambda.yml`**: The `smoke-test` job's frontend step now has `if: success() || failure()` so it always runs after deployment, even if the backend smoke step fails. `SMOKE_URL` is overridden to `http://localhost:5173` at step level (so Playwright targets the local Vite preview server, not the backend API), and `VITE_ALLOTMINT_API_BASE` is set to the backend URL so the built frontend app calls the real deployed API. Auth secrets (`SMOKE_TEST_ID_TOKEN`, `SMOKE_AUTH_TOKEN`) and a Playwright artifact upload step are added.
- **`opt-in-smoke-tests.yml`**: The frontend step condition is prefixed with `(success() || failure()) &&` so it runs when `SMOKE_TARGET=all` regardless of backend smoke outcome.

## Root cause

`SMOKE_URL` in the `smoke-test` job was the *backend* API URL (from `needs.deploy.outputs.backend_url`), but `smoke.spec.ts` uses `SMOKE_URL` as the *frontend* base URL. The two scripts use the variable for different purposes. The frontend step also had no `if:` guard, so any backend failure silently skipped it.

## Test plan

- [ ] Trigger a tag deploy and confirm "Run frontend smoke tests" appears in the CI log after the backend smoke step (pass or fail)
- [ ] Trigger `opt-in-smoke-tests` workflow with `smoke_target=all` on a branch where the backend smoke URL is unset; confirm the frontend step still runs

Closes #3055

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -441,7 +441,28 @@ jobs:
         run: npm run smoke:test
 
       - name: Run frontend smoke tests
+        # Run even if the backend smoke step failed so frontend results are
+        # always visible in CI logs (fixes issue #3055).  SMOKE_URL at job
+        # level is the backend API URL; override it to localhost so Playwright
+        # navigates to the local preview server, and expose the backend URL via
+        # VITE_ALLOTMINT_API_BASE so the built app calls the real deployed API.
+        if: success() || failure()
         run: npm --prefix frontend run smoke:frontend
+        env:
+          VITE_ALLOTMINT_API_BASE: ${{ env.SMOKE_URL }}
+          SMOKE_URL: 'http://localhost:5173'
+          TEST_ID_TOKEN: ${{ secrets.SMOKE_TEST_ID_TOKEN }}
+          SMOKE_AUTH_TOKEN: ${{ secrets.SMOKE_AUTH_TOKEN }}
+
+      - name: Upload Playwright deploy-smoke artifacts
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: playwright-deploy-smoke-artifacts
+          path: |
+            frontend/playwright-report/
+            frontend/test-results/
+          if-no-files-found: ignore
 
       - name: Fetch Lambda logs from smoke test
         if: always()
--- a/.github/workflows/opt-in-smoke-tests.yml
+++ b/.github/workflows/opt-in-smoke-tests.yml
@@ -100,9 +100,12 @@ jobs:
         # (success() || failure()) ensures this step runs even when the backend
         # smoke step failed, so frontend results are always visible in CI logs
         # regardless of backend health (fixes issue #3055).
+        # Note: the if: field evaluates as an expression context already;
+        # ${{ }} wrappers are intentionally omitted here so that the
+        # status-check functions (success, failure) are resolved correctly.
         if: >-
-          ${{ (success() || failure()) &&
-          (env.SMOKE_TARGET == 'all' || env.SMOKE_TARGET == 'frontend') }}
+          (success() || failure()) &&
+          (env.SMOKE_TARGET == 'all' || env.SMOKE_TARGET == 'frontend')
         run: npm --prefix frontend run smoke:frontend
         env:
           TEST_ID_TOKEN: ${{ secrets.SMOKE_TEST_ID_TOKEN }}
--- a/.github/workflows/opt-in-smoke-tests.yml
+++ b/.github/workflows/opt-in-smoke-tests.yml
@@ -97,7 +97,12 @@ jobs:
           TEST_ID_TOKEN: ${{ secrets.SMOKE_TEST_ID_TOKEN }}
 
       - name: Run frontend Playwright smoke tests
-        if: ${{ env.SMOKE_TARGET == 'all' || env.SMOKE_TARGET == 'frontend' }}
+        # (success() || failure()) ensures this step runs even when the backend
+        # smoke step failed, so frontend results are always visible in CI logs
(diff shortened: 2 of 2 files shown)


leonarduk:
Addressing the review's specific concerns with code citations.

---

### "No local server started before Playwright runs"

This is **not a gap**. Playwright starts the preview server automatically via the `webServer` config in [`frontend/playwright.config.ts`](frontend/playwright.config.ts):

```ts
webServer: {
  command: 'npm run preview -- --host 0.0.0.0 --port 5173',
  url: 'http://localhost:5173',
  timeout: 2 * 60 * 1000,
  reuseExistingServer: !process.env.CI,
},
```

When `playwright test` runs, it executes that command, waits up to 2 minutes for the server to respond at `http://localhost:5173`, then starts the tests. No manual `vite preview` step is needed — it is built into the test runner.

### "`smoke:frontend` script existence unverified"

It exists in [`frontend/package.json`](frontend/package.json):

```json
"smoke:frontend": "npm run build:preview && playwright install chromium && playwright test --config playwright.config.ts tests/smoke.spec.ts"
```

The script: (1) builds the frontend with `npm run build:preview`, (2) installs Chromium, (3) runs Playwright which starts the preview server via `webServer` config above.

### "`VITE_ALLOTMINT_API_BASE` is a build-time variable, not runtime"

Correct — and that is exactly why it works here. The `smoke:frontend` script runs `npm run build:preview` **at step execution time** (not during the CDK deploy), so `VITE_ALLOTMINT_API_BASE` set in the step `env` block is available to Vite and gets inlined into the bundle during that local build. The frontend app then calls the real deployed backend API.

### "Playwright base URL wiring unverified"

[`frontend/tests/smoke.spec.ts`](frontend/tests/smoke.spec.ts) line 3:

```ts
const baseUrl = process.env.SMOKE_URL ?? 'http://localhost:5173';
```

Setting `SMOKE_URL: 'http://localhost:5173'` at step level means `baseUrl` resolves to the local preview server — exactly where the `webServer` config starts it.

### "`upload-artifact@v7` version inconsistency"

`@v7` is already the version used in the existing [`opt-in-smoke-tests.yml`](https://github.com/leonarduk/allotmint/blob/main/.github/workflows/opt-in-smoke-tests.yml#L123). The new step is consistent.

### "Both `SMOKE_AUTH_TOKEN` and `TEST_ID_TOKEN` — are both needed?"

Both are passed because `smoke.spec.ts` reads either:

```ts
const authToken = process.env.SMOKE_AUTH_TOKEN ?? process.env.TEST_ID_TOKEN ?? null;
```

`SMOKE_AUTH_TOKEN` takes precedence; `TEST_ID_TOKEN` is the fallback. Both are passed so the test picks whichever secret is provisioned in the repository.

### `if: failure()` on artifact upload — scope

The artifact upload step has `if: failure()` which triggers if **any** prior step failed, including the backend smoke step. This is intentional — if the backend step fails it's useful to have Playwright artifacts to distinguish "frontend also failed" from "frontend passed despite backend issues". If it becomes noisy it can be tightened later to `steps.frontend-smoke.outcome == 'failure'`.

### `opt-in-smoke-tests.yml` missing `VITE_ALLOTMINT_API_BASE`

In the opt-in workflow, `SMOKE_URL` is the **frontend** deployment URL (set from `inputs.smoke_url` or `vars.SMOKE_URL`), not the backend URL. So there is no URL fix-up needed — `SMOKE_URL` is already the correct frontend target, and the app's API base is embedded at its own deploy time. The `VITE_ALLOTMINT_API_BASE` override is only needed in `deploy-lambda.yml` because there `SMOKE_URL` happens to be the backend URL.

PR #3068, opened 2026-05-19
fix: use data-testid selector for timeseries Load button in smoke test

### LINKED ISSUE #3065: test: timeseries smoke test "Load button not found" — button selector falls through to console.warn path
## Observation

The `timeseries edit resilience` smoke test in `frontend/tests/smoke.spec.ts` emits in CI:

```
Load button not found; skipping explicit click and asserting marker state.
```

The test tries to find a button with `role=button, name='Load'`, then falls back to `text=Load`, and if neither is found, skips the click with a warning. The marker assertion (`data-mode='timeseries'`) is still checked.

## Likely cause

The `vite preview` build renders the timeseries page differently from the `vite dev` build — possibly due to lazy loading, the backend returning an error (no auth, no real timeseries data), or the "Load" button being conditionally rendered only after some data is available.

## Impact

Low: the main resilience assertion (route marker stays visible after a failed edit load) is still checked. The click is only skipped when the button is absent. The test passes the marker check.

## Suggested fix

Investigate whether the "Load" button in the timeseries component uses a different element type (e.g. `<div onClick>` instead of `<button>`) in the preview build, or whether it is gated behind a state that isn't reached without auth/data. If confirmed, update the test selector or add the necessary setup fixture.

leonarduk:
## Summary

- Adds `data-testid="load-button"` to the Load `<button>` in `TimeseriesEdit.tsx` so it can be located by a stable, i18n-independent selector
- Replaces the fragile three-layer try/catch/fallback in the `timeseries edit resilience` smoke test with a single `page.getByTestId('load-button')` call
- The click is now always asserted (no more silent skip path), making the test a genuine check rather than a best-effort one

**Root cause:** In CI with `vite preview`, i18n may not be initialized at first render, so `t("timeseriesEdit.load")` returns the key string `"timeseriesEdit.load"` instead of `"Load"`. Both `getByRole('button', { name: 'Load' })` and `text=Load` then miss the button and fall through to `console.warn`.

## Test plan

- [ ] `npm --prefix frontend run test -- --run` — existing unit test for TimeseriesEdit uses `/load/i` regex so still passes
- [ ] `npm run smoke:test` against a `vite preview` build — the `timeseries edit resilience` test should now click the button and assert `requested === true`

Closes #3065

### DIFF
--- a/frontend/src/pages/TimeseriesEdit.tsx
+++ b/frontend/src/pages/TimeseriesEdit.tsx
@@ -408,7 +408,7 @@ export function TimeseriesEdit() {
             ))}
           </select>
         </label>{" "}
-        <button onClick={handleLoad} disabled={!ticker}>
+        <button data-testid="load-button" onClick={handleLoad} disabled={!ticker}>
           {t("timeseriesEdit.load")}
         </button>
       </div>
--- a/frontend/tests/smoke.spec.ts
+++ b/frontend/tests/smoke.spec.ts
@@ -410,23 +410,9 @@ test.describe('timeseries edit resilience', () => {
 
     await page.goto(target.href);
 
-    const loadButton = page.getByRole('button', { name: 'Load' });
-    // The 'Load' button may not always be rendered as a role=button in preview builds;
-    // try a fallback selector (by text) before failing the test.
-    try {
-      await expect(loadButton).toBeEnabled();
-      await loadButton.click();
-    } catch {
-      const fallback = page.locator("text=Load");
-      if (await fallback.count() > 0) {
-        await expect(fallback.first()).toBeVisible();
-        await fallback.first().click();
-      } else {
-        // If no clickable element is available, proceed to assert that the route marker
-        // remains visible as the main resilience check.
-        console.warn('Load button not found; skipping explicit click and asserting marker state.');
-      }
-    }
+    const loadButton = page.getByTestId('load-button');
+    await expect(loadButton).toBeEnabled();
+    await loadButton.click();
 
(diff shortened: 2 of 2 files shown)


leonarduk (commented):
Responding to the Claude AI review's request-changes verdict.

**On the `disabled={!ticker}` concern**

`await expect(loadButton).toBeEnabled()` uses Playwright's built-in auto-wait — it retries until the element becomes enabled or the timeout expires. The `useEffect` that reads `ticker` from the URL query string fires after mount and sets `ticker = "FAIL"`, which enables the button. Playwright will wait for that transition, so no explicit fixture is needed.

**On removing the fallback path being a "regression"**

The old fallback wasn't actually graceful. Trace the full test flow when the click was skipped:

1. `console.warn('Load button not found...')` — button click skipped
2. `await expect.poll(() => requested).toBeTruthy()` — **this would time out and fail**, because `requested` is only set to `true` by the intercepted API call, which only fires when `handleLoad()` is called, which only happens when the button is clicked

So the test was *already* failing in that environment — the `console.warn` path just delayed where it blew up, making the failure harder to diagnose. The marker assertion at the end was never the safety net the review assumed it to be.

**On the component not mounting (auth wall)**

The route interception of `**/timeseries/edit?*` and the `applyAuth` call both apply equally before and after this PR. If the component wasn't mounting before, the old code also would have failed (at `expect.poll`). Nothing in this diff changes auth behaviour.

**Conclusion**

The fix is correct. `data-testid` makes the selector i18n-independent, and the assertive `toBeEnabled()` + `click()` is the right shape for a test that claims to verify a click happens. Safe to merge.

PR #3089, opened 2026-05-22
fix(tests): eliminate act() warnings and guard fetch in InstrumentResearch tests

### LINKED ISSUE #3087: fix: InstrumentResearch tests emit React act() warnings and unhandled fetch errors
## Problem

Running the frontend test suite emits repeated warnings from `tests/unit/pages/InstrumentResearch.test.tsx`:

```
An update to InstrumentResearch inside a test was not wrapped in act(...).
Failed to fetch instrument metadata [TypeError: fetch failed]
```

The warnings appear for the test case **"hides navigation links when corresponding tab is disabled"** (and likely others in the same file). Tests currently pass but these warnings indicate:

1. Asynchronous state updates happening after the assertion phase — a sign that the test does not await component settlement correctly.
2. `fetch` calls are not mocked, causing `TypeError: fetch failed` noise in stderr.

## Impact

- Noisy CI output makes it harder to spot genuine regressions.
- The warnings are often a leading indicator of flaky tests (the state update may complete *after* the assertion in some environments, causing intermittent failures).

## Fix

- Wrap async interactions in `act(...)` / use `waitFor` / `findBy*` queries in the affected tests.
- Add a `vi.stubGlobal('fetch', ...)` mock (or use `msw`) so instrument-metadata fetches are handled cleanly rather than throwing.

## Acceptance criteria

- [ ] No `act(...)` warnings from `InstrumentResearch.test.tsx` in CI output
- [ ] No `TypeError: fetch failed` stderr noise from those tests
- [ ] All 20 tests in that file continue to pass

leonarduk:
## Summary

- Converts the only synchronous test (`"hides navigation links when corresponding tab is disabled"`) to async and awaits the page heading after render, so `listInstrumentMetadata` and `getNews` mock promises fully settle before the test exits — eliminating all `act()` warnings.
- Adds `vi.stubGlobal('fetch', ...)` in `beforeEach` (restored in `afterEach` via `vi.unstubAllGlobals()`) so any fetch call that escapes the `@/api` mocks in CI returns a clean empty response instead of `TypeError: fetch failed`.
- All 20 tests continue to pass.

## Test plan

- [ ] `npm --prefix frontend run test -- --run tests/unit/pages/InstrumentResearch.test.tsx` passes with 20/20 tests and no `act()` warnings in stderr
- [ ] No regressions: `npm --prefix frontend run test -- --run` passes across the full suite

Closes #3087

### DIFF
--- a/frontend/tests/unit/pages/InstrumentResearch.test.tsx
+++ b/frontend/tests/unit/pages/InstrumentResearch.test.tsx
@@ -89,18 +89,22 @@ function renderPage(config?: Partial<ConfigContextValue>) {
 }
 
 describe("InstrumentResearch page", () => {
-  afterEach(() => {
-    vi.unstubAllGlobals();
-  });
-
   beforeEach(() => {
+    // Safety net: return an empty-array JSON response for any fetch call that
+    // escapes the @/api mocks (e.g. in CI environments). All intentional API
+    // calls go through the vi.mock("@/api") module above and never reach the
+    // global fetch. Using mockImplementation (not mockResolvedValue) so that
+    // each intercepted call receives its own fresh Response instance — a
+    // single Response body can only be consumed once.
     vi.stubGlobal(
       "fetch",
-      vi.fn().mockResolvedValue(
-        new Response(JSON.stringify([]), {
-          status: 200,
-          headers: { "Content-Type": "application/json" },
-        }),
+      vi.fn().mockImplementation(() =>
+        Promise.resolve(
+          new Response(JSON.stringify([]), {
+            status: 200,
+            headers: { "Content-Type": "application/json" },
+          }),
+        ),
       ),
     );
 
@@ -212,6 +216,10 @@ describe("InstrumentResearch page", () => {
     ];
     mockListInstrumentMetadata.mockResolvedValue(catalogue);
     mockUpdateInstrumentMetadata.mockResolvedValue({} as any);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
   });
 
   it("renders overview summary and defers chart to timeseries tab", async () => {
--- a/frontend/tests/unit/pages/InstrumentResearch.test.tsx
+++ b/frontend/tests/unit/pages/InstrumentResearch.test.tsx
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 vi.mock("@/hooks/useInstrumentHistory", () => ({
   useInstrumentHistory: vi.fn(),
   getCachedInstrumentHistory: vi.fn(() => null),
@@ -89,7 +89,21 @@ function renderPage(config?: Partial<ConfigContextValue>) {
 }
 
 describe("InstrumentResearch page", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
   beforeEach(() => {
+    vi.stubGlobal(
+      "fetch",
+      vi.fn().mockResolvedValue(
+        new Response(JSON.stringify([]), {
+          status: 200,
+          headers: { "Content-Type": "application/json" },
+        }),
+      ),
+    );
+
     mockUseInstrumentHistory.mockReset();
     mockUseInstrumentHistory.mockReturnValue({
       data: {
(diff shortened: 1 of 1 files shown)


leonarduk:
## Response to code review

Thanks for the detailed review. Two of the three concerns were actionable — addressed in commit `3b4ed499`.

### AC1 — "only partially satisfied"

The review suggests other tests may also trigger `act()` warnings. I verified this empirically before opening the PR: running the suite before the fix produced warnings **only** under the `"hides navigation links when corresponding tab is disabled"` label:

```
stderr | … > hides navigation links when corresponding tab is disabled
Warning: An update to InstrumentResearch inside a test was not wrapped in act(...)
…(×6)
```

No other test in the file emitted a warning. This makes sense structurally — every other test is already `async` and either awaits `userEvent` interactions (which internally flush updates via `act`) or awaits `findBy*` queries (which poll inside `waitFor`/`act`). The one synchronous test was the only escape.

AC1 is fully satisfied, not partially. The issue description's "likely others" was speculative; the test output confirmed only one case.

### `findByRole("heading", { level: 1 })` fragility → fixed

This was a valid concern on its face. Even though the `h1` is genuinely absent on initial render (`displayName` is `null` until `listInstrumentMetadata` resolves, so the component renders a `<div>AAA.L</div>` fallback instead), a bare `level: 1` selector made that guarantee invisible.

**Fixed** in the latest commit by tightening to:

```ts
await screen.findByRole("heading", { level: 1, name: /Acme Corp/ });
```

…with an inline comment explaining the dependency:

> The h1 only appears once `listInstrumentMetadata` resolves and the "Acme Corp" name is set in state — so this selector is definitively post-fetch.

This makes the intent explicit and ensures the sentinel can't silently resolve early if the component structure changes.

### Mock response shape (`[]`)

`listInstrumentMetadata` returns `InstrumentMetadata[]` — a flat array. The component iterates it with `for (const entry of catalogue ?? [])`, so an empty array is the canonical "no results" shape and won't cause null-dereference. The stub is only a safety net for any fetch call that escapes the `@/api` mocks (e.g. in CI); the actual catalogue data comes from `mockListInstrumentMetadata.mockResolvedValue(catalogue)` in `beforeEach`.

All 20 tests pass with no `act()` warnings in stderr on the updated branch.

leonarduk:
## Response to second review round

Both actionable items are addressed in the latest commit (`f049e514`).

### 1. `Response` body reuse — fixed ✅

The `mockResolvedValue` returning a single shared `Response` instance was a real latent bug. Switched to `mockImplementation` so every intercepted call gets a fresh `Response`:

```ts
vi.stubGlobal(
  "fetch",
  vi.fn().mockImplementation(() =>
    Promise.resolve(
      new Response(JSON.stringify([]), {
        status: 200,
        headers: { "Content-Type": "application/json" },
      }),
    ),
  ),
);
```

### 2. Sentinel correctness — explained in the comment, not a bug

The review wonders whether `findByRole("heading", { level: 1, name: /Acme Corp/ })` resolves before or after the async work. The answer is: **after**, because the two data paths are completely independent:

- **`listInstrumentMetadata`** goes through `vi.mock("@/api")` → `mockListInstrumentMetadata.mockResolvedValue(catalogue)` → returns `[{ name: "Acme Corp", ... }, ...]`. This is what populates `metadata.name` and makes the `h1` appear.
- **`vi.stubGlobal('fetch', ...)`** is a safety net for any code that calls raw `fetch` directly (bypassing the `@/api` mock). Its `[]` response is never used by `listInstrumentMetadata`.

On initial render, `metadata.name = ""` and `detail.name = undefined` (the `useInstrumentHistory` mock in `beforeEach` doesn't include `name`), so `displayName = null` and the component renders `<div>AAA.L</div>` (not an `h1`). Only once the `listInstrumentMetadata` promise resolves and `setMetadata({ name: "Acme Corp" })` fires does the `h1` appear. The `findByRole` is therefore a genuine post-fetch gate.

The comment in the test now makes this relationship explicit:

```ts
// Wait for the h1 that contains the catalogue name: this element is absent
// on initial render (displayName is null until listInstrumentMetadata
// resolves) and only appears once the async state update has flushed,
// making it a reliable post-fetch sentinel.
await screen.findByRole("heading", { level: 1, name: /Acme Corp/ });
```

### 3. `afterEach` ordering — moved to follow `beforeEach` conventionally ✅

All 20 tests pass clean on the updated branch.

leonarduk:
The third review pass appears to have run against a stale diff — it is identical to the second review and raises the same two concerns that were already addressed in commit `f049e514` (see the [previous response](https://github.com/leonarduk/allotmint/pull/3089#issuecomment-4521624781)):

| Review concern | Status |
|---|---|
| `mockResolvedValue` shares a single `Response` instance | ✅ Fixed — switched to `mockImplementation` in `f049e514` |
| Sentinel correctness (`findByRole` might resolve before fetch) | ✅ Explained — `listInstrumentMetadata` uses `@/api` mock, not raw `fetch`; `h1` is absent on initial render |
| `afterEach` before `beforeEach` ordering | ✅ Fixed — reordered in `f049e514` |

No new changes required. All 20 tests pass on the current HEAD.

PR #3105, opened 2026-05-23
ci: add pre-deploy pip dry-run validation for backend/requirements.txt

### LINKED ISSUE #3103: ci: add pre-deploy dependency validation for backend/requirements.txt to catch Docker build failures before deploy
## Problem

Run #70436255587 (2026-05-23) demonstrated that a dependency conflict in `backend/requirements.txt` can pass all CI tests yet fail the Docker build — because tests run against the **root** `requirements.txt`, not the Lambda-target file. By the time the conflict is discovered, CDK synthesis has already completed and a deploy is in progress.

Tracked separately in #3100 (root vs backend requirements drift) and #3099 (the specific `typing_extensions` conflict that triggered this).

## Proposed pre-deploy checks

Add the following as CI steps that run **before** `cdk deploy`, gated so a failure blocks the deploy:

### 1. Pip dry-run validation
```yaml
- name: Validate backend/requirements.txt is resolvable
  run: pip install --dry-run -r backend/requirements.txt
```
Fast, no network writes needed (metadata-only). Catches `ResolutionImpossible` in seconds without building a Docker layer.

### 2. pip-compile lockfile (pip-tools)
Introduce `backend/requirements.in` as the source of truth and generate `backend/requirements.txt` via:
```bash
pip-compile backend/requirements.in -o backend/requirements.txt
```
Run `pip-compile --dry-run` in CI to verify the lockfile is up to date. Conflicts are caught at PR time rather than deploy time.

### 3. Run pytest against backend/requirements.txt
Add a matrix job (or replace the existing install step) that installs from `backend/requirements.txt` first, then layers dev-only extras on top:
```bash
pip install -r backend/requirements.txt -r requirements-dev.txt
pytest
```
This ensures the test environment matches the Lambda runtime environment.

## Acceptance criteria

- [ ] At least check 1 (dry-run) is implemented and blocks deploy on failure
- [ ] CI failure message clearly identifies which package conflict triggered the failure
- [ ] `backend/requirements.txt` duplicate entries (`numpy`, `pyarrow`) removed (done in fix for #3099)

leonarduk:
## Summary

- Adds `pip install --dry-run -r backend/requirements.txt` to `ci.yml` so every PR catches dependency conflicts at review time
- Adds the same step to `deploy-lambda.yml` immediately after pip setup, before any CDK diff/deploy steps, so the deploy is blocked on conflict

## Why

Tests run against the root `requirements.txt`, not `backend/requirements.txt`. A `ResolutionImpossible` error in the Lambda requirements file previously only surfaced during the Docker build mid-deploy (as seen in run #70436255587). This step costs a few seconds of metadata-only pip resolver time and surfaces the conflict with pip's native conflict message before any AWS work begins.

## Test plan

- [ ] Verify CI passes on this PR (no current conflict in `backend/requirements.txt`)
- [ ] Confirm the new step appears in both `ci.yml` and `deploy-lambda.yml` job logs
- [ ] To validate the gate works: temporarily introduce a version conflict in `backend/requirements.txt` and verify the step fails with a clear pip `ResolutionImpossible` message

Closes #3103


backend/requirements.txt
```diff
@@ -41,8 +41,6 @@ starlette~=0.49.1
 pydantic~=2.13.4
 requests>=2.32.4,<3
 peewee~=3.18.2
```
leonarduk, line 43: why are these deleted?

leonarduk:
Those two lines were **duplicate entries** — `numpy~=2.4.6` and `pyarrow~=21.0.0` both appeared twice in `backend/requirements.txt`: once in the first block (lines 31 and 35 of the original file) and again in the second block (lines 44–45). Removing the later pair leaves the single canonical pin in place; nothing is actually dropped.

This is AC3 from issue #3103: *"backend/requirements.txt duplicate entries (numpy, pyarrow) removed"*. pip tolerates duplicates silently, but they're confusing to read and can mask which version is authoritative if the two pins ever diverge.

You can confirm by checking the current file — `numpy` and `pyarrow` still appear once each.

PR #3109, opened 2026-05-23
ci: add pre-deploy dependency validation for backend/requirements.txt

### LINKED ISSUE #3103: ci: add pre-deploy dependency validation for backend/requirements.txt to catch Docker build failures before deploy

leonarduk:
## Summary

Implements #3103.

Dependency conflicts in `backend/requirements.txt` previously passed all CI tests (which install from the root `requirements.txt`) but caused failures at Docker build time during deployment. This PR adds two defence layers to catch such conflicts earlier.

### New required CI job — `validate-backend-deps`
A dedicated `validate-backend-deps` job runs `pip install --dry-run -r backend/requirements.txt` on every PR and push to `main`. Because it runs in a clean environment (fresh runner, no pre-installed packages), it faithfully reproduces the constraint set that Docker sees, so conflicts are surfaced at PR stage rather than mid-deploy.

The job is registered as a required status check in both `.github/rulesets/default-branch-protection.json` and `EXPECTED_REQUIRED_CHECKS` in `scripts/check_branch_protection_required_checks.py`, so the existing self-validating guard catches any future drift.

### Pre-deploy gate in `deploy-lambda.yml`
An early `Validate backend/requirements.txt is resolvable (pre-deploy)` step in the deploy workflow runs the same dry-run check before any deployment work begins. On failure it emits a `::error::` annotation that names `backend/requirements.txt` and directs operators to the pip output for the conflicting package names.

### Remove duplicate entries from `backend/requirements.txt`
`numpy` and `pyarrow` appeared twice in the file (once in the middle section, once at the bottom). The lower duplicate entries are removed.

## Testing

- `python scripts/check_branch_protection_required_checks.py` passes locally with the updated ruleset and check set.
- CI dry-run step output clearly names conflicting packages via standard pip error messages.

Closes #3103

---


leonarduk:
Thanks for the detailed review. Addressing each point:

**`actions/setup-python@v6`** — this version is already used in `backend-integration.yml`, `iac-validation.yml`, `siteplan.yml`, and the two other jobs in `ci.yml` itself. No change needed.

**AC3 (duplicate removal)** — the diff does include this: `backend/requirements.txt` removes `numpy~=2.4.6` and `pyarrow~=21.0.0` from the block below `peewee` (they appear twice in the file on `main`; the upper occurrence is retained as the canonical pin). `git diff origin/main...HEAD -- backend/requirements.txt` confirms 2 deletions.

**AC2 (missing `::error::` annotation on CI job)** — this was a genuine gap. Fixed in the follow-up commit [`9e1e1ce3`](https://github.com/leonarduk/allotmint/commit/9e1e1ce3): the `validate-backend-deps` CI step now wraps the dry-run in an `if !` block and emits the same `::error file=backend/requirements.txt::` annotation used in `deploy-lambda.yml`, so a conflict surfaces a clearly labelled annotation in the GitHub UI rather than raw pip output alone.

**Belt-and-suspenders (both CI job and deploy step)** — intentional by design. The deploy workflow can be triggered by pushing a tag directly, bypassing a PR entirely. The pre-deploy step closes that gap. Added a brief comment in the PR description to make the reasoning explicit.

PR #3113, opened 2026-05-23
ci: add lambda-compat pytest job against backend/requirements.txt

### LINKED ISSUE #3108: ci: run backend pytest suite against backend/requirements.txt (lambda-compat job)
## Problem

Follow-on from #3100 (addressed by #3107).

PR #3107 adds `pip install --dry-run` validation that catches `ResolutionImpossible` errors in `backend/requirements.txt` at PR time. However, it does **not** close the behavioural gap: the backend `pytest` suite still runs against the root `requirements.txt`, while the Lambda container uses `backend/requirements.txt`. A package that resolves successfully in both files but at different versions (e.g. pandas 2.x vs 3.x, protobuf 6.x vs 7.x) can cause a silent behavioural regression that tests would pass but the deployed Lambda would fail at runtime.

## Suggested approach

Add a dedicated `lambda-compat` CI job that:

1. Creates a clean venv
2. Installs `backend/requirements.txt` plus the test-only extras needed to run pytest (pytest, pytest-cov, pytest-asyncio, mypy, etc.) — either from `requirements-dev.txt` with conflicting pins resolved, or from a new `backend/requirements-test.txt` that lists only the test tooling
3. Runs the full backend pytest suite in that venv

The job should be added to `ci.yml` and registered as a required branch-protection check.

## Blocker to resolve first

The two requirements files currently have incompatible pins for several packages:

| Package | `requirements.txt` (root/CI) | `backend/requirements.txt` (Lambda) |
|---------|-------------------------------|--------------------------------------|
| pandas | `~=2.3.1` | `~=3.0.3` |
| protobuf | `>=6.31.1,<7` | `>=7.34.1,<8` |
| lxml | `~=6.1.0` | `~=5.3.0` |
| numpy | `~=2.4.4` | `~=2.4.6` |
| beautifulsoup4 | `~=4.13.4` | `~=4.14.3` |

These need to be reconciled (either by pinning `backend/requirements.txt` to the Lambda-correct versions throughout, or by aligning root `requirements.txt` to match the Lambda) before a combined install will succeed.

## Acceptance criteria

- A `lambda-compat` (or similar) CI job passes on every PR
- The job runs pytest against `backend/requirements.txt` in an isolated venv
- The job is a required branch-protection check
- `check_branch_protection_required_checks.py` is updated to match

leonarduk:
## Summary

Closes #3108

Adds a dedicated `lambda-compat` CI job that runs the full backend pytest suite against `backend/requirements.txt` — the same dependency set used by the Lambda container — rather than the root `requirements.txt`. This surfaces behavioural regressions where a package resolves at a different version between the two files (e.g. pandas 2.x vs 3.x, protobuf 6.x vs 7.x) that would pass CI today but fail at Lambda runtime.

## Changes

| File | What changed |
|---|---|
| `backend/requirements-test.txt` | **New file** — minimal test tooling (pytest, pytest-cov, pytest-asyncio) that has no pin conflicts with `backend/requirements.txt` |
| `.github/workflows/ci.yml` | New `lambda-compat` job: installs `backend/requirements.txt` + `backend/requirements-test.txt`, then runs `pytest` with `ALLOTMINT_SKIP_SNAPSHOT_WARM=true` |
| `.github/rulesets/default-branch-protection.json` | Registers `CI / Lambda-compat pytest (backend/requirements.txt)` as a required branch-protection check |
| `scripts/check_branch_protection_required_checks.py` | Adds the new check to `EXPECTED_REQUIRED_CHECKS`; validator passes locally |

## How it works

```yaml
lambda-compat:
  name: Lambda-compat pytest (backend/requirements.txt)
  runs-on: ubuntu-latest
  steps:
    - uses: actions/checkout@v6
    - uses: actions/setup-python@v6
      with:
        python-version: '3.12'
    - name: Install Lambda-compatible deps + test tooling
      run: |
        python -m pip install --upgrade pip
        pip install -r backend/requirements.txt -r backend/requirements-test.txt
    - name: Run backend pytest suite against Lambda deps
      env:
        ALLOTMINT_SKIP_SNAPSHOT_WARM: 'true'
      run: pytest
```

`backend/requirements-test.txt` contains only the test runners — no application-level pins — so there are no conflicts with `backend/requirements.txt`:

```
pytest~=9.0.3
pytest-cov~=7.1.0
pytest-asyncio~=1.3.0
```

## Notes

- The job is intentionally **separate** from the existing `validate-backend-deps` job (which checks `pip install --dry-run` conflicts) — they guard different failure modes.
- If the job fails after merge due to version-skew between the two requirements files, that is a signal to reconcile pins, not to disable the check.


leonarduk:
## PR Merge Checker — Summary

### Merge readiness

| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #3108` present in PR body |
| Issue coverage | ⚠️ | All structural ACs met; venv isolation gap fixed in 41e7c897; incompatible-pin blocker from issue not yet resolved (see below) |
| Review comments | ✅ | One open Codex inline comment (missing `jsonschema`) — addressed in commit `c0e3c03` by adding `jsonschema~=4.23.0` to `backend/requirements-test.txt` |
| CI | ❌ **BLOCKER** | `CI / Lambda-compat pytest (backend/requirements.txt)` is failing. The job is working as intended — it is surfacing real version-skew regressions between the root CI deps and the Lambda runtime deps. The job cannot pass until those regressions are fixed. |
| Logic correctness | ✅ | Job structure, check name alignment across all three files, and `pytest tests/` path are all correct |

### Actions taken

- **Commit `41e7c897`** — Added explicit `python -m venv .venv-lambda` step to the `lambda-compat` job, satisfying the "isolated venv" AC from issue #3108. Mirrors the pattern already used by `validate-backend-deps`.
- **Issue [#3140](https://github.com/leonarduk/allotmint/issues/3140)** — Created tracking issue for the pin reconciliation work needed to make the lambda-compat job pass.

### Remaining blocker

**CI failure requires pin reconciliation before this PR can merge.**

The `lambda-compat` job is failing because the test suite (written and passing against root `requirements.txt`) uses package APIs that have breaking changes in the Lambda-pinned versions:

| Package | Root (CI tests pass) | Lambda (`backend/requirements.txt`) |
|---------|----------------------|--------------------------------------|
| pandas | `~=2.3.1` | `~=3.0.3` |
| peewee | `~=3.18.2` | `~=4.0.6` |
| protobuf | `>=6.31.1,<7` | `>=7.35.0,<8` |

This is the intended purpose of the job — it correctly exposes that deployed Lambda functions may fail at runtime on code that passes CI. The fix is tracked in [#3140](https://github.com/leonarduk/allotmint/issues/3140): investigate which tests fail with the Lambda pins, then either align the tests to the Lambda-correct API or align the pins.

### Verdict

**NOT READY TO MERGE** — `CI / Lambda-compat pytest (backend/requirements.txt)` is failing due to version-skew regressions that must be resolved in #3140 before this required check can go green.

leonarduk:
**Update**: The CI failure has been diagnosed and fixed.

**Root cause** — `tests/test_timeseries_cache_merge.py` (2 tests) failed with:
```
AssertionError: Attribute "dtype" are different
[left]:  datetime64[s]   ← result from _rolling_cache fallback path
[right]: datetime64[ms]  ← expected from _load_parquet
```

In `_rolling_cache`'s exception and empty-fetch fallback handlers the code does `ex["Date"] = ex["Date"].dt.date` to compare against Python `date` objects, converting timestamps to native `date` objects. When `_ensure_schema` then calls `pd.to_datetime()` on those `date` objects:
- **pandas 2.x** (root `requirements.txt`): infers `datetime64[ns]`
- **pandas 3.x** (Lambda `backend/requirements.txt`): infers `datetime64[s]`

Meanwhile `expected` came from reading parquet via `_load_parquet`, where pyarrow stores and returns `datetime64[ms]`, so both pandas versions preserved `ms` and `_ensure_schema` left it unchanged.

**Fix (commit `3019dda8`)** — pin `_ensure_schema` to an explicit `.astype("datetime64[ms]")` after `pd.to_datetime`:
```python
df["Date"] = pd.to_datetime(df["Date"], errors="coerce").astype("datetime64[ms]")
```
This normalises the Date column to `ms` regardless of whether it arrived as Python `date` objects, `datetime64[ns]`, or `datetime64[s]`, matching the resolution pyarrow writes to parquet by default. The fix is robust across pandas 2.x and 3.x.

No other tests are affected — no test in the suite does an explicit `datetime64[ns]` dtype assertion on Date columns returned from `_ensure_schema`. The venv commit (`41e7c897`) from earlier is also on the branch.

CI should now go green on the lambda-compat job. Closing the tracking issue [#3140](https://github.com/leonarduk/allotmint/issues/3140) once confirmed.

leonarduk:
## Review response — all bot points addressed

The automated review has been raising two items as "hard blockers" that are not actually blocking issues. Clarifying here so reviewers can assess the true state of the PR:

### ❌ Bot claim: "`backend/requirements-test.txt` is missing"
This file **is present** — it was added in commit `c0e3c03` and exists at `backend/requirements-test.txt` in the branch. The bot appears to be reading from `main` rather than the PR branch when making this assertion.

### ❌ Bot claim: "unresolved pin conflicts between root and backend requirements"
There are **no conflicting pins** in this PR's CI setup. The `lambda-compat` job only ever installs `backend/requirements.txt + backend/requirements-test.txt` into an isolated venv — it never touches the root `requirements.txt`. The two files are never installed together, so their version ranges cannot conflict at runtime.

---

### ✅ Valid points from the review — now resolved

| Point | Commit |
|---|---|
| `datetime64[s]` vs `datetime64[ms]` mismatch in test failures | `3019dda8` |
| `tz_localize(None)` bug → must use `tz_convert(None)` | `45a8bafe` |
| Missing `datetime64[ms]` regression test | `9bcff194` |
| Add `logger.warning()` when tz-aware Date is encountered | `4e7f95fc` |
| Add non-UTC tz-aware test case (US/Eastern) | `4e7f95fc` |

The branch is now at `4e7f95fc`. CI should be green.

PR #3160, opened 2026-05-28
fix: block client-side request forgery in fetchJson (CodeQL #218)

### LINKED ISSUE #3153: security: fix client-side request forgery in frontend/src/api.ts (CodeQL alert #218)
## Summary

CodeQL reports one **error-severity `js/client-side-request-forgery` alert** in the frontend.

The URL of an `axios`/`fetch` request at [`frontend/src/api.ts:157`](https://github.com/leonarduk/allotmint/blob/main/frontend/src/api.ts#L157) is constructed from a user-provided value without validation.  
An attacker could craft a value that redirects the request to an arbitrary host, potentially leaking credentials or CSRF tokens.

## Location

| File | Line | Alert |
|------|------|-------|
| `frontend/src/api.ts` | 157 | [#218](https://github.com/leonarduk/allotmint/security/code-scanning/218) |

## Fix

Validate the URL before use — either:

1. **Allowlist approach** (preferred): check the constructed URL starts with the expected API base before making the request.

```typescript
const ALLOWED_BASE = import.meta.env.VITE_API_BASE_URL ?? "";

function assertSafeUrl(url: string): void {
  if (!url.startsWith(ALLOWED_BASE)) {
    throw new Error(`Blocked request to unexpected host: ${url}`);
  }
}
```

2. **Use a URL object** and verify the `origin` matches the known API origin before dispatching the request.

The fix should ensure no user-controlled string can redirect the outgoing request to an unintended host.

## References

- CodeQL rule: [`js/client-side-request-forgery`](https://codeql.github.com/codeql-query-help/javascript/js-client-side-request-forgery/)
- CWE-918: Server-Side Request Forgery (client-side variant)
- Code scanning alert: [#218](https://github.com/leonarduk/allotmint/security/code-scanning/218)

### LINKED ISSUE #3156: security: fix reflective XSS in timeseries_meta route (CodeQL alert #30)
## Summary

CodeQL reports one **error-severity `py/reflective-xss` alert** in the backend.  
A user-provided value is reflected in an HTTP response without being HTML-escaped, allowing an attacker to inject arbitrary HTML/JavaScript into the page.

## Location

| File | Line | Alert |
|------|------|-------|
| `backend/routes/timeseries_meta.py` | 97 | [#30](https://github.com/leonarduk/allotmint/security/code-scanning/30) |

## Fix

Never interpolate raw user input into an HTML response. Options depending on the route's intent:

1. **Return JSON** (preferred for API routes) — use `jsonify()` or a Pydantic model instead of rendering HTML with user data.

2. **Escape before rendering** — if HTML output is genuinely required, use `markupsafe.escape()`:

```python
from markupsafe import escape

return f"<p>{escape(user_value)}</p>", 200
```

3. **Use a template** — pass the value as a Jinja2 template variable (auto-escaped by default) rather than string-interpolating into an HTML string.

## References

- CodeQL rule: [`py/reflective-xss`](https://codeql.github.com/codeql-query-help/python/py-reflective-xss/)
- CWE-79: Improper Neutralization of Input During Web Page Generation (Cross-site Scripting)
- Code scanning alert: [#30](https://github.com/leonarduk/allotmint/security/code-scanning/30)

leonarduk:
## Summary

- Adds an origin guard inside `fetchJson` in `frontend/src/api.ts` that validates `fullUrl`'s origin matches the configured API base origin before every fetch call.
- Moves `resolveBase()` before the try/catch so the resolved base is available for both URL construction and the guard.
- Adds 4 new Vitest cases in `frontend/tests/unit/api.test.ts` covering blocked cross-origin URLs, allowed same-origin absolute URLs, allowed relative paths, and continued blocking after `setApiBase` changes the origin.

### Root cause

When the `url` argument to `fetchJson` is already an absolute URL, the original code skipped the base-resolution branch and used `url` as-is at line 157 with no opportunity to detect a cross-origin target. CodeQL flagged this as `js/client-side-request-forgery` (alert #218).

### Fix

```typescript
// Guard against client-side request forgery (CodeQL js/client-side-request-forgery, CWE-918).
if (new URL(fullUrl).origin !== new URL(resolvedBase).origin) {
  throw new Error(`Blocked request to unexpected host: ${new URL(fullUrl).origin}`);
}
```

Closes #3153
Closes #3156 

## Test plan

- [x] `npm --prefix frontend run test -- --run tests/unit/api.test.ts` — 15 tests pass (11 pre-existing + 4 new)
- [x] Full test suite: 469 tests pass; the 1 pre-existing failure in `main.apiBaseOverride.test.ts` is unrelated to this change


leonarduk (commented):
changes made

leonarduk:
## Round 3 review fixes — commit `94af7051`

Addressing the blocking concern and minor issues flagged in the third automated review.

### Blocking: regex scoped to user-input paths only

`_resolve_ticker_exchange` previously applied `_TICKER_SEGMENT_RE` after *all three* resolution branches, including the `_resolve_full_ticker` inference path (internal data). That could silently reject valid portfolio entries if the internal resolver ever returns segments outside `[A-Z0-9_-]`.

The regex now fires only inside the two **user-controlled** branches:
- `if exchange:` — user supplied both ticker and exchange
- `if "." in t:` — user supplied a dotted ticker like `ABC.L`

The `_resolve_full_ticker` branch (internal data) trusts the resolver's output and bypasses the regex entirely. `sanitise_log_value` still protects the log call in that branch.

New test `test_resolve_inferred_path_trusts_internal_data` pins the intentional asymmetry with an exchange code containing a `.` (outside the allowlist) to prove no false 400 is raised.

### Misleading comment fixed

The previous comment claimed "Yahoo Finance normalised identifiers use underscores". Yahoo Finance uses hyphens (`BRK-B`). The comment now accurately says underscores accommodate normalised internal identifiers and hyphens cover standard exchange codes.

### `api.ts` — TODO #3170 annotated

The origin-only guard comment now reads:
```ts
// Origin-only check; path-prefix validation is not enforced (TODO #3170).
```

### `api.test.ts` — empty-string `resolveBase()` test

Added `createClient` to imports and a new test:
```ts
it("throws a clear error when resolveBase() returns an empty string", async () => {
  const { fetchJson: testFetchJson } = createClient("");
  await expect(testFetchJson("/health")).rejects.toThrow(
    "API base is not a valid absolute URL",
  );
});
```
This uses `createClient` directly (since `setApiBase("")` is a no-op) to cover the `new URL("")` throw path independently of the `"not-a-valid-url"` case.

### Test counts
- Backend: 33/33 ✅
- Frontend: 470/470 ✅
- `ruff check` clean ✅
- `eslint` clean ✅

leonarduk:
## Round 4 review response — commit `343dfbde`

Addressing the two substantive items from the COMMENT verdict:

### Test coverage gap: `if "." in t` branch with bad input (now covered)

Added `test_resolve_rejects_unsafe_dotted_ticker_no_exchange` with three parametrized cases where `exchange=None` and the ticker contains a dot, making the `if "." in t` branch fire with user-controlled `sym`/`ex` values:

| Input | Bad segment |
|---|---|
| `"<script>.L"` | sym = `"<SCRIPT>"` |
| `"ABC."` | ex = `""` |
| `"ABC&foo=bar.L"` | sym = `"ABC&FOO=BAR"` |

All three raise `HTTPException(400)`. This was the only user-controlled branch without a direct bad-input test.

### Parametrize naming (fixed)

`bad_ticker, bad_exchange` → `ticker_input, exchange_input` in `test_resolve_rejects_unsafe_ticker_exchange`. The previous names implied both arguments were always malicious, but three of the four cases have `exchange="L"` (valid) with only the ticker being the payload.

### Items not actioned

- **TODO #3170**: issue #3170 already exists (created earlier in this PR chain).
- **Log message distinction**: the current three branches already emit distinct messages (`"Ticker resolved (provided exchange)"`, `"Ticker resolved (inferred from ticker)"`, `"Ticker resolved (inferred exchange)"`). No change needed.
- **Comment wording** around "internal resolver output": the current phrasing ("trust the resolver, no regex rejection") is accurate and concise; no change made.

### Test counts
- Backend: 36/36 ✅
- `ruff check` clean ✅

PR #3194, opened 2026-05-31
fix(deploy): add s3:ListBucket to deploy role and distinguish 403 from 404 in snapshot verify

### LINKED ISSUE #3191: fix(deploy): head-object 403 blocks deploy — add s3:ListBucket to deploy role and distinguish 403 from 404
## Problem

Every deploy of tag v0.16 fails at the "Warm price snapshot / verify" step with:

```
An error occurred (403) when calling the HeadObject operation: Forbidden
##[error]Process completed with exit code 255.
```

The retry loop (3 × 20 s) exhausts without success and the job exits with 255.

## Root cause

The CDK grants the GitHub deploy role `s3:GetObject` scoped to `prices/latest_prices.json` (see `cdk/stacks/backend_lambda_stack.py` ~line 590-594). However, **without `s3:ListBucket` the deploy role cannot determine whether a key is absent versus forbidden**: AWS returns HTTP 403 for a missing key when the caller lacks bucket-level `ListBucket` permission. The retry loop treats all non-zero exits as an IAM-propagation transient, so it never distinguishes "key not there yet" from "no permission at all" — the loop will always exhaust retries and fail.

There is also a secondary timing issue: the IAM policy grant is applied by `cdk deploy` in the same pipeline run; the current STS session does not automatically pick up inline-policy changes mid-session, so even if the key exists the grant may not be visible to the current workflow run.

## Evidence

From deploy log `logs_71561779839`:
- `BackendLambdaStack` CDK deploy reports `UPDATE_COMPLETE` at 23:03:42
- head-object attempt 1/3 fails at 23:03:48 (6 s after deploy)
- attempt 2/3 fails at 23:04:08 (+20 s)
- attempt 3/3 + extra 20 s wait fails at 23:04:50 (total ~68 s)
- Error: `403 Forbidden`

## Fix

1. **CDK (`cdk/stacks/backend_lambda_stack.py`)**: add `s3:ListBucket` (with `s3:prefix` condition `prices/*`) to the deploy-role grant alongside the existing `s3:GetObject` so that a missing key returns 404 instead of 403.

2. **Workflow (`deploy-lambda.yml`)**: in the retry loop, distinguish exit codes:
   - `403` → hard-fail immediately (no permission; retrying won't help)
   - `404` → retry (key not written yet; Lambda may still be running)
   - other non-zero → retry (transient)

3. **Optional**: increase max-wait from ~80 s to ~120 s (the Lambda cold-start + S3 write can exceed 80 s on a fresh container).

## Pre-deploy test

```bash
# Verify the deploy role's effective S3 permissions before pushing a tag:
aws iam simulate-principal-policy \
  --policy-source-arn <GITHUB_DEPLOY_ROLE_ARN> \
  --action-names s3:GetObject s3:ListBucket \
  --resource-arns \
    arn:aws:s3:::backendlambdastack-portfoliodatabucketb2b93eb9-pe1hjid72iul/prices/latest_prices.json \
    arn:aws:s3:::backendlambdastack-portfoliodatabucketb2b93eb9-pe1hjid72iul
```

Run `npx cdk diff BackendLambdaStack` locally (with AWS creds) to confirm the IAM diff includes the ListBucket statement.

leonarduk:
## Summary

- **CDK**: adds `s3:ListBucket` (conditioned on `prices/` prefix) to the GitHub deploy role alongside the existing `s3:GetObject` grant — without this, a missing key returns HTTP 403 instead of 404, making permission errors and write-lag indistinguishable
- **Workflow**: rewrites the `head-object` retry loop to fail immediately on 403 (permission error; retrying will never help) and retry on 404 (key not yet written); increases retry count from 3 to 5
- **Tests**: adds `test_github_deploy_role_has_s3_listbucket_on_prices_prefix` to cover the new CDK grant

## Root cause of the deploy failure

From log `logs_71561779839` (v0.16 tag deploy):
```
head-object attempt 1/3 failed (IAM propagation delay); retrying in 20s...
head-object attempt 2/3 failed (IAM propagation delay); retrying in 20s...
head-object failed on attempt 3/3; waiting 20s for IAM propagation before final check...
An error occurred (403) when calling the HeadObject operation: Forbidden
##[error]Process completed with exit code 255.
```

The existing grant (`s3:GetObject` on `prices/latest_prices.json`) is correct, but without `s3:ListBucket` AWS returns 403 for any key the caller cannot enumerate — whether that key exists or not. The retry loop treated every non-zero exit as an IAM-propagation transient, so it spun through all retries on what may have been a missing key or a genuine permission gap, always ending in 255.

## Test plan

- [ ] CDK unit tests pass: `cd cdk && python -m pytest tests/test_backend_lambda_stack_permissions.py -x -q`
- [ ] `cdk diff BackendLambdaStack` (with GITHUB_DEPLOY_ROLE_ARN set) shows the new `s3:ListBucket` policy statement in the diff
- [ ] Trigger a deploy: the `head-object` step should exit 0 on success, or print a clear `403 Forbidden — deploy role lacks permission` message and fail immediately on a real permission error

Closes #3191

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -225,8 +225,7 @@ jobs:
             if [ "$attempt" -eq 5 ]; then
               echo "head-object failed on attempt 5/5; giving up." >&2
               echo "$head_out" >&2
-              aws s3api head-object --bucket "$data_bucket" --key prices/latest_prices.json >/dev/null
-              break
+              exit 1
             fi
             echo "head-object attempt $attempt/5 failed (key not yet written); retrying in 20s..." >&2
             sleep 20
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -211,21 +211,24 @@ jobs:
           if not isinstance(payload, dict) or "snapshot" not in payload or "timestamp" not in payload:
               raise SystemExit(f"Unexpected PriceRefreshLambda response: {payload!r}")
           PY
-          # Retry with backoff: IAM policy changes are eventually consistent (~60s).
-          # On the first deploy after s3:GetObject is newly granted to the deploy role
-          # the role may get 403 for up to a minute even though CloudFormation has
-          # already reported UPDATE_COMPLETE. See issue #3166.
-          for attempt in 1 2 3; do
-            if aws s3api head-object --bucket "$data_bucket" --key prices/latest_prices.json >/dev/null 2>&1; then
-              break
+          # Retry with backoff: the Lambda write may lag the invocation response.
+          # With s3:ListBucket now granted to the deploy role (issue #3191), a
+          # missing key returns 404 (retryable) instead of 403 (hard failure).
+          # 403 means a genuine permission problem — retrying will never help.
+          for attempt in 1 2 3 4 5; do
+            head_out="$(aws s3api head-object --bucket "$data_bucket" --key prices/latest_prices.json 2>&1)" && break
+            if echo "$head_out" | grep -q "(403)"; then
+              echo "head-object returned 403 Forbidden — deploy role lacks s3:GetObject or s3:ListBucket on the data bucket. Check CDK grants for GITHUB_DEPLOY_ROLE_ARN. See issue #3191." >&2
+              echo "$head_out" >&2
+              exit 1
             fi
-            if [ "$attempt" -eq 3 ]; then
-              echo "head-object failed on attempt 3/3; waiting 20s for IAM propagation before final check..." >&2
-              sleep 20
+            if [ "$attempt" -eq 5 ]; then
+              echo "head-object failed on attempt 5/5; giving up." >&2
+              echo "$head_out" >&2
               aws s3api head-object --bucket "$data_bucket" --key prices/latest_prices.json >/dev/null
               break
             fi
-            echo "head-object attempt $attempt/3 failed (IAM propagation delay); retrying in 20s..." >&2
+            echo "head-object attempt $attempt/5 failed (key not yet written); retrying in 20s..." >&2
             sleep 20
           done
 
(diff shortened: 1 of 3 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | PR body contains `Closes #3191` |
| Issue coverage | ✅ | All three required ACs addressed: CDK `s3:ListBucket` grant, 403 hard-fail / 404 retry distinction, retry count increased |
| Review comments | ✅ | One open Codex P1 comment addressed — see below |
| CI | ✅ | 17 checks passing per PR page (GitHub API reports no commit statuses, but check runs all pass) |
| Logic correctness | ✅ (fixed) | Critical bug fixed in commit `466eddad` — see below |

### Actions taken

**Critical bug fixed** (commit `466eddad`): On attempt 5, the original code made a second `head-object` call and then unconditionally `break`-ed regardless of exit code — silently exiting 0 even when the key was still missing. This directly contradicted the PR's goal of surfacing failures clearly. Fixed by replacing the discarded call + unconditional `break` with `exit 1`.

**Codex P1 comment** (retry transient 403s on first deploy): This is a valid edge case — on the very first deploy after this PR merges, IAM propagation could cause a transient 403 before the new `s3:ListBucket` policy takes effect. However, the correct mitigation is already in place: the CDK deploy runs *before* the `head-object` step, and the `s3:GetObject` grant (which was already present before this PR) is what governs access. The new `s3:ListBucket` grant is additive; if it hasn't propagated yet, `head-object` will still return 403 — but that 403 now causes an immediate, clear failure message rather than a silent retry loop. The operator can re-trigger the deploy once IAM propagates (~60 s). Adding retry-on-403 would reintroduce the original ambiguity (can't tell "propagation lag" from "genuine permission gap") and is a design-level decision best tracked separately if desired.

### Remaining blockers
None.

### Verdict
**READY TO MERGE** — the critical exit-code bug has been fixed; all issue acceptance criteria are satisfied; CI is green.

PR #3205, opened 2026-05-31
fix(deploy): move changeset IAM grant to StaticSiteStack; enforce --method=changeset in diff

### LINKED ISSUE #3192: fix(deploy): cdk diff always falls back to template-only comparison — changeset IAM grants are applied after the diff step runs
## Problem

The `cdk diff` step in `deploy-lambda.yml` always emits:

```
Could not create a change set, will base the diff on template differences
```

This means diff output lacks replacement annotations (e.g. `[replace]` vs `[update]`) and can hide destructive changes.

## Root cause

The CDK stack grants the GitHub deploy role `cloudformation:CreateChangeSet / DescribeChangeSet / DeleteChangeSet` **inside `BackendLambdaStack.__init__`** (see `cdk/stacks/backend_lambda_stack.py` lines 565-581). These permissions are only applied when `cdk deploy BackendLambdaStack` runs. The `cdk diff` step in the workflow **runs before** the BackendLambdaStack deploy, so the grants are not in effect for the current diff invocation — they were applied in the previous deploy run.

The grants do persist to AWS after the previous deploy, but because CDK uses `add_to_principal_policy` on an imported role with `mutable=True`, the generated inline policy name is derived from the stack construct tree. If the policy content or construct path changes between releases the old policy is deleted and a new one created, causing a brief window where the permission is absent.

## Fix

Move the changeset IAM permissions out of CDK and into the base IAM role definition for `AWS_ROLE_TO_ASSUME` (managed outside of CDK), so they are always present regardless of whether a CDK deploy has run. Alternatively, add the role policy to a separate CDK stack (or the `StaticSiteStack`) that is deployed before the diff step, or grant changeset permissions to all stacks unconditionally in the base role.

## Pre-deploy test

Add an `iac-validation` step (or extend the existing `iac-validation.yml` workflow) that asserts `cdk diff --method=changeset` exits 0; a fallback to template-only diff should be treated as a lint failure, not silently accepted.

leonarduk:
## Summary

- Moves `cloudformation:CreateChangeSet/DescribeChangeSet/DeleteChangeSet` grant from `BackendLambdaStack` to `StaticSiteStack`, which is deployed first in the workflow — making the permissions stable across BackendLambdaStack structural changes that would otherwise briefly remove/recreate the inline policy
- Adds `--method=changeset` to the `cdk diff` command so a silent fallback to template-only diff causes a CI failure instead of producing misleading output
- Updates `test_backend_lambda_stack_permissions.py` to assert the grant is NOT in BackendLambdaStack
- Adds two new tests in `test_static_site_stack.py` asserting the grant IS in StaticSiteStack (with correct scope)

## Test plan

- [ ] `pytest cdk/tests/test_static_site_stack.py::test_github_deploy_role_gets_changeset_permissions_on_both_stacks` — passes
- [ ] `pytest cdk/tests/test_static_site_stack.py::test_no_changeset_grant_when_github_deploy_role_arn_absent` — passes
- [ ] `pytest cdk/tests/test_backend_lambda_stack_permissions.py::test_no_cfn_changeset_grant_in_backend_lambda_stack` — passes (pending local Jenkins EACCES env fix)
- [ ] Next deploy run: `cdk diff` should use changeset method without falling back

Closes #3192


leonarduk:
Closing this PR — all three changes it introduced have already landed on `main` through subsequent PRs:

- **`StaticSiteStack` changeset grant** — present in `main` at `cdk/stacks/static_site_stack.py` (added as part of this PR's commit that was merged separately).
- **`BackendLambdaStack` changeset grant removal** — done in #3207 ("move deploy-role IAM grants out of CDK into bootstrap script"), which also added the grant to `scripts/bash/bootstrap-deploy-role.sh`.
- **`--method=changeset` in `cdk diff`** — present in `main`'s `deploy-lambda.yml`.

After `git pull` the branch is now identical to `main` (empty diff). The reviewer's feedback was accurate: the PR's diff against the then-current `main` was misleading because the branch had diverged from a stale base. The fix was correct but arrived via a different route.

Issue #3192 is resolved in `main`.

PR #3286, opened 2026-06-03
fix: always post full Claude review regardless of verdict

### LINKED ISSUE #3285: Claude AI review missing reason when requesting changes
## Problem
When Claude requests changes on a PR, only a generic failure notice is posted, hiding the detailed review with the actual concerns Claude raised.

## Root Cause
The 'Post Claude review comment' step in claude-pr-review.yml only runs when the verdict is APPROVE (`if: success()`). When Claude requests changes, this step is skipped, and only a generic "Claude review failed" message is posted.

## Expected Behavior
The full Claude review with detailed reasoning should always be posted to the PR, regardless of whether the verdict is APPROVE or REQUEST CHANGES.

## Solution
Change the 'Post Claude review comment' step to use `if: always()` so it posts the complete review in all cases.

leonarduk:
## Summary
When Claude review verdict is REQUEST CHANGES, the full review details are not being posted to the PR. This causes users to only see a generic failure notice without understanding the specific concerns Claude raised.

## Changes
- Modified 'Post Claude review comment' step to use 'if: always()' so the full review is always posted, regardless of verdict
- Removed separate 'Post Claude failure notice' step since we're now always posting the full review

## Impact
Users will now see the complete Claude review with detailed reasoning for any changes requested, making it easier to understand and address the concerns.

Closes #3285

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -74,14 +74,25 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
-          {
-            echo "## Claude AI Code Review"
-            echo ""
-            cat /tmp/claude_review_body.md
-            echo ""
-            echo "---"
-            echo "*Reviewed by Claude via [claude-pr-review.yml](.github/workflows/claude-pr-review.yml). Advisory only.*"
-          } > /tmp/claude_comment_body.md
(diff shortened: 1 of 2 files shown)


.github/workflows/claude-pr-review.yml
```diff
@@ -84,15 +85,3 @@ jobs:
           } > /tmp/claude_comment_body.md
 
           gh pr comment "$PR_NUMBER" --body-file /tmp/claude_comment_body.md
-
-      - name: Post Claude failure notice
-        if: failure()
-        continue-on-error: true
```
leonarduk, line 90: @claude I need this still to fail if not APPROVED - will this still work with this code removed?

leonarduk (commented):
@claude address the comment

leonarduk (commented):
@claude do we need the same changes for the GPT reviewer?

PR #3694, opened 2026-06-08
fix(cdk): make canonical-domain redirect conditional on customDomain flag

### LINKED ISSUE #3693: CRITICAL: Production frontend unreachable — CloudFront redirects all traffic to app.allotmint.io which has no DNS record (NXDOMAIN)
# Production frontend is unreachable: CloudFront canonical-redirect Function sends every request to a non-existent domain (`app.allotmint.io` has no DNS record)

## What

The AllotMint production frontend appears to be **completely unreachable by any browser**. Every request to the CloudFront distribution gets 301-redirected to `https://app.allotmint.io/`, but that hostname **has no DNS record at all** (`NXDOMAIN`). This was discovered while investigating why [Deploy Infrastructure run 27104660131](https://github.com/leonarduk/allotmint/actions/runs/27104660131/job/79992630214) failed at the "Run Lighthouse CI" step with `CHROME_INTERSTITIAL_ERROR` — Lighthouse was not flaky; it correctly detected a real, deterministic outage.

(This supersedes #3692, which I opened first and then closed — that issue mischaracterized this as a CI-robustness problem before I traced it to the actual cause below.)

## Evidence

```
$ curl -sI https://d32n5ua14kb82t.cloudfront.net/        # = StaticSiteStack.DistributionDomain / LIGHTHOUSE_URL
HTTP/1.1 301 Moved Permanently
Server: CloudFront
Location: https://app.allotmint.io/
X-Cache: FunctionGeneratedResponse from cloudfront
...

$ curl -v https://app.allotmint.io/
* Could not resolve host: app.allotmint.io
curl: (6) Could not resolve host: app.allotmint.io
```

`X-Cache: FunctionGeneratedResponse from cloudfront` confirms the redirect is generated by a CloudFront Function (not an origin response). DNS resolution for `app.allotmint.io` fails outright (`NXDOMAIN` / "Could not resolve host") — not a timeout, not a propagation delay. **Every single request to the production CloudFront domain dead-ends here.** This is exactly what produces `CHROME_INTERSTITIAL_ERROR` / `chrome-error://chromewebdata` in headless Chrome (and presumably `DNS_PROBE_FINISHED_NXDOMAIN` in a normal browser): the browser follows the 301, fails to resolve the target host, and shows its internal network-error page.

## Root cause

`cdk/functions/viewer-request.js` is a CloudFront `viewer-request` Function, unconditionally attached to the distribution's `default_behavior` (`cdk/stacks/static_site_stack.py:236-241`). It hardcodes:

```js
var canonical = 'app.allotmint.io';
...
var redirectNeeded = proto !== 'https' || host !== canonical || targetUri !== uri;
if (redirectNeeded) {
    return { statusCode: 301, ..., headers: { location: { value: 'https://' + targetHost + targetUri + qs } } };
}
```

— i.e. it 301-redirects **any** request whose `Host` header isn't exactly `app.allotmint.io` to `https://app.allotmint.io<path>`.

However, the actual custom-domain provisioning (ACM certificate, CloudFront `domain_names` alias, Route53 alias record) is **conditionally** created only when a `customDomain` CDK context flag is truthy (`cdk/stacks/static_site_stack.py:181-223`):

```python
_custom_domain = "app.allotmint.io"
_enable_custom_domain = _is_truthy_context(self.node.try_get_context("customDomain"))
...
if _enable_custom_domain:
    # creates ACM cert, Route53 hosted-zone lookup, etc.
...
_distribution_extra: dict = (
    {"domain_names": [_custom_domain], "certificate": _certificate}
    if _enable_custom_domain
    else {}
)
...
if _enable_custom_domain and _hosted_zone is not None:
    # creates the Route53 A/AAAA alias record pointing app.allotmint.io -> distribution
```

**`customDomain` is never passed as a context value anywhere in the repo.** I grepped every workflow, script, and doc:
- `.github/workflows/deploy-lambda.yml` — only passes `-c prod=true` (lines 238, 247, 384)
- `scripts/deploy-to-AWS.ps1` — only passes `-c data_bucket=... -c prod=...` (lines 333, 357, 374)
- No other script/workflow references `customDomain` or `hostedZoneId`

So in the actually-deployed production stack:
- `_enable_custom_domain` = **False**
- No ACM certificate for `app.allotmint.io` exists
- The CloudFront distribution has **no alias** for `app.allotmint.io` (only the default `*.cloudfront.net` domain)
- **No Route53 record for `app.allotmint.io` was ever created**

...yet the unconditionally-attached `viewer-request.js` Function still hardcodes `app.allotmint.io` as the canonical host and redirects 100% of traffic there — to a domain that doesn't exist in DNS and that the distribution wouldn't even be able to serve via TLS/SNI even if it did resolve (no alias, no cert).

## How long has this been broken?

The canonical-redirect feature was added in commit `00e1c21a` ("feat: add CloudFront canonical redirect", **2025-09-04**, hardened in `11d9c3ca` the same day) and is on `main` today. The failing run shows `StaticSiteStack: (no changes)` for both `cdk deploy` invocations, meaning the distribution has been sitting in this exact (no custom domain, but with the unconditional dead-end redirect) configuration for some time — **this is not a one-off misconfiguration introduced by the v0.18.6 deploy; it is the steady-state of the production stack.**

This strongly suggests the production frontend has been unreachable via its CloudFront URL for a meaningful period — and this has gone unnoticed because:
1. **Lighthouse essentially never runs.** Checking the last 25 `Deploy Infrastructure` runs, this is the *only* one where the "Run Lighthouse CI" step actually executed — every other run shows it (and the preceding "Set LIGHTHOUSE_URL" step) as `skipped`, because `deploy` fails earlier in the pipeline for unrelated reasons (IAM preflight checks, "no successful ci.yml run", price-snapshot warm-up, etc.). The one audit that would have caught this immediately almost never gets to run.
2. **Smoke tests target the backend API directly** (`SMOKE_URL` = `BackendApiUrl`, not the frontend), so they wouldn't catch a broken frontend redirect.
3. Anyone manually visiting `app.allotmint.io` (e.g. from docs/bookmarks) would also hit `NXDOMAIN` directly, with no CloudFront involvement at all to leave a trace.

## Why it matters

If this analysis is correct, **the production frontend is currently inaccessible to end users** — this is a P0/critical availability issue, not a CI nuisance. Please verify independently from a normal browser/network (my sandbox's DNS may have its own restrictions, though `curl: (6) Could not resolve host` is a fairly unambiguous signal of a missing record, not a transient resolver hiccup — I cross-checked with both the local resolver and got consistent NXDOMAIN-shaped failures).

## How to fix

**Correction to my own first pass:** I originally wrote this up as "pick Option A or Option B — both are simple, just different end-states." That understated Option A badly enough that it's worth walking through explicitly, because someone could otherwise reasonably try it as a "quick fix" and hit a wall (or worse, half-apply it).

### Why "just enable customDomain" (my original Option A) is NOT a quick fix

`static_site_stack.py:193-198` has an explicit synth-time guard:
```python
if not Token.is_unresolved(self.region) and self.region != "us-east-1":
    raise ValueError(f"customDomain requires deployment to us-east-1 ...")
```
CloudFront requires ACM certificates for custom domains to live in `us-east-1`. But `cdk/app.py` instantiates `StaticSiteStack(app, "StaticSiteStack")` **without an `env=` override**, so it deploys to whatever region the CLI/credentials default to. The evidence shows that's **`eu-west-1`**: `BackendLambdaStack` exposes `BackendApiUrl = https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com`, and the workflow configures AWS credentials/region identically for both stacks via `secrets.AWS_REGION || vars.AWS_REGION`.

So simply passing `-c customDomain=true -c hostedZoneId=<Z...>` against the stack as currently deployed would **fail immediately at synth time** with that `ValueError`. Properly enabling the custom domain requires one of:
- **(A1) Relocating `StaticSiteStack` to `us-east-1`** — a new CloudFront distribution, new S3 bucket (full asset re-sync), and critically a **new Cognito User Pool** (user pools are regional and not portable — existing users/sessions need an explicit migration plan), plus a coordinated DNS cutover. This is a multi-day infrastructure migration with real downtime/data-migration risk — not a config flag.
- **(A2) Splitting the stack** — provision only the ACM certificate in a separate `us-east-1` stack (the standard CDK cross-region-cert pattern: a small cert-only stack whose ARN is exported/imported via SSM or cross-region references) while keeping the distribution, bucket, and Cognito pool in `eu-west-1`. Less disruptive than (A1), but still real CDK restructuring work — new stack, new cross-region references, new deploy ordering — not a flag flip.

**Either A-path is a substantial, separately-scoped infrastructure project. It should not be attempted as part of restoring availability — that would compound an outage with a risky migration under time pressure.**

### Recommended near-term fix: stop redirecting to a domain that isn't provisioned

This restores availability fast and with low risk:

1. Make the canonical-host check in `viewer-request.js` conditional on whether the custom domain is actually active for *this* deployment. **Important implementation nuance I got wrong in my first pass:** you cannot "inject the CloudFront `domain_name` as a build-time constant" — that value is allocated by AWS at deploy time (it's a CloudFormation-resolved token/`Fn::GetAtt`, not a Python literal), so it is **not known at CDK synth time** when the Function's source code is being assembled. The values that genuinely *are* known at synth time are `_enable_custom_domain` (bool) and `_custom_domain` (`"app.allotmint.io"`). So the fix has to be expressed in those terms:
   - When `_enable_custom_domain` is true: keep today's behavior — redirect any non-canonical `Host` to `https://app.allotmint.io<uri>`.
   - When `_enable_custom_domain` is false: **skip host-based canonicalization entirely** — there is only one domain the distribution actually serves (its default `*.cloudfront.net` domain), so there is nothing to redirect *to*. Whether to keep the function's HTTPS-enforcement/URI-normalization logic is a separate call — note `ViewerProtocolPolicy.REDIRECT_TO_HTTPS` (`static_site_stack.py:232`) likely already covers HTTP→HTTPS at the distribution level, so the function's own `proto !== 'https'` check may be partly redundant; verify with a synthesized-template diff before assuming either branch needs to be preserved.
   - Mechanically, the simplest CDK-idiomatic approach: read `viewer-request.js` as a template string in Python, substitute a placeholder (e.g. `__CANONICAL_HOST__` → `"app.allotmint.io"` or `""`/`null`) with the synth-time-known `_custom_domain`/`_enable_custom_domain` values, and pass the result to `cloudfront.FunctionCode.from_inline(...)` instead of `from_file(...)`. (CloudFront Functions KeyValueStore is another way to externalize config like this, but it adds a new resource + association to manage — only worth it if this needs to change without a redeploy, which doesn't appear to be the case here.)
2. **Simpler alternative**, if the URI-normalization half of the function isn't load-bearing: gate the entire `function_associations=[...]` attachment (`static_site_stack.py:236-241`) behind `_enable_custom_domain`. The canonicalizing redirect's whole purpose is to push traffic onto a canonical custom domain — without one, the function shouldn't be attached at all. **Caveat:** confirm first whether the URI-normalization/charset-validation behavior in `viewer-request.js` is relied on for anything else (malformed-path sanitization before requests reach the origin, etc.) — fully removing the function for non-custom-domain deployments would remove that too, which may or may not be acceptable.
3. Add/extend a CDK unit test in `cdk/tests/test_static_site_stack.py` (which already has some `customDomain`-related coverage — see `git log -S customDomain`) asserting that the synthesized CloudFront Function code does **not** contain a hardcoded external redirect target when `customDomain` is disabled. This is the regression test that would have caught the original bug before it ever reached production.

### Longer term (separate, deliberately-scoped work — do not bundle with the availability fix)

- If `app.allotmint.io` genuinely is meant to be the production domain, scope (A1)/(A2) above as their own migration issue with a real plan (region move or cert-stack split, Cognito migration strategy, DNS cutover window, rollback plan).
- Fix the CI pipeline so Lighthouse and other late-stage checks actually run on most deploys instead of being skipped on ~95% of them due to earlier, unrelated failures — that's *why* this sat undetected for ~9 months. (Originally explored in #3692 before I traced it to this root cause; worth a dedicated follow-up issue.)

## Where

- `cdk/functions/viewer-request.js` (hardcoded `canonical = 'app.allotmint.io'`; the file to template/parameterize)
- `cdk/stacks/static_site_stack.py:171-179` (unconditional `function_associations` attachment — candidate for gating), `:181-223` (conditional custom-domain provisioning — the source of the mismatch), `:193-198` (the `us-east-1` guard that blocks a quick `customDomain=true` enable)
- `cdk/app.py` (no `env=` passed to `StaticSiteStack` — confirms it inherits the `eu-west-1` deploy region; relevant context for why the longer-term Option A requires a region change/split, not a flag)
- `cdk/tests/test_static_site_stack.py` (add/extend the regression test described above)
- `.github/workflows/deploy-lambda.yml:247,384` and `scripts/deploy-to-AWS.ps1:333,357,374` (only relevant if/when the longer-term custom-domain migration is separately undertaken)

## Constraints

- Verify in a real browser/from a non-sandboxed network before treating this as confirmed-live — but don't delay the fix pending that, given how unambiguous the DNS evidence is (`NXDOMAIN`, not a timeout).
- The near-term fix must be tested against a synthesized template (per this repo's CDK testing conventions) to confirm the Function code's behavior actually matches the distribution's alias/cert state in both the `customDomain` enabled and disabled cases — don't just eyeball the Python or the JS in isolation.
- Do not attempt the `us-east-1` migration (A1/A2) as part of restoring availability — scope it separately, with its own plan and review, once the site is reachable again via the near-term fix.
- If/when the longer-term custom-domain migration is undertaken, the Route53 hosted zone ID must not be hardcoded into source — follow the existing pattern of passing it as deploy-time context/secret (see `static_site_stack.py:199-205`).
- Don't conflate this with #3692's original (incorrect) framing — this is an availability bug in deployed infrastructure, not a CI-step robustness problem.

## LLM Tier

**Opus**

This requires: (1) recognizing that the "obvious" fix (finish enabling the custom domain) is blocked by a region constraint and is actually a separate, much larger migration — easy to miss without reading `static_site_stack.py:193-198` *and* `cdk/app.py` together; (2) writing a CloudFront Function template/parameterization that correctly branches on synth-time-known values without assuming access to deploy-time-only values like the distribution's assigned domain name; (3) careful, minimal CDK changes to a security-relevant code path (the response-headers/CSP policy lives in the same stack) where mistakes are costly and a bad CloudFront Function can break the site further; (4) adding a regression test that actually encodes the lesson (function-vs-infrastructure drift) rather than just re-asserting today's behavior. This is exactly the kind of infra-correctness problem where a plausible-looking quick fix (region flag) is a trap, and getting that wrong burns real time during an active outage.

## Success

- A normal browser navigating to the production CloudFront domain (or to `app.allotmint.io`, whichever ends up canonical) successfully loads the AllotMint frontend — no redirect to a non-resolving host.
- `app.allotmint.io` either (a) resolves and serves the app via a properly-aliased, certificate-bearing CloudFront distribution, or (b) is no longer the redirect target of a distribution that doesn't serve it.
- `npx lhci autorun` against `LIGHTHOUSE_URL` completes a full audit run (no `CHROME_INTERSTITIAL_ERROR`) the next time the Lighthouse step actually executes.

## Failure modes to avoid

- Don't "fix" this by simply removing the Lighthouse step or making it `continue-on-error` (that was my earlier mistaken recommendation in #3692) — that would hide the very check that caught this outage.
- Don't enable `customDomain=true` without first confirming DNS/cert provisioning will succeed (wrong hosted zone ID, wrong region, or missing Route53 permissions could turn a "redirect to nowhere" outage into a "stack stuck mid-deploy" outage).
- Don't assume this is low-impact because "no one filed a complaint" — the lack of complaints is itself explained by the redirect making the app invisible to anyone arriving via the canonical URL, and by Lighthouse (the one check that would surface it) almost never running.
- Don't reach for the `us-east-1` migration as a "fix" for the outage — it solves a different problem (finishing the custom-domain feature) on a much longer timeline, and attempting it under outage pressure is how a one-system outage becomes a two-system outage.

leonarduk:
## Summary

- The CloudFront `viewer-request` Function unconditionally redirected **every** request to `https://app.allotmint.io`, regardless of whether that domain has an alias, ACM certificate, or DNS record on the deployed distribution. Since the production `StaticSiteStack` deploys *without* `customDomain` enabled (and `app.allotmint.io` resolves to NXDOMAIN), every viewer request was redirected straight into a dead domain — taking production fully offline.
- This implements the "near-term fix" from #3693: make the host-based canonical redirect conditional on whether a custom domain is actually configured for *this* deployment.
- Mechanically: `viewer-request.js` now contains a synth-time placeholder `__CANONICAL_HOST__`. `static_site_stack.py` reads the file as a template string, substitutes either the JSON-encoded custom domain (when `customDomain=true`) or the literal `null` (when it isn't), and loads the result via `cloudfront.FunctionCode.from_inline(...)` instead of `from_file(...)`. When `canonical` is `null`, the Function skips host-based canonicalization entirely (it still upgrades to HTTPS and normalizes the URI path).
- Added regression tests (`test_viewer_request_function_skips_redirect_target_without_custom_domain`, `test_viewer_request_function_redirects_to_custom_domain_when_enabled`) asserting the synthesized Function source neither contains the unsubstituted placeholder nor hardcodes `app.allotmint.io` when `customDomain` is disabled, and correctly canonicalizes to it when enabled.

Out of scope (per #3693's constraints): the `us-east-1` migration / enabling `customDomain=true` in production. This PR only stops the active outage by removing the redirect-to-nowhere; it does not attempt to stand up the custom domain.

Closes #3693

## Test plan
- [x] `python -m pytest cdk/tests/test_static_site_stack.py -k "viewer_request or custom_domain" -v` — 12 passed
- [x] Verified the pre-existing `make lint`/`black`/`ruff` findings on these files are unchanged from `main` (not introduced by this change)
- [ ] CI: full `cdk/tests` suite and lint checks

### DIFF
--- a/cdk/functions/viewer-request.js
+++ b/cdk/functions/viewer-request.js
@@ -34,10 +34,21 @@ function handler(event) {
     }
 
     var targetHost = host;
-    if (canonical) {
-        targetHost = host === canonical ? host : canonical;
-        if (!/^[A-Za-z0-9.-]+$/.test(targetHost)) {
+    if (canonical && host !== canonical) {
+        targetHost = canonical;
+    }
+    if (!/^[A-Za-z0-9.-]+$/.test(targetHost)) {
+        if (canonical) {
             targetHost = canonical;
+        } else {
+            // No canonical host is configured for this deployment and the
+            // incoming Host header fails the safety check, so there is no
+            // trustworthy value to build a redirect Location from. Returning
+            // it verbatim would be a host-header injection / open redirect
+            // (see issue #3693 review). Pass the request through unmodified;
+            // ViewerProtocolPolicy.REDIRECT_TO_HTTPS at the distribution
+            // level still enforces the HTTPS upgrade for this viewer.
+            return request;
         }
     }
     var targetUri = uri;
@@ -47,6 +58,12 @@ function handler(event) {
     if (targetUri.length > 1 && targetUri.endsWith('/')) {
         targetUri = targetUri.slice(0, -1);
     }
+    // The distribution's ViewerProtocolPolicy.REDIRECT_TO_HTTPS already forces
+    // an HTTPS upgrade, so `proto !== 'https'` is largely redundant here.
+    // It is intentionally kept as defense-in-depth (e.g. for the
+    // cloudfront-forwarded-proto header path, and in case the distribution
+    // policy ever changes) and folded into the same single 301 alongside
+    // host/URI canonicalization, avoiding a double redirect round-trip.
     var redirectNeeded =
         proto !== 'https' ||
         (canonical !== null && host !== canonical) ||
--- a/cdk/functions/viewer-request.js
+++ b/cdk/functions/viewer-request.js
@@ -8,7 +8,15 @@ function handler(event) {
         (protoHeader.value === 'http' || protoHeader.value === 'https')
             ? protoHeader.value
             : null;
-    var canonical = 'app.allotmint.io';
+    // __CANONICAL_HOST__ is substituted at CDK synth time (see
+    // static_site_stack.py): a JSON string holding the configured custom
+    // domain when this deployment has one (alias + ACM certificate + DNS
+    // record), or `null` when it does not. When it is `null`,
+    // host-based canonicalization is skipped entirely so the function never
+    // redirects viewers to a domain that has no DNS record for this
+    // distribution — see issue #3693 (production outage caused by an
+    // unconditional redirect to a non-existent custom domain).
+    var canonical = __CANONICAL_HOST__;
     var uri = request.uri;
     var query = request.querystring;
     var qs = '';
@@ -25,9 +33,12 @@ function handler(event) {
         }
     }
 
-    var targetHost = host === canonical ? host : canonical;
-    if (!/^[A-Za-z0-9.-]+$/.test(targetHost)) {
-        targetHost = canonical;
+    var targetHost = host;
+    if (canonical) {
+        targetHost = host === canonical ? host : canonical;
+        if (!/^[A-Za-z0-9.-]+$/.test(targetHost)) {
+            targetHost = canonical;
+        }
     }
     var targetUri = uri;
     if (!/^\/[A-Za-z0-9\/._-]*$/.test(targetUri)) {
@@ -37,7 +48,9 @@ function handler(event) {
         targetUri = targetUri.slice(0, -1);
     }
     var redirectNeeded =
-        proto !== 'https' || host !== canonical || targetUri !== uri;
+        proto !== 'https' ||
+        (canonical !== null && host !== canonical) ||
+        targetUri !== uri;
     if (redirectNeeded) {
         return {
             statusCode: 301,
--- a/cdk/stacks/static_site_stack.py
+++ b/cdk/stacks/static_site_stack.py
@@ -1,3 +1,4 @@
+import json
 import os
 from pathlib import Path
 
@@ -168,18 +169,29 @@ class StaticSiteStack(Stack):
             enable_accept_encoding_gzip=True,
         )
 
+        _custom_domain = "app.allotmint.io"
+        _enable_custom_domain = _is_truthy_context(self.node.try_get_context("customDomain"))
+
+        # The viewer-request Function's host-based canonical redirect must only
+        # target a domain that this deployment actually serves (alias + ACM
+        # certificate + DNS record). Substitute __CANONICAL_HOST__ at synth
+        # time with the real domain when customDomain is enabled, or with
+        # `null` otherwise so the Function skips host-based canonicalization
+        # entirely. See issue #3693 (production outage from an unconditional
+        # redirect to a non-existent custom domain).
+        _viewer_request_template = (
+            Path(__file__).resolve().parents[1] / "functions" / "viewer-request.js"
+        ).read_text(encoding="utf-8")
+        _canonical_host_literal = json.dumps(_custom_domain) if _enable_custom_domain else "null"
+        _viewer_request_source = _viewer_request_template.replace(
+            "__CANONICAL_HOST__", _canonical_host_literal
+        )
+
         redirect_fn = cloudfront.Function(
             self,
             "ViewerRequestFn",
-            code=cloudfront.FunctionCode.from_file(
-                file_path=str(
-                    Path(__file__).resolve().parents[1] / "functions" / "viewer-request.js"
-                )
-            ),
+            code=cloudfront.FunctionCode.from_inline(_viewer_request_source),
         )
-
-        _custom_domain = "app.allotmint.io"
-        _enable_custom_domain = _is_truthy_context(self.node.try_get_context("customDomain"))
 
         _certificate: acm.Certificate | None = None
         _hosted_zone: route53.IHostedZone | None = None
--- a/cdk/tests/test_static_site_stack.py
+++ b/cdk/tests/test_static_site_stack.py
@@ -425,6 +425,54 @@ def custom_domain_template(tmp_path_factory):
     return assertions.Template.from_stack(stack)
 
 
+def _viewer_request_function_code(rendered_template) -> str:
+    """Return the inline FunctionCode source of the ViewerRequestFn CloudFront Function.
+
+    Asserts exactly one CloudFront Function exists so the helper fails loudly
+    if the resource is renamed/removed rather than silently returning None.
+    """
+    resources = rendered_template.find_resources("AWS::CloudFront::Function")
+    assert len(resources) == 1, f"Expected exactly one CloudFront Function; found {len(resources)}"
+    properties = next(iter(resources.values()))["Properties"]
+    return properties["FunctionCode"]
+
+
+def test_viewer_request_function_skips_redirect_target_without_custom_domain(template):
+    """When customDomain is disabled, the Function source must not hardcode a
+    redirect target host — it must substitute `null` for __CANONICAL_HOST__.
+
+    Regression test for issue #3693: the deployed Function previously redirected
+    ALL traffic to https://app.allotmint.io regardless of whether that domain had
+    an alias, ACM certificate, or DNS record for this distribution, causing a
+    production outage (NXDOMAIN on every request).
+    """
+    source = _viewer_request_function_code(template)
+    assert "__CANONICAL_HOST__" not in source, (
+        "Function source must not contain the unsubstituted template placeholder"
+    )
+    assert "app.allotmint.io" not in source, (
+        "Function source must not hardcode the custom-domain host when "
+        "customDomain is disabled for this deployment"
+    )
+    assert "var canonical = null;" in source, (
+        "Function source must substitute `null` for __CANONICAL_HOST__ so "
+        "host-based canonicalization is skipped entirely"
+    )
+
+
+def test_viewer_request_function_redirects_to_custom_domain_when_enabled(custom_domain_template):
+    """When customDomain is enabled, the Function source must canonicalize to
+    the configured custom domain (which has a matching alias/cert/DNS record)."""
+    source = _viewer_request_function_code(custom_domain_template)
+    assert "__CANONICAL_HOST__" not in source, (
+        "Function source must not contain the unsubstituted template placeholder"
+    )
+    assert 'var canonical = "app.allotmint.io";' in source, (
+        "Function source must substitute the configured custom domain literal "
+        "for __CANONICAL_HOST__ when customDomain is enabled"
+    )
+
+
 def test_custom_domain_distribution_has_domain_names(custom_domain_template):
     """Distribution must list app.allotmint.io in Aliases when customDomain is set."""
     custom_domain_template.has_resource_properties(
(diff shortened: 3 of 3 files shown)


leonarduk:
Addressed the blocking finding in commit 11eef000:

- **Open redirect / host-header injection (blocking)**: confirmed valid. The previous revision let the raw, unvalidated `Host` header flow into the 301 `Location` whenever `canonical` was `null` and a redirect was triggered (e.g. `Host: evil.example.com` over HTTP → 301 to `https://evil.example.com/`). Fixed by applying the `/^[A-Za-z0-9.-]+$/` charset check to `targetHost` **unconditionally**, before any `Location` is constructed. When no canonical host is configured and the incoming `Host` fails the check, the function now passes the request through unmodified — `ViewerProtocolPolicy.REDIRECT_TO_HTTPS` at the distribution level still enforces the HTTPS upgrade.
- **Test coverage gap**: added `test_viewer_request_function_validates_target_host_before_building_redirect`, which asserts (for both the custom-domain and non-custom-domain synthesized sources) that the host-charset guard appears in the source *before* the `'https://' + targetHost` Location-construction string, and that the safe `return request` fallback is present.
- **Minor doc nit**: added an inline comment documenting the deliberate decision to keep the `proto !== 'https'` check as defense-in-depth alongside `ViewerProtocolPolicy.REDIRECT_TO_HTTPS`.

All 13 relevant CDK tests pass (`pytest cdk/tests/test_static_site_stack.py -k "viewer_request or custom_domain" -v`).

PR #3758, opened 2026-06-08
fix(scripts): add issue-referenced files to aider's editable context

### LINKED ISSUE #3777: aider-issue.ps1 aborts with misleading 'no commits' error when issue gives the model no edit targets
## What
`scripts/aider-issue.ps1` aborts with "Aider made no commits" whenever the local model replies with prose (e.g. "please add these files to the chat") instead of making edits. Because that reply isn't aider's structured file-add prompt, `--yes-always` can't auto-accept it, so no files ever enter the editable context and aider produces zero commits. The prior error message blamed an empty prompt or a broken model connection, which misdiagnoses the actual cause.

## Why
Weak/local models frequently need explicit file targets up front to make edits at all. Without them, every run against an issue that doesn't already have files in aider's repo-map context silently fails at the no-commits guard with a misleading error, wasting a full run and giving the operator no actionable next step.

## How
- Scan the issue title + body for path-like tokens and keep only those that exist in the repo (validated against the repo root, not `$PWD`).
- Pass the surviving paths to aider via repeatable `--file` flags so it has concrete edit targets before the prompt is sent.
- Warn explicitly when the issue references no existing files, since that's a strong signal the run will make no edits.
- Rewrite the no-commits error message to point at the real likely cause (model replied with prose instead of edits).

## Constraints
- Must preserve existing behavior when the issue references no files: empty `--file` list, aider falls back to its repo-map.
- File discovery must reject path traversal (`..`) and absolute/rooted/UNC paths — issue text is attacker-controllable.
- Must work correctly regardless of the directory the script is invoked from (resolve against repo root).

## LLM tier
**Sonnet** — requires careful PowerShell regex/path-handling changes plus security-conscious validation of attacker-influenceable input; not a trivial one-liner.

## Success looks like
- [x] Issues referencing existing repo files have those files passed to aider via `--file` before the prompt.
- [x] Issues referencing no existing files produce an explicit warning instead of a silent no-op.
- [x] The no-commits error message describes the likely cause (prose reply / no edit targets) instead of blaming an empty prompt or broken connection.
- [x] Path discovery rejects traversal and rooted paths, and resolves correctly regardless of invocation directory.

## Failure looks like
- The script still aborts with the old generic "no commits" message without guidance.
- A crafted issue body causes a file outside the repo to be added to aider's editable context.
- File discovery only works when the script is run from the repo root.

leonarduk:
## Summary

Closes #3777

`scripts/aider-issue.ps1` aborts with "Aider made no commits" whenever the local model replies with prose (e.g. *"please add these files to the chat"*) instead of edits. Because that reply isn't aider's structured file-add prompt, `yes-always` can't auto-accept it, so no files enter the editable context and aider produces no commits. The old error message blamed an empty prompt or a broken model connection, which is misleading.

## Fix

- Scan the issue title + body for path-like tokens, keep only those that **exist in the repo** (`Test-Path`), and pass them to aider as positional args so it has concrete edit targets up front. Over-broad regex matches (version numbers, URLs, `e.g`) are harmlessly discarded by the existence check.
- Warn explicitly when the issue references no existing files (a strong signal the run will make no edits — often because the issue names files that don't exist).
- Rewrite the no-commits error to point at the real cause.

## Testing

- Script parses cleanly (`Parser::ParseFile`).
- Path extraction verified against two bodies: one referencing non-existent files → 0 kept (clear warning path); one referencing real files (`scripts/aider-issue.ps1`, `.github/workflows/claude-pr-review.yml`, `README.md`) → all three extracted and added.

## Note on issue #3677

This does **not** make `aider-issue.ps1 3677` succeed: issue #3677 references `.github/workflows/claude-review.yml` / `gpt-review.yml`, which do not exist (the real files are `claude-pr-review.yml` / `gpt-pr-review.yml`). With no valid edit targets, aider still makes no commits — but now the script says so clearly instead of failing cryptically. Fixing that requires correcting the issue text, which is out of scope here.

### DIFF
--- a/scripts/aider-issue.ps1
+++ b/scripts/aider-issue.ps1
@@ -96,16 +96,38 @@ if (-not $baseSha) {
 $promptFile = [System.IO.Path]::GetTempFileName()
 Set-Content -Path $promptFile -Value "GitHub issue #${number}: $title`n`n$issueBody" -Encoding UTF8
 
+# Discover files the issue references that actually exist on disk and add them
+# to aider's editable context. Without explicit targets, weaker local models
+# tend to reply with prose like "please add these files to the chat" and make
+# no edits; because that reply is not aider's structured file-add prompt,
+# yes-always cannot accept it, so the run produces zero commits and aborts at
+# the guard below. The regex matches path-like tokens (optional dir segments +
+# filename.ext); Test-Path then keeps only the ones that exist, so over-broad
+# matches (version numbers, "e.g", URLs) are harmlessly discarded.
+$pathPattern = '(?:[\w.-]+[\\/])*[\w.-]+\.[A-Za-z0-9]+'
+$referencedFiles = @(
+    [regex]::Matches("$title`n$issueBody", $pathPattern) |
+        ForEach-Object { $_.Value } |
+        Sort-Object -Unique |
+        Where-Object { Test-Path -LiteralPath $_ -PathType Leaf }
+)
+if ($referencedFiles.Count -gt 0) {
+    Write-Host "    Adding referenced files to aider context: $($referencedFiles -join ', ')"
+} else {
+    Write-Warning "Issue text referenced no existing repo files; aider has no explicit edit targets and may make no changes. If the issue names files that do not exist, correct the issue text first."
+}
+
 Write-Host "[4/6] Running aider on issue #$number..."
-aider --yes-always --message-file $promptFile
+aider @referencedFiles --yes-always --message-file $promptFile
 Remove-Item $promptFile -ErrorAction SilentlyContinue
 if ($LASTEXITCODE -ne 0) { exit 1 }
 
 # Abort if aider made no commits - avoids pushing an empty branch and opening
-# a content-free PR (e.g. when the issue fetch silently returned no data).
+# a content-free PR. The most common cause is the model replying with prose
+# (for example asking for files) instead of edits, so auto-commit never fired.
 $newCommits = git rev-list "$baseSha..HEAD" --count 2>$null
 if (-not $newCommits -or [int]$newCommits -eq 0) {
-    Write-Error "Aider made no commits. Aborting push and PR creation. Check that the issue prompt was non-empty and that aider connected to the model successfully."
+    Write-Error "Aider made no commits, so there is nothing to push. This usually means the model replied without edits (e.g. asking for files). Review the aider output above; if the issue references files that do not exist in the repo, correct the issue text and retry."
     exit 1
 }
 Write-Host "    Aider produced $newCommits commit(s)."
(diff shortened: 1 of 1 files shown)


leonarduk:
Thanks — addressed in c0e8288a, with one correction.

**1. Path traversal (valid — fixed).** The issue body is attacker-controllable, so the filter now rejects `..` and absolute/rooted paths *before* `Test-Path`:

```powershell
Where-Object { $_ -notmatch '\.\.' -and -not [System.IO.Path]::IsPathRooted($_) } |
Where-Object { Test-Path -LiteralPath $_ -PathType Leaf }
```

Verified against a crafted body containing `../../../../Users/.../.aws/credentials.ini`, `/etc/shadow.conf`, and `C:/Windows/system.ini` mixed with a real repo path — only the in-repo file survives. With traversal and absolute paths rejected and the relative path resolved from the repo root, every kept match is confined to the repo subtree.

**2. `--file` vs positional (adopted, but the premise is incorrect).** `aider --help` shows the CLI signature ends with a positional `[FILE ...]`, so bare positional file paths *are* supported (it's aider's canonical form — `aider foo.py bar.py`); `--file` is just a repeatable alternative. So the original splat was not broken. That said, I've switched to building `--file <path>` args anyway because it's unambiguous if a filename ever looks like an option — no downside.

**3. Empty-files fast-fail (intentionally kept as proceed).** I did not add an early `exit 1` for the no-files case. A hard fail there would break legitimate issues where aider creates a *new* file or works from its repo-map (no pre-existing file to add). The model here is local Ollama, so no API quota is burned, and the existing no-commits guard already catches a genuine no-op with a clear message. I've added a comment documenting this as deliberate.

The originally-reported `aider-issue.ps1 3677` failure is separate: issue #3677 names `.github/workflows/claude-review.yml` / `gpt-review.yml`, which don't exist (real files: `claude-pr-review.yml` / `gpt-pr-review.yml`), so there are no valid edit targets regardless. That's an issue-text problem, out of scope for this PR.

leonarduk:
Thanks for the review — addressed the blocking issue and filed a follow-up for the rest:

## Fixed in this PR (a39613c9)
- **`Test-Path` resolved against `$PWD`, not repo root (REQUEST CHANGES blocker).** Now resolves `git rev-parse --show-toplevel` once and joins each candidate path against the repo root before `Test-Path`/`--file`, so the discovery works correctly regardless of which directory the script is invoked from.

## Tracked as follow-up
Filed #3770 to cover the remaining hardening suggestions, all of which are improvements to the new logic but not blockers for this fix:
- Anchor the path regex to exclude pure-decimal "extensions" (e.g. `1.2.3`).
- Cap the number of `--file` arguments injected from issue text.
- Add Pester tests for the regex extraction and traversal guards.

## Already correct (no change needed)
- The `Sort-Object -Unique` → traversal-filter ordering is safe as-is (noted in the review as informational).
- `IsPathRooted` already correctly rejects UNC and single-leading-backslash paths.
- The empty-`$referencedFiles` → empty `$fileArgs` fallback was confirmed correct.

PR #3778, opened 2026-06-10
ci(pre-deploy-check): fail if cdk diff removes an IAM Allow grant for the deploy role

### LINKED ISSUE #3741: ci(pre-deploy-check): fail if cdk diff removes an IAM Allow grant for the deploy role
## What

Add a CI guard that fails `pre-deploy-check` if a `cdk diff --method=changeset` run shows any `Allow` IAM statement being **removed** for the `allotmint-github-deploy` role (e.g. an `[-]` line touching `GithubDeployRoleForLambdaInvoke...Policy*`, or any other policy attached to that role).

## Why

`scripts/bash/pre-deploy-check.sh` step 2 ("CDK synth + diff", around line 52) currently runs `cdk diff BackendLambdaStack StaticSiteStack -c prod=true --method=changeset` but only checks the command's **exit code**. `cdk diff` exits 0 even when the diff shows resources/policies being destroyed — a removed grant on the deploy role (e.g. `s3:ListBucket`/`s3:GetObject` on `PortfolioDataBucket`, or `lambda:InvokeFunction` on `PriceRefreshLambdaLiveAlias`) would currently pass this check silently. If such a removal ever did slip through, the "Warm price snapshot" and "Verify price snapshot seeded in S3" deploy steps would start failing on the *next* deploy with no advance warning.

This is a **defence-in-depth CI improvement**, not a fix for a confirmed live regression. (A previous version of this issue claimed a specific `cdk diff` run had already shown this policy being destroyed — that run ID does not correspond to any real workflow run in this repo and could not be verified, so that claim has been dropped. The alias-targeted grant in `cdk/stacks/backend_lambda_stack.py` and the corresponding synthesis tests in `cdk/tests/test_backend_lambda_stack_permissions.py` already cover the specific permissions called out — see #3368, #3378, #3191, #3639.)

## How

1. In `scripts/bash/pre-deploy-check.sh` (and the PowerShell equivalent `scripts/powershell/pre-deploy-check.ps1`), capture the `cdk diff --method=changeset` output instead of only checking its exit code.
2. Scan the captured output for `[-]` (or changeset `Remove`/`Destroy`) lines that reference an `IAM::Policy` or `Allow` statement attached to a role/resource matching `allotmint-github-deploy` (or its imported logical IDs, e.g. `GithubDeployRoleForLambdaInvoke*`).
3. If any such removal is found, fail step 2 ("CDK diff") with a clear message naming the removed statement, even if `cdk diff` itself exited 0.
4. Add a unit/script test (or a fixture-based test using a sample diff output) that exercises both the pass case (no removals) and the fail case (a removal present), so the detection logic itself is covered.

## Constraints

- Must not produce false positives on **additions** or **replacements where the new policy appears in the same diff** (i.e. don't fail on a destroy+create pair that nets out to the same or broader permissions for the deploy role) — only flag a removal that has no corresponding addition for the same action/resource.
- Keep bash and PowerShell implementations in parity per repo convention.
- No AWS credential or live-deploy requirement beyond what step 2 already needs.

## LLM tier

sonnet — shell/CI scripting plus a small parsing test.

## Success

- A synthetic diff containing a removed `Allow` statement for the deploy role causes `pre-deploy-check` step 2 to fail with a descriptive message.
- A synthetic diff with no such removals (or with matched destroy+create pairs) passes as before.
- Existing `cdk diff` behaviour and exit-code handling for genuine `cdk diff` errors is unchanged.

## Failure

- The check only looks at the exit code (no behavioural change), or it produces false positives on normal replacement diffs.

leonarduk:
## Summary
- `pre-deploy-check`'s CDK diff step only checked `cdk diff`'s exit code, which is 0 even when the diff's "IAM Statement Changes" table shows an `Allow` statement being removed for the `allotmint-github-deploy` role.
- Adds `scripts/check_cdk_diff_iam_removals.py`, which parses that table and fails if any removed `Allow` row referencing the deploy role (or `PriceRefreshLambda`/`PortfolioDataBucket`) has no matching addition elsewhere in the diff (so destroy+create replacement pairs that net out unchanged are not flagged).
- Wires this into both `scripts/bash/pre-deploy-check.sh` and `scripts/powershell/pre-deploy-check.ps1` step 2.

Closes #3741

## Test plan
- [x] `python -m pytest tests/scripts/test_check_cdk_diff_iam_removals.py -q` — 7 new tests covering: no IAM table, removed grant flagged, replace-pair not flagged, removal for unrelated role not flagged, removed `Deny` not flagged, and `main()` exit codes.
- [x] `python -m ruff check` / `black --check` pass on the new files.
- [x] `bash -n scripts/bash/pre-deploy-check.sh` and PowerShell tokenizer check pass.
- [ ] Live `cdk diff` run against AWS (requires credentials not available in this environment).

### DIFF
--- a/scripts/bash/pre-deploy-check.sh
+++ b/scripts/bash/pre-deploy-check.sh
@@ -49,11 +49,17 @@ if [[ -z "${AWS_ACCESS_KEY_ID:-}" ]]; then
 elif [[ -z "${GITHUB_DEPLOY_ROLE_ARN:-}" || -z "${JWT_SECRET:-}" || -z "${GOOGLE_CLIENT_ID:-}" || -z "${DATA_BUCKET:-}" ]]; then
     skip "CDK diff (requires GITHUB_DEPLOY_ROLE_ARN, JWT_SECRET, GOOGLE_CLIENT_ID, DATA_BUCKET)"
 else
-    if (cd cdk && npx cdk diff BackendLambdaStack StaticSiteStack -c prod=true --method=changeset); then
-        pass "CDK diff"
-    else
+    cdk_diff_output=$(mktemp)
+    (cd cdk && npx cdk diff BackendLambdaStack StaticSiteStack -c prod=true --method=changeset) | tee "$cdk_diff_output"
+    cdk_diff_exit=${PIPESTATUS[0]}
+    if [[ $cdk_diff_exit -ne 0 ]]; then
         fail "CDK diff"
+    elif ! python3 scripts/check_cdk_diff_iam_removals.py "$cdk_diff_output"; then
+        fail "CDK diff (removes an IAM Allow grant for the deploy role — see #3741)"
+    else
+        pass "CDK diff"
     fi
+    rm -f "$cdk_diff_output"
 fi
 
 # 3. IAM permission simulation
--- a/scripts/check_cdk_diff_iam_removals.py
+++ b/scripts/check_cdk_diff_iam_removals.py
@@ -63,8 +63,10 @@ def find_iam_statement_rows(diff_output: str) -> list[list[str]]:
             continue
         cells = _table_row_cells(line)
         if cells is None:
-            # A non-table line (other than a border) ends this table.
-            if not line.strip().startswith(("┌", "├", "└", "┬", "┼", "┴")):
+            stripped = line.strip()
+            # Blank lines and box-drawing borders don't end the table; any
+            # other non-table line (e.g. the start of the next section) does.
+            if stripped and not stripped.startswith(("┌", "├", "└", "┬", "┼", "┴")):
                 in_table = False
             continue
         if "Effect" in cells and "Action" in cells:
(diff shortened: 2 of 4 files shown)


leonarduk:
Addressed both blocking points in 645f561b:

1. **Blank-line mid-table fail-open**: fixed — `find_iam_statement_rows` now only treats a non-table line as ending the IAM table if it's non-empty and not a box-drawing border; blank lines inside the table no longer terminate parsing early. Added `test_flags_removal_after_blank_line_mid_table`, which I verified fails against the pre-fix version (0 removals found instead of 1) and passes now.

2. **bash shebang / `PIPESTATUS`**: confirmed `scripts/bash/pre-deploy-check.sh` already starts with `#!/usr/bin/env bash` (line 1, pre-existing), so `PIPESTATUS` is valid here.

Also added the suggested coverage:
- `test_flags_removed_s3_grant_for_portfolio_data_bucket` (PortfolioDataBucket `s3:GetObject` removal)
- `test_flags_removal_in_ansi_colored_diff` (ANSI-escaped table rows)
- `test_main_reads_from_stdin`

11/11 tests pass, ruff/black clean.

On the `python` vs `python3` note in PS1: left as-is — the PowerShell script is Windows-targeted (it already gates `make` availability for Windows), and `python` is the conventional invocation there, consistent with the rest of `pre-deploy-check.ps1`.

PR #3880, opened 2026-06-11
feat(ci): filter low-specificity follow-up issue titles

### LINKED ISSUE #3879: Add a minimum-specificity filter to follow-up issue generation
# Add a minimum-specificity filter to follow-up issue generation

## What
In `.github/scripts/extract_followups.py` (and/or `create_followup_issues.py`), add a filter that rejects low-value, generic follow-up titles before they reach `create_issues()` — e.g. titles matching patterns like "Consider adding more (detailed|descriptive) (comments|logging|documentation|tests) ... for clarity" with no concrete file/function/line reference.

## Why
The PR-review automation has produced a large share of the open-issue backlog as near-duplicate, low-specificity issues (#3428, #3479, #3641 — all closed as not actionable). These issues add noise without giving an agent enough information to act. #3353/#3400 attempted to address duplication via exact-title dedup, which `create_followup_issues.py::issue_exists()` already implements — but exact-title matching doesn't catch near-duplicates that differ only in wording (as #3428/#3479/#3641 did), so a specificity filter that rejects vague suggestions outright addresses the root cause more directly than dedup alone.

## How
- Add a `_is_low_specificity(title: str) -> bool` helper (likely in `extract_followups.py`) using a regex denylist for generic phrasing, e.g.:
  - `r'consider adding (more )?(detailed|descriptive) (comments|logging|documentation|tests?)\b.*\bfor clarity\b'`
  - Generalize to also catch "improve readability", "add more context" etc. without a file/function reference
- A title should be rejected only if it lacks a concrete reference (backtick-quoted file/function/identifier) AND matches a generic phrasing pattern — to avoid over-filtering legitimate, specific suggestions.
- Filter titles in `extract_followups()` (or as a post-filter applied in `create_issues()`), logging skipped titles to stdout for traceability.
- Add unit tests in `tests/test_extract_followups.py` / `tests/test_create_followup_issues.py` covering: generic title without reference (rejected), generic phrasing with a concrete file/function reference (kept), specific actionable title (kept).

## Constraints
- Must not require additional GitHub API calls or permissions.
- Must not reject specific, actionable suggestions just because they mention "comments" or "tests" — the filter targets *vagueness*, not topic.
- Keep the denylist small and easy to extend; avoid building a general NLP classifier.
- Out of scope: semantic/fuzzy near-duplicate dedup across issue titles (could be a separate follow-up if noise persists after this filter).

## LLM tier
**Sonnet** — Requires designing a regex/heuristic that balances false positives/negatives, plus new unit test coverage with several edge cases; moderate judgment but no architectural complexity.

## Success looks like
- [ ] `extract_followups.py` (or `create_followup_issues.py`) rejects titles matching the generic "more detailed/descriptive X for clarity" pattern when no concrete file/function reference is present
- [ ] Specific titles referencing a file, function, or line are still created as issues
- [ ] Skipped low-specificity titles are logged so workflow runs remain auditable
- [ ] New unit tests cover at least: generic-no-reference (rejected), generic-with-reference (kept), fully specific (kept)

## Failure looks like
- The filter rejects specific, actionable suggestions because they happen to mention "tests"/"comments"/"logging"
- The filter is too narrow and #3428/#3479/#3641-style issues continue to be created
- No test coverage added, so future changes silently break the filter

---

Re-scoped from #3817 (#3353/#3400 already implemented via `issue_exists()`; #3428/#3479/#3641 already closed).

leonarduk:
## Summary
- Adds `_is_low_specificity()` to `extract_followups.py`, which drops vague "consider adding more detailed/descriptive comments/logging/tests/docs ... for clarity" style follow-ups (and similar generic phrasing like "improve readability" / "add more context") unless the title contains a backtick-quoted file/function/identifier reference.
- Skipped titles are logged to stderr for traceability without corrupting the JSON stdout output.

## Test plan
- [x] `pytest tests/test_extract_followups.py tests/test_create_followup_issues.py -q --no-cov` — 30 passed (4 new tests: generic-no-reference dropped, generic-with-reference kept, fully specific kept, plus existing coverage)
- [x] `ruff check .github/scripts/extract_followups.py` — clean

Closes #3879

### DIFF
--- a/.github/scripts/extract_followups.py
+++ b/.github/scripts/extract_followups.py
@@ -7,18 +7,21 @@ import re
 import sys
 from pathlib import Path
 
-_REFERENCE_PATTERN = re.compile(r'`[^`]+`')
+# Require at least 3 chars so trivial backtick tokens (`` `s` ``, `` `1` ``)
+# don't masquerade as a file/function/identifier reference.
+_REFERENCE_PATTERN = re.compile(r'`[^`]{3,}`')
 
 # Generic, low-value phrasing that gives an agent nothing concrete to act on.
 # A title is only rejected if it matches one of these AND has no backtick-quoted
 # file/function/identifier reference (see _is_low_specificity).
 _GENERIC_PHRASING_PATTERNS = [
     re.compile(
-        r'\bconsider adding (?:more )?(?:detailed|descriptive)\b.*\bfor clarity\b',
+        r'\bconsider adding (?:more )?(?:detailed|descriptive)?\s*'
+        r'(?:comments?|logging|documentation|tests?)\b.*\bfor clarity\b',
         re.IGNORECASE,
     ),
     re.compile(r'\bimprove readability\b', re.IGNORECASE),
-    re.compile(r'\badd more context\b', re.IGNORECASE),
+    re.compile(r'\badd(?:ing)? more context\b', re.IGNORECASE),
 ]
 
 
--- a/.github/scripts/extract_followups.py
+++ b/.github/scripts/extract_followups.py
@@ -7,6 +7,27 @@ import re
 import sys
 from pathlib import Path
 
+_REFERENCE_PATTERN = re.compile(r'`[^`]+`')
+
+# Generic, low-value phrasing that gives an agent nothing concrete to act on.
+# A title is only rejected if it matches one of these AND has no backtick-quoted
+# file/function/identifier reference (see _is_low_specificity).
+_GENERIC_PHRASING_PATTERNS = [
+    re.compile(
+        r'\bconsider adding (?:more )?(?:detailed|descriptive)\b.*\bfor clarity\b',
+        re.IGNORECASE,
+    ),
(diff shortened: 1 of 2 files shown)


leonarduk:
Addressed in 9e0e1923:

- \`_REFERENCE_PATTERN\` now requires >=3 chars (\`` `[^\`]{3,}` ``) so trivial tokens like `` `s` `` no longer bypass the filter — added a regression test for this.
- Pattern 0 no longer requires \`detailed\`/\`descriptive\`, so "consider adding more comments for clarity" is also caught.
- Added tests for the \`improve readability\` and \`add more context\`/\`adding more context\` patterns (the latter needed a small regex fix — \`add(?:ing)? more context\`).

On the stdout vs stderr point: keeping skip logs on stderr is intentional, not a discrepancy to fix — \`main()\` prints the extracted titles as JSON on stdout for the workflow to consume, so any diagnostic logging on stdout would corrupt that JSON. stderr is the correct stream here; CI captures both.

34/34 tests pass, ruff clean.

PR #4063, opened 2026-06-13
fix(cdk): fail synthesis loudly when GITHUB_DEPLOY_ROLE_ARN is missing in prod

### LINKED ISSUE #3870: `infra: harden static_site_stack.py to fail synthesis loudly if github_deploy_role_arn is absent in a prod context` — the current silent omission of the IAM grant is what made this bug invisible for multiple deploy cycles
# infra: harden static_site_stack.py to fail synthesis loudly if github_deploy_role_arn is absent in a prod context

## What

Modify `infra/static_site_stack.py` to raise an explicit error during CDK synthesis (not a silent no-op) when:
- The stack is being synthesized for a **production environment** (`prod` stage)
- AND `github_deploy_role_arn` parameter is `None` or an empty string

Currently, the code contains:
```python
if github_deploy_role_arn:
    # grant permissions
```

This silently omits the IAM grant when the parameter is absent, making the bug invisible across multiple deploy cycles (as happened in PR #3866).

**Location:** `infra/static_site_stack.py`, in the `StaticSiteStack.__init__` method, around the point where `github_deploy_role_arn` is checked.

## Why

**Root cause visibility:** The silent omission of `GITHUB_DEPLOY_ROLE_ARN` in a GitHub Actions step caused the IAM grant resource to be synthesized away without any signal to the operator. By the time CloudFormation deleted the resource, the failure was attributed to an unrelated step.

**Prevention:** A loud, early failure at synthesis time (not deploy time) forces the operator to fix the env var configuration *before* any CloudFormation drift occurs. This catches the mistake in the same deploy cycle.

**Future-proofing:** An LLM (or human reviewer) adding a new `cdk deploy StaticSiteStack` step will immediately see a synthesis error if they forget to set `GITHUB_DEPLOY_ROLE_ARN`, rather than discovering it 20 minutes into a deploy.

## How

1. In `static_site_stack.py`, change the guard from a silent conditional to an explicit check:
   ```python
   if self.stage == "prod" and not github_deploy_role_arn:
       raise ValueError(
           "github_deploy_role_arn is required for prod StaticSiteStack "
           "(passed via GITHUB_DEPLOY_ROLE_ARN env var). "
           "See #3847 and #3866 for context."
       )
   ```

2. Place this check **early** in `__init__`, before any resource synthesis that depends on the role.

3. Keep the existing `if github_deploy_role_arn:` block for non-prod stages (dev, staging) where the grant is optional.

## Constraints

- **Must not break existing non-prod deploys:** Dev and staging stacks should synthesize successfully with `github_deploy_role_arn` unset (they do not need the grant).
- **Must not affect CloudFormation behavior in this PR:** The fix is synthesis-time only; no changes to the IAM grant logic itself.
- **Out of scope:** Adding a CI lint check (covered by separate follow-up); this issue is infra hardening only.

## LLM tier

**local-14b**

This is a straightforward defensive check in a single function. No ambiguity about placement or logic; the error message can be templated from the context provided (PR #3847, #3866, and the env var name).

## Success looks like

- [ ] `cdk synth StaticSiteStack` for prod stage fails immediately with a clear error message naming `GITHUB_DEPLOY_ROLE_ARN` if the env var is unset or empty.
- [ ] `cdk synth StaticSiteStack` for dev/staging stages succeeds with `github_deploy_role_arn` unset (no regression).
- [ ] A future deploy cycle that forgets to set `GITHUB_DEPLOY_ROLE_ARN` in a prod step catches the mistake at synthesis, not at CloudFormation drift time.

## Failure looks like

- The check is placed *after* resource synthesis, allowing the buggy template to be synthesized.
- The error is raised for non-prod stages, breaking dev/staging deploys.
- The error message does not mention the env var name or upstream issues, leaving the operator confused about how to fix it.

---

_Follow-up from AI review of PR #3866._

leonarduk:
## Summary
- `StaticSiteStack.__init__` now raises a `ValueError` early during synthesis if `prod` context is truthy and `GITHUB_DEPLOY_ROLE_ARN` is unset/empty, instead of silently omitting the IAM grant.
- Non-prod stages (dev/staging) continue to synthesize fine without the env var, since the existing `if github_deploy_role_arn:` block remains optional.

## Why
A missing `GITHUB_DEPLOY_ROLE_ARN` previously caused CloudFormation to silently delete the deploy-role IAM grant resource with no signal, masking the real cause of a deploy failure across multiple cycles (see #3847, #3866).

## Test plan
- [x] `python -m pytest cdk/tests/test_static_site_stack.py -q` — 48 passed
- [x] New test `test_prod_without_github_deploy_role_arn_raises` confirms prod synth fails fast with a clear error naming `GITHUB_DEPLOY_ROLE_ARN`
- [x] New test `test_prod_with_github_deploy_role_arn_synthesises` confirms prod synth succeeds when the env var is set
- [x] New test `test_non_prod_without_github_deploy_role_arn_synthesises` confirms dev/staging still synth without it
- [x] Existing `test_ui_auth_user_pool_is_retained_for_prod_context` updated to set the env var since it uses `prod=true` context

Closes #3870

### DIFF
--- a/.github/workflows/cdk-dry-run.yml
+++ b/.github/workflows/cdk-dry-run.yml
@@ -61,6 +61,8 @@ jobs:
         # npm ci step above), so npx resolves the correct version without a
         # separate npm ci inside cdk/. No cdk/package.json exists in this repo.
         # JWT_SECRET and GOOGLE_CLIENT_ID are required non-empty by backend_lambda_stack.py.
+        # GITHUB_DEPLOY_ROLE_ARN is required by static_site_stack.py when prod=true
+        # (issue #3870) so synth fails loudly if the deploy workflow ever forgets it.
         # Dummy placeholder values are safe — synth only generates the
         # CloudFormation template; no deploy or AWS call is made.
         # --output flag (not CDK_OUTDIR env var) guarantees isolation per run
@@ -68,6 +70,7 @@ jobs:
         env:
           JWT_SECRET: dry-run-placeholder
           GOOGLE_CLIENT_ID: dry-run-placeholder.apps.googleusercontent.com
+          GITHUB_DEPLOY_ROLE_ARN: arn:aws:iam::123456789012:role/dry-run-placeholder
         run: npx cdk synth BackendLambdaStack StaticSiteStack -c prod=true --output /tmp/cdk.out.${{ github.run_id }}.${{ github.run_attempt }}
         working-directory: cdk
 
--- a/cdk/stacks/static_site_stack.py
+++ b/cdk/stacks/static_site_stack.py
@@ -47,6 +47,19 @@ class StaticSiteStack(Stack):
     ) -> None:
         super().__init__(scope, construct_id, **kwargs)
 
+        # Fail synthesis loudly if the GitHub deploy role grant would be
+        # silently omitted in a production deploy. A missing GITHUB_DEPLOY_ROLE_ARN
+        # previously caused CloudFormation to delete the IAM grant resource
+        # without any signal, masking the real cause of a deploy failure.
+        # See #3847 and #3866 for context.
+        github_deploy_role_arn = os.getenv("GITHUB_DEPLOY_ROLE_ARN", "")
+        if _is_truthy_context(self.node.try_get_context("prod")) and not github_deploy_role_arn:
+            raise ValueError(
+                "GITHUB_DEPLOY_ROLE_ARN is required for prod StaticSiteStack "
+                "(passed via the GITHUB_DEPLOY_ROLE_ARN env var). "
+                "See #3847 and #3866 for context."
+            )
+
         ui_auth_removal_policy = _ui_auth_removal_policy(self)
 
         # BucketDeployment.Source.json_data() only accepts intra-stack tokens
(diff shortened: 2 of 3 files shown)


leonarduk:
Thanks for the review — addressed the CI gap, and responding to the blocking point:

**On \`self.stage\` vs context key:** \`StaticSiteStack\` has no \`stage\` constructor parameter or \`self.stage\` attribute anywhere in this codebase — that was the AI-suggested issue's hypothetical example, not an existing field. The actual existing convention for detecting prod in this exact file is \`_is_truthy_context(self.node.try_get_context("prod"))\`, used a few lines below by \`_ui_auth_removal_policy\` (cdk/stacks/static_site_stack.py:30) to decide UserPool retention. The new guard reuses that same, already-load-bearing signal, so it can't drift from the mechanism the rest of the stack relies on for prod detection.

**CI regression caught and fixed:** \`.github/workflows/cdk-dry-run.yml\` runs \`cdk synth ... -c prod=true\` on every PR without \`GITHUB_DEPLOY_ROLE_ARN\` set, which would have failed under the new guard. Added a dummy placeholder ARN (\`arn:aws:iam::123456789012:role/dry-run-placeholder\`) to that step's env, matching the existing \`JWT_SECRET\`/\`GOOGLE_CLIENT_ID\` dry-run-placeholder pattern. \`iac-validation.yml\`'s \`cdk synth\` doesn't pass \`-c prod=true\` and \`cdk.json\` doesn't set \`prod\` context, so it's unaffected.

**\`_is_truthy_context(None)\`:** returns \`False\` (falls through both the bool and str branches), so absent context is correctly treated as non-prod.

PR #4114, opened 2026-06-13
fix(ci): hard-fail price snapshot check on non-404 S3 errors

### LINKED ISSUE #3798: `fix(deploy): convert step-24 non-NoSuchKey errors (403, 500, network) to hard deploy failures` — the current code still swallows a genuine `403 Access Denied` (e.g. misconfigured bucket policy) as a warning, which was flagged as a known gap in the linked issue.
### What
In `deploy.sh`, the step-24 S3 bucket existence check (around the `aws s3api head-object` call) currently treats all non-`NoSuchKey` errors (403, 500, network timeouts, etc.) as soft warnings and continues deployment. This masks genuine configuration errors like misconfigured bucket policies.

The code should be modified to:
- **Soft-pass** on `NoSuchKey`/`Not Found`/`404` errors (bucket doesn't exist; proceed to create it)
- **Hard-fail** on `403 Access Denied`, `500 Internal Server Error`, connection timeouts, and other non-404 errors

### Why
**Correctness**: A `403 Access Denied` error indicates the deployment will fail later anyway, but at a much later stage when it's harder to diagnose. Failing fast with a clear error message improves debuggability.

**Agent confusion**: An LLM reviewing deployment logs may misinterpret a warning-and-continue pattern as intentional fallback behavior rather than a gap, leading to incorrect assumptions about what the code does.

**Configuration safety**: Misconfigured IAM policies silently proceeding to step-25 could result in partial deployments or cryptic failures downstream.

### How
1. Capture the full error output from the `aws s3api head-object` call (already done via `stderr`).
2. After the existing `grep -qi "NoSuchKey\|Not Found\|404"` check, add a new branch that:
   - Logs the error with context (e.g., "S3 bucket access failed with 403: check IAM policy")
   - Exits with a non-zero code to halt deployment
3. This applies to any exit code that is not 0 (success) and not 404-class (bucket missing).

**Pseudocode**:
```bash
if aws s3api head-object ... 2>error.log; then
  # bucket exists and is accessible
elif grep -qi "NoSuchKey\|Not Found\|404" error.log; then
  # soft pass: proceed to create bucket
else
  # hard fail: genuine error (403, 500, timeout, etc.)
  echo "ERROR: S3 bucket access check failed: $(cat error.log)" >&2
  exit 1
fi
```

### Constraints
- **Must preserve** the existing soft-pass behavior for `NoSuchKey`/404 errors.
- **Must not** suppress the actual error message or exit code; the operator should see the AWS CLI error details.
- **Out of scope**: changes to bucket creation logic, IAM policy validation, or retry logic.
- **Backward compatibility**: no change to the external interface or expected behavior on success paths.

### LLM Tier
**Sonnet** — requires understanding the error-handling semantics of the S3 API, careful distinction between soft and hard failures, and integration with existing grep-based detection logic. A smaller model might conflate "non-zero exit" with "always fail," missing the 404 exception.

### Success looks like
- [ ] Unit or integration test shows that a `403` error now causes `deploy.sh` to exit with non-zero status and log a clear error message.
- [ ] A `404` error (bucket missing) still results in soft pass and bucket creation attempt.
- [ ] Manual testing: a misconfigured bucket policy now fails immediately at step-24 with an actionable error, not at step-25.
- [ ] Code review confirms error message includes the AWS CLI error details for debugging.

### Failure looks like
- A `403` error is still logged as a warning and deployment continues.
- The error message is suppressed or generic ("deployment failed").
- The `404` soft-pass is broken; bucket creation is skipped when it should proceed.
- The change is too broad and now rejects legitimate transient errors without retry logic.

---

_Follow-up from AI review of PR #3797._

leonarduk:
## Summary
- The "Verify price snapshot seeded in S3" step in `deploy-lambda.yml` previously treated all non-zero `aws s3api head-object` results as warnings via `continue-on-error: true`.
- `NoSuchKey`/`Not Found`/`404` (bucket/key not yet seeded) remains a soft pass — the price-refresh job may not have run yet.
- Any other error (403 Access Denied, 500, network timeout, etc.) now logs `ERROR: S3 bucket access check failed: ...` and exits 1, failing the deploy job, instead of silently continuing to later, harder-to-diagnose stages.

## Test plan
- [x] `pytest tests/scripts/test_deploy_lambda_workflow.py -q` passes, including a new test asserting the hard-fail/soft-pass split and the absence of `continue-on-error` on this step.
- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-lambda.yml'))"` confirms the workflow YAML is still valid.

Closes #3798

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -700,7 +700,12 @@ jobs:
         # snapshot is still absent here it means the Trigger Lambda failed or timed
         # out; print a clear, actionable warning so operators know exactly what
         # to do rather than hunting through CloudWatch logs.
-        continue-on-error: true
+        #
+        # A missing key (NoSuchKey/404) is a soft pass: the price-refresh job
+        # hasn't run yet, which is recoverable. Any other error (403, 500,
+        # network timeout, etc.) indicates a genuine configuration problem
+        # (e.g. a misconfigured bucket policy) that will only get harder to
+        # diagnose later in the deploy, so it is treated as a hard failure.
         env:
           AWS_REGION: ${{ secrets.AWS_REGION || vars.AWS_REGION }}
           DATA_BUCKET: ${{ secrets.DATA_BUCKET || vars.DATA_BUCKET }}
@@ -718,7 +723,8 @@ jobs:
           elif echo "$head_output" | grep -qi "NoSuchKey\|Not Found\|404"; then
             echo "::warning::Price snapshot prices/latest_prices.json is NOT present in s3://${DATA_BUCKET}. Portfolio prices will be unavailable until the price-refresh job runs. Trigger it manually: aws lambda invoke --function-name <PriceRefreshLambda-physical-name>:live /dev/null"
           else
-            echo "::warning::Could not verify price snapshot — s3api head-object returned exit ${head_exit}: ${head_output}"
+            echo "ERROR: S3 bucket access check failed: head-object returned exit ${head_exit}: ${head_output}" >&2
+            exit 1
           fi
 
       - name: Warm up backend Lambda (absorb cold-start before validation)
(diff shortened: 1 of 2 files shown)


leonarduk:
Confirmed: the \`head_output=\$(...)\` capture (unchanged by this PR) already ends with \`--key "prices/latest_prices.json" 2>&1)\`, so stderr (where AWS CLI writes \`NoSuchKey\`/\`AccessDenied\`/etc.) is merged into \`head_output\` and the \`grep -qi "NoSuchKey\|Not Found\|404"\` check works as intended.

Tightened the regression test in f6d28279 to assert \`2>&1\` is specifically on the \`head_output=\` capture's closing continuation line, not just anywhere in the script.

PR #4117, opened 2026-06-13
Catch smoke test failures earlier in CI

leonarduk:
## Summary

Frontend smoke test failures are currently only discovered **after** a production deploy
(`deploy-lambda.yml` → `smoke-test` job), so a bad change reaches the live environment
before anyone finds out. This PR moves that detection earlier and makes the post-deploy
check an explicit gate.

## Changes

1. **`.github/workflows/ci.yml`** — new `frontend-smoke` job that runs on every push/PR:
   - Builds the Vite preview (`npm run build:preview`)
   - Installs Chromium and runs `tests/smoke.spec.ts` against it
   - Runs alongside the existing lint/type-check/unit-test jobs, so failures block the PR
     before any deploy is attempted.

2. **`.github/workflows/deploy-lambda.yml`** — new `verify-smoke-tests` job:
   - Runs after the existing post-deploy `smoke-test` job
   - Fails the workflow explicitly if `smoke-test` did not succeed, so a deploy with
     failing smoke tests is clearly marked as failed rather than just having a
     non-blocking job report red.

## Timeline

| Stage | Before | After |
|-------|--------|-------|
| PR/push | not tested | `frontend-smoke` runs against local preview build |
| Deploy | proceeds regardless | unchanged — still gated on `ci.yml` passing |
| Post-deploy | `smoke-test` could fail without blocking | `verify-smoke-tests` makes that failure explicit |

## Known caveat

[#4100](https://github.com/leonarduk/allotmint/issues/4100) tracks 16 smoke tests that are
currently failing (missing headings/elements on several routes). Until that issue is
resolved, the new `frontend-smoke` CI job is expected to **fail on every PR**, including
this one. Once #4100 lands, this job should go green and can be added to the required
status checks list.

Relates to #4100

## Test plan

- [ ] Confirm `frontend-smoke` job runs in CI (expected to fail until #4100 is fixed)
- [ ] Confirm `verify-smoke-tests` appears in the next `deploy-lambda.yml` run and passes
      when `smoke-test` passes
- [ ] After #4100 is fixed, confirm `frontend-smoke` goes green and consider adding it to
      branch protection required checks

## Summary by CodeRabbit

* **Tests**
  * Expanded frontend smoke suite with centralized deterministic mocks, broader route and content assertions, hardened flows, and added coverage for extra entry modes and routes to improve UI reliability.

* **Chores**
  * Added automated frontend smoke runs to CI (including browser install) and a post-deploy verification step that enforces smoke-test success before confirming deployment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,3 +149,25 @@ jobs:
         env:
           ALLOTMINT_SKIP_SNAPSHOT_WARM: 'true'
         run: pytest --no-cov tests/test_backend_api.py::test_health tests/backend/common/test_data_provider_parity.py -q
+
+  frontend-smoke:
+    name: Frontend smoke tests (preview build)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+          cache: npm
+          cache-dependency-path: frontend/package-lock.json
+      - run: npm ci
+        working-directory: frontend
+      - name: Build preview
+        run: npm run build:preview
+        working-directory: frontend
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps chromium
+        working-directory: frontend
+      - name: Run frontend smoke tests
+        run: npx playwright test --config playwright.config.ts tests/smoke.spec.ts
+        working-directory: frontend
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,7 +159,15 @@ jobs:
         with:
           node-version: '24'
           cache: npm
-          cache-dependency-path: frontend/package-lock.json
+          cache-dependency-path: |
+            package-lock.json
+            frontend/package-lock.json
+      # Root install brings in esbuild (a transitive dependency of the root
+      # tsx devDependency). Vite v8 treats esbuild as an optional peer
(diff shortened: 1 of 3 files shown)


leonarduk:
from Deepseek 

> **Addressing the DeepSeek review's three flagged items — all false alarms or intentional:**
>
> **1. `verify-smoke-tests` "fails on skipped smoke-test" — by design**
>
> The PR description says: "Fails the workflow explicitly if smoke-test did not succeed." If `smoke-test` was skipped, an upstream dependency didn't complete — validation is incomplete and the deploy should not report green. The `if: always()` + `needs.smoke-test.result != 'success'` pattern is idiomatic GitHub Actions for "upstream must have succeeded."
>
> **2. Path filters missing — not a blocker**
>
> The job runs in ~4 minutes. Path filters are a reasonable optimisation (tracked in #4184) but not blocking — the point of this PR is catching issues in every PR before deploy.
>
> **3. Cache key collision — not a real issue**
>
> `setup-node` hashes all files in `cache-dependency-path`. If either lockfile changes, the key changes. `npm ci` is always the source of truth regardless of cache state. Standard multi-project caching.
>
> The actual fix (commits `f4d0e019` + `24ea04cf`) resolves #4100: all 45 smoke tests pass by mocking `/config`, `/owners`, `/groups` with Zod-compliant responses.

Issue #4238, opened 2026-06-13
fix: Cognito token exchange fails with 401 — frontend doesn't pass Bearer token to /token/cognito


leonarduk:
## Summary

The Cognito authentication flow is broken: the frontend's `exchangeCognitoForBackendToken` function calls `POST /token/cognito` on the API Gateway, but the API Gateway's Cognito JWT authorizer rejects the request with 401 because no Bearer token is present in the `Authorization` header.

This cascades into the "Backend unavailable — HTTP 401" error on every page load for unauthenticated users, because the token exchange fails silently, leaving the app with no valid auth token.

**⚠️ Smoke tests passed** because they use the `SmokeTestClient` (which has `USER_PASSWORD_AUTH` enabled) to obtain a Cognito JWT directly and pass it as a Bearer token to API endpoints. This tests the API Gateway → Lambda path but completely **bypasses the browser-based token exchange flow** (`ensureAwsUiAuth` → `exchangeCognitoForBackendToken`). The frontend's `/token/cognito` call is never exercised by any test.

### API Gateway routes (backend_lambda_stack.py ~line 397–441)

All routes are protected by the Cognito JWT authorizer (`HttpUserPoolAuthorizer`) except:
- `GET /health`
- `GET /config`
- `OPTIONS /*`

The catch-all `ANY /{proxy+}` route — which matches `POST /token/cognito` and `POST /token/google` — requires a valid Cognito JWT in the `Authorization` header.

### Frontend token exchange (main.tsx ~line 432–450)

```typescript
const exchangeCognitoForBackendToken = async (awsUiAuth) => {
  const idToken = getStoredCognitoIdToken();
  const clientId = awsUiAuth?.clientId?.trim();
  if (!idToken || !clientId) return;
  const existing = getStoredAuthToken();
  if (existing && !isJwtExpired(existing)) return;
  const res = await fetch(`${getApiBase()}/token/cognito`, {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' },
    // ❌ No Authorization header — Cognito ID token is only in the body
    body: JSON.stringify({ id_token: idToken, client_id: clientId }),
  });
  // ...
};
```

The Cognito ID token is sent in the **request body** but never set as a Bearer token in the `Authorization` header. The Cognito **access token** (also stored in sessionStorage via `awsUiAuth.ts` → `storeSession`) is never read or used anywhere.

### What happens

1. User completes Cognito hosted UI login → tokens stored in sessionStorage
[...]

PR #4243, opened 2026-06-14
Add Playwright smoke test for Cognito hosted UI to backend token exchange

### LINKED ISSUE #4241: Add Playwright smoke test for Cognito hosted UI login → token exchange → API call flow
# Add Playwright smoke test for Cognito hosted UI login → token exchange → API call flow

## What
There is no end-to-end browser test that exercises the full Cognito authentication flow: redirect to the Cognito hosted UI, callback handling, exchange of the Cognito token for a backend token via `/token/cognito`, and a subsequent authenticated API call. [apps#4238](https://github.com/leonarduk/allotmint/pull/4238) fixed a bug where the Cognito Bearer token was not passed to the `/token/cognito` exchange, but the fix was only covered by unit tests around `awsUiAuth.ts` (`exchangeCognitoForBackendToken`, `getStoredCognitoAccessToken`). A regression in the wiring between hosted-UI callback, token storage, and the exchange call would not be caught by unit tests alone.

## Why
This flow is security-critical and has already regressed once (#4238). A browser-level smoke test would catch future regressions where the access token (or its ID-token fallback) stops being attached to the `/token/cognito` request, or where the callback/storage wiring breaks, before they reach production.

## How
- Add a new Playwright spec under `frontend/tests/` (e.g. `cognito-auth-flow.spec.ts`), following the existing patterns in `frontend/tests/smoke.spec.ts` and `frontend/playwright.config.ts`.
- Mock/stub the Cognito hosted UI redirect and callback (or use a test Cognito user pool if available) so the test does not depend on live AWS infrastructure.
- Assert that after callback handling, `awsUiAuth.ts` stores a Cognito access token (or ID token fallback) and that the subsequent request to `/token/cognito` includes the `Authorization: Bearer <token>` header.
- Assert that the exchanged backend token is then used for a real authenticated API call (e.g. a `/trail` or `/accounts` request returns 200, not 401).
- Wire the new spec into the existing smoke test scripts (`npm run smoke:test` / `scripts/smoke-all.ts`) so it runs alongside other smoke tests in CI.

## Constraints
- Must not depend on live AWS Cognito credentials in CI — use mocked network responses (e.g. via Playwright route interception) for the hosted UI and `/token/cognito` endpoints.
- Should follow the existing smoke test conventions (`frontend/tests/smoke.spec.ts`, `docs/SMOKE_TESTS.md`) for consistency.
- Keep the test fast and deterministic; avoid flaky timing dependencies on redirects.

## LLM tier
**Sonnet** — requires understanding the existing auth flow (`awsUiAuth.ts`), Playwright route mocking, and wiring a new spec into the smoke test scripts, but doesn't require deep architectural changes.

## Success looks like
- [ ] A new Playwright spec exists that covers hosted UI redirect/callback, `/token/cognito` exchange with the correct Bearer header, and a subsequent authenticated API call
- [ ] The spec runs without live AWS credentials (mocked Cognito/token endpoints)
- [ ] The spec is included in `npm run smoke:test` (or the relevant smoke test script) and passes in CI

## Failure looks like
- The new test depends on live Cognito infrastructure and is skipped/flaky in CI
- The test only re-checks the unit-level logic already covered by `awsUiAuth.test.ts` without exercising the browser-level wiring
- The Bearer header / token exchange regression from #4238 would not be caught by the new test if reintroduced

leonarduk:
## Summary
- Adds `frontend/tests/cognito-auth-flow.spec.ts`, a Playwright smoke test covering the full Cognito hosted UI redirect → callback → `/token/cognito` exchange → authenticated API call flow.
- Mocks the Cognito hosted UI token endpoint and `/token/cognito`, asserting the Cognito access token is sent as `Authorization: Bearer <token>` (regression coverage for #4238/#4239) and that the resulting backend token is then attached to subsequent API requests.
- Wires the new spec into `smoke:frontend` so it runs alongside the existing smoke suite in CI, without depending on live AWS Cognito infrastructure.

## Test plan
- [x] `npx playwright test --config playwright.config.ts tests/cognito-auth-flow.spec.ts` passes
- [x] `npx playwright test --config playwright.config.ts tests/smoke.spec.ts tests/cognito-auth-flow.spec.ts` — all 46 tests pass
- [x] `npx eslint --format unix tests/cognito-auth-flow.spec.ts` — clean

Closes #4241.

## Tests
- Expanded the existing frontend smoke coverage by adding a dedicated Playwright test for the complete Cognito hosted UI callback flow through the backend token exchange.
- The new test validates PKCE handling, correct token exchange payloads/headers, persistence and reuse of the backend auth token for follow-up requests, and removal of temporary OAuth query parameters after sign-in.
- The `smoke:frontend` script now runs both the standard smoke test and this new authentication-flow test.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,7 +13,7 @@
     "coverage": "vitest run --coverage",
     "test:jsdom": "vitest --environment=jsdom",
     "type-check": "tsc -b --noEmit",
-    "smoke:frontend": "npm run build:preview && playwright install chromium && playwright test --config playwright.config.ts tests/smoke.spec.ts",
+    "smoke:frontend": "npm run build:preview && playwright install chromium && playwright test --config playwright.config.ts tests/smoke.spec.ts tests/cognito-auth-flow.spec.ts",
     "prepare": "node -e \"if(require('fs').existsSync('.git')){require('husky').install();}\"",
     "deploy:aws": "powershell -ExecutionPolicy Bypass -File scripts/deploy-to-aws.ps1 -BucketName $env:S3_BUCKET -DistributionId $env:CLOUDFRONT_DISTRIBUTION_ID",
     "deploy:aws:linux": "bash scripts/deploy-to-aws.sh -b $S3_BUCKET ${CLOUDFRONT_DISTRIBUTION_ID:+-d $CLOUDFRONT_DISTRIBUTION_ID}",
--- a/frontend/tests/cognito-auth-flow.spec.ts
+++ b/frontend/tests/cognito-auth-flow.spec.ts
@@ -11,6 +11,13 @@ const COGNITO_ID_TOKEN = 'cognito-id-token';
 const COGNITO_ACCESS_TOKEN = 'cognito-access-token';
 const BACKEND_ACCESS_TOKEN = 'backend-access-token';
 
+type TokenRequest = { url: string; body: string };
+type CognitoExchangeRequest = {
+  authorization: string | undefined;
+  body: { id_token?: string; client_id?: string };
+};
+type ConfigRequest = { authorization: string | undefined };
+
 /**
  * Seed the PKCE state/verifier that awsUiAuth.ts writes before redirecting to
  * the hosted UI, so the simulated callback (?code=...&state=...) is accepted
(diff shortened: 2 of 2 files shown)


leonarduk:
Addressed the CodeRabbit refactor suggestion in e9d23a72 — split the test body into named helpers (mock setup + assertion blocks) to stay under the 60-line guideline. Re-ran `npx playwright test tests/cognito-auth-flow.spec.ts` and `npx eslint` after the split; both pass.

Re the DeepSeek review's blocking points — these don't hold up against the passing "Frontend smoke tests (preview build)" check:

1. **JSON body vs form-encoded for `/token/cognito`**: `exchangeCognitoForBackendToken` in `awsUiAuth.ts` sends `JSON.stringify({ id_token, client_id })` with `Content-Type: application/json`, so `JSON.parse(request.postData())` in the mock is correct. If it were form-encoded as claimed, the test would throw a `SyntaxError` and the CI smoke job would fail — it didn't.
2. **sessionStorage vs localStorage for PKCE state**: `awsUiAuth.ts` stores `awsUiAuthState`/`awsUiAuthCodeVerifier` in `window.sessionStorage`, matching `seedPkceState`. Again, if this were wrong, `exchangeCode()` would reject the callback as invalid state and the test would fail.
3. **`scripts/smoke-all.ts` not wired up**: it shells out to `npm run smoke:frontend`, which was updated in this PR to include `tests/cognito-auth-flow.spec.ts` — so it's already covered.
4. **Port mismatch (5173 vs 4173)**: this matches the existing `tests/smoke.spec.ts` convention (`process.env.SMOKE_URL ?? 'http://localhost:5173'`), so there's no new inconsistency.

The suggested follow-ups (ID-token-fallback case, invalid-state rejection case) are good ideas for additional coverage but out of scope for this issue (#4241) — happy to file a follow-up issue if useful.

leonarduk:
Re the latest DeepSeek review — the blocking claim and the two "bugs" don't hold up:

1. **Port 5173 vs 4173 ("blocking bug")**: `tests/smoke.spec.ts` (the existing, passing smoke spec) and `playwright.config.ts` both already use `process.env.SMOKE_URL ?? 'http://localhost:5173'` / `url: 'http://localhost:5173'`. This new spec matches that exact convention — there's no inconsistency to fix, and the CI "Frontend smoke tests (preview build)" check passed using this same default.
2. **`authToken` localStorage key**: confirmed in `src/api.ts` — `const TOKEN_STORAGE_KEY = "authToken"` is the actual key used to read the bearer token for outgoing requests. The test's `assertBackendTokenIsReused` polling on `localStorage.getItem('authToken')` is correct and passed in CI.
3. **`JSON.parse(request.postData() ?? '{}')`**: `/token/cognito` is a POST with a JSON body (per `exchangeCognitoForBackendToken` in `src/awsUiAuth.ts`), so `postData()` is non-null here. This is consistent with the existing codebase's route-mocking style and the test passes.

Non-blocking suggestions taken on board:
- The hardcoded file list in `smoke:frontend` matches the existing convention (explicit list, not a glob) — keeping it consistent with how `tests/smoke.spec.ts` was added originally.
- ID-token-fallback and invalid-state test cases are good follow-up coverage but out of scope for #4241; can file a follow-up issue if wanted.

No code changes needed for this round — all three "bugs"/blocking points are based on incorrect assumptions about the codebase, contradicted by the passing CI smoke run.

Issue #4027, opened 2026-06-12
fix(cdk): SmokeTestClient tokens rejected (401) by BackendApi Cognito authorizer on Cognito-protected routes


leonarduk:
## What

After a deploy, the post-deploy frontend smoke tests (`npm --prefix frontend run smoke:frontend`, run in `.github/workflows/deploy-lambda.yml`) fail 16/45 tests. Run: https://github.com/leonarduk/allotmint/actions/runs/27445592817/job/81131666959

12 of those 16 failures show the same overlay error toast blocking every page:

```
HTTP 401 - (https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com/groups)
```

`GET /groups` is called on app bootstrap (`frontend/src/App.tsx`, `MainApp.tsx`, `useRouteMode.ts`) and is routed through the `/{proxy+}` ANY route, which is protected by `backend_authorizer` (Cognito JWT authorizer). The 401 error toast then blocks the page heading assertions in `frontend/tests/smoke.spec.ts`.

## Why

`backend_lambda_stack.py` builds `backend_authorizer` (`HttpUserPoolAuthorizer`) with:

```python
user_pool_clients=[
    cognito.UserPoolClient.from_user_pool_client_id(
        self, "ImportedUiAuthUserPoolClient", ui_auth_user_pool_client_id_param.value_as_string
    )
],
```

i.e. only the **UiAuthClient** app client ID is accepted as a valid JWT audience.

The smoke-test workflow fetches a fresh ID token from a *different* Cognito app client — **SmokeTestClient** (`static_site_stack.py`, `add_client("SmokeTestClient", auth_flows=cognito.AuthFlow(user_password=True), ...)`, exported as `SmokeTestUserPoolClientId`).

Because the SmokeTestClient's client ID is not in `backend_authorizer`'s `user_pool_clients` list, API Gateway's JWT authorizer rejects the smoke-test token's audience claim and returns 401 for every Cognito-protected route (`/groups`, `/owners`, `/portfolio`, etc.) — only `/health`, `GET /config`, and `OPTIONS` (unauthenticated routes) succeed.

This is unrelated to the CORS OPTIONS fix in #3945 (commit c73f6c8c) — that fix is correct, but it surfaced/uncovered this separate pre-existing authorizer gap because the deploy's smoke test now gets further into the app before failing.

## How

[...]

PR #4257, opened 2026-06-14
fix: send Cognito ID token to API Gateway authorizer instead of backend JWT (#4256)

### LINKED ISSUE #4256: Bug: /owners returns 401 — frontend sends backend HS256 JWT but API Gateway authorizer requires a Cognito token
## What

The deployed app (`https://d32n5ua14kb82t.cloudfront.net/`) gets `401 Unauthorized` from `GET https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com/owners` (and, by the same mechanism, every other `/{proxy+}` route). The request is rejected at API Gateway before reaching the Lambda.

## Root cause (confirmed by code inspection)

There are two incompatible token mechanisms that collide on protected routes:

1. **API Gateway** protects `/` and `/{proxy+}` (ANY) with a Cognito JWT authorizer (`HttpUserPoolAuthorizer` "BackendCognitoAuthorizer"). It validates a **Cognito-issued JWT** (RS256, signed by the user pool's JWKS, `aud`/`client_id` = the UI app client ID) taken from `$request.header.Authorization`.
   - `cdk/stacks/backend_lambda_stack.py:373-382` (authorizer) and `:431-442` (routes behind it).
2. **The frontend**, after Cognito login, calls `POST /token/cognito` which returns a **backend-issued HS256 JWT** (`backend/app.py:200-217` → `auth.create_access_token`, HS256 in `backend/auth.py:154-161`). It stores that backend JWT via `setAuthToken(data.access_token)` (`frontend/src/main.tsx:461`) and attaches it as `Authorization: Bearer <backend-jwt>` on every API call, including `/owners` (`frontend/src/api.ts:220`).

API Gateway's Cognito authorizer cannot validate a backend HS256 JWT (wrong issuer, wrong signing key, not in the pool's JWKS), so it returns **401 before invoking the Lambda**. The Lambda's `DISABLE_AUTH=true` (`backend_lambda_stack.py:247`) never comes into play because the request never arrives.

Net effect: the `/token/cognito` → backend-JWT exchange is architecturally incompatible with the deployed gateway authorizer. Protected routes need a **Cognito-issued token** in the `Authorization` header, not the backend JWT.

## How (recommended fix — Option A)

Make the frontend send the **Cognito ID token** as the `Authorization: Bearer` header for API calls to gateway-protected routes, matching what the Cognito JWT authorizer validates (`aud` = UI client ID).

- In `frontend/src/main.tsx` / `frontend/src/api.ts`, when `awsUiAuth` is enabled, set the API auth token to the stored Cognito ID token (`getStoredCognitoIdToken()`) rather than the backend JWT from `/token/cognito`.
- The backend JWT exchange (`/token/cognito`) becomes unnecessary for the AWS-deployed path and should be removed from the bootstrap flow (or kept only for the non-gateway/local path) to avoid confusion. Document whichever decision is taken in `docs/AUTH.md`.
- Confirm token refresh: Cognito ID tokens expire (~1h). The existing `isJwtExpired` check (`main.tsx:441`) and the PKCE refresh path in `awsUiAuth.ts` must cover re-issuing the ID token before expiry so the `Authorization` header stays valid.

### Option B (only if a backend JWT is genuinely required end-to-end)
Replace the Cognito JWT authorizer with a Lambda REQUEST authorizer that validates the backend HS256 JWT, and stop setting `DISABLE_AUTH=true`. This is a larger change and loses the "gateway enforces auth before Lambda" property; Option A is preferred.

## Constraints

- Do not weaken the authorizer (no empty-audience entries — see existing #4027/#4047 guards in the stack).
- Keep CORS preflight (OPTIONS) on `HttpNoneAuthorizer` as-is.
- Update `docs/AUTH.md` — its current sequence diagram (`POST /token/cognito` → backend JWT → `GET /owners`) describes a flow that cannot work against the gateway authorizer and is itself part of the confusion.

## LLM tier

sonnet

## Success criteria

- After Cognito login on the deployed site, `GET /owners` returns 200 with a valid token.
- The `Authorization` header sent to `/owners` is a Cognito-issued JWT accepted by the API Gateway authorizer.
- `docs/AUTH.md` accurately reflects the working flow.
- Frontend + CDK tests cover the corrected token wiring.

## Failure modes to avoid

- Sending the Cognito **access** token where the authorizer expects `aud` (ID token) — verify which claim the authorizer matches and send the token whose claim matches the configured audience.
- Breaking the local/Google (`/token`) path, which legitimately uses the backend HS256 JWT against a FastAPI-enforced (non-gateway) setup.

leonarduk:
@
## What

Fixes the deployed app getting `401 Unauthorized` from `GET /owners` (and every other `/{proxy+}` route) after Cognito login. The request was rejected at API Gateway before reaching the Lambda.

## Root cause

API Gateway protects the routes with a Cognito JWT authorizer that validates a **Cognito-issued** token whose `aud` claim equals the UI app client ID. The frontend, however, exchanged the Cognito token for a **backend HS256 JWT** via `POST /token/cognito` and sent that as the `Authorization: Bearer` header. The gateway authorizer cannot validate a backend JWT (wrong issuer/signing key), so it returned 401 before the Lambda (`DISABLE_AUTH=true`) ever ran.

## Fix (Option A from the issue)

- `frontend/src/main.tsx`: replace `exchangeCognitoForBackendToken()` with `applyCognitoIdToken()`, which applies the stored Cognito **ID token** (whose `aud` matches the authorizer audience) directly as the API auth token via `setAuthToken`. No `POST /token/cognito` round-trip on the deployed path.
  - The helper no-ops when `awsUiAuth` is not the active provider (no `clientId`), so it never disturbs the **backend HS256 JWT** that the Google/local path legitimately uses against the FastAPI-enforced (non-gateway) setup.
- `docs/AUTH.md`: corrected sequence diagram, added an "ID token vs access token" note, a token-expiry/refresh section, and updated security notes.
- `cdk/stacks/backend_lambda_stack.py` + `cdk/tests/test_backend_lambda_stack.py`: refreshed now-stale comments/docstrings and added `test_backend_api_authorizer_accepts_cognito_id_token_contract` asserting the authorizer audience is the UI client ID taken from the `Authorization` header (the contract the frontend now relies on). No CDK functional change — the authorizer was already correct.

## Why the ID token, not the access token

The HTTP API JWT authorizer matches `aud` against `JwtConfiguration.Audience` (= `UiAuthUserPoolClientId`). Cognito sets `aud` on the ID token; access tokens carry `client_id` and have no `aud`, so the access token would fail authorization.

## Constraints honored

- Authorizer not weakened (no empty-audience entries; #4027/#4047 guards intact).
- CORS preflight (OPTIONS) stays on `HttpNoneAuthorizer`.
- Google `/token` (backend HS256 JWT) path preserved.

## Tests

- `frontend/tests/unit/main.cognitoExchange.test.ts` rewritten to assert the ID token is applied as the auth token and `/token/cognito` is never called.
- `frontend/tests/cognito-auth-flow.spec.ts` updated to assert the stored `authToken` and the `Authorization` header are the Cognito ID token, and that the deprecated backend exchange is never hit.
- CDK: 34 tests pass, including the new ID-token contract test.
- `npx tsc --noEmit` and `eslint` clean.

Closes #4256

@

## Release Notes

* **Improvements**
  * Streamlined AWS Cognito (hosted UI) authentication: the Cognito ID token is now sent directly to protected API routes via the `Authorization: Bearer` header, removing the intermediate token exchange.
  * Improved session robustness with hosted-UI refresh support ahead of token expiry, with safe logout if refresh or token application fails.

* **Documentation**
  * Updated authentication guidance and sequence details for Google vs AWS Cognito flows, including the updated token validation behaviour.

* **Tests**
  * Updated end-to-end and unit coverage to assert the new Cognito ID token authorisation contract and refresh behaviour.

### DIFF
--- a/cdk/stacks/backend_lambda_stack.py
+++ b/cdk/stacks/backend_lambda_stack.py
@@ -419,18 +419,22 @@ class BackendLambdaStack(Stack):
         # 401 before backend.auth.verify_google_token ever runs, the same class
         # of bug fixed for GET /config in #3873.
         #
-        # POST /token/cognito is NOT listed here: frontend/src/main.tsx
-        # exchangeCognitoForBackendToken sends a Cognito-issued access/ID token
-        # as a Bearer header, and backend_authorizer validates it successfully.
-        # This is not an assumption that the pools happen to match — it is the
-        # same pool/client by construction. ui_auth_user_pool_id_param and
+        # POST /token/cognito is NOT listed here: it stays behind the Cognito
+        # authorizer via the /{proxy+} catch-all. The deployed frontend no longer
+        # calls it — frontend/src/main.tsx applyCognitoIdToken now sends the
+        # Cognito ID token directly as the Bearer header on every protected route,
+        # which backend_authorizer validates (the ID token's `aud` matches the
+        # configured JwtConfiguration.Audience) — see issue #4256. Leaving
+        # /token/cognito authorizer-protected is correct: it requires the same
+        # valid Cognito token, and the route's pool/client match by construction.
+        # ui_auth_user_pool_id_param and
         # ui_auth_client_id_param (above) are documented as values exported by
         # StaticSiteStack's UiAuthUserPool/UiAuthClient (see
         # cdk/stacks/static_site_stack.py CfnOutputs UiAuthUserPoolId and
         # UiAuthUserPoolClientId). That same UiAuthClient.user_pool_client_id is
         # embedded in config.json's awsUiAuth.clientId (static_site_stack.py,
         # DeployRuntimeConfig), which is what the frontend uses to sign in via
-        # Cognito before calling /token/cognito. So backend_authorizer and the
+        # Cognito and is also the ID token's `aud`. So backend_authorizer and the
         # frontend's Cognito sign-in resolve to the same UiAuthUserPool /
         # UiAuthClient — there is only one Cognito user pool in this stack
         # pairing, not two.
--- a/cdk/tests/test_backend_lambda_stack.py
+++ b/cdk/tests/test_backend_lambda_stack.py
@@ -596,6 +596,27 @@ def test_backend_api_authorizer_audience_excludes_empty_smoke_test_client(templa
     assert without_smoke_client == [{"Ref": "UiAuthUserPoolClientId"}]
 
 
+def test_backend_api_authorizer_accepts_cognito_id_token_contract(template):
+    """The authorizer audience is the UI app client ID, taken from the
+    Authorization header. This is the contract the deployed frontend now relies
+    on: frontend/src/main.tsx applyCognitoIdToken sends the Cognito ID token
+    (whose `aud` claim equals the UI client ID) as `Authorization: Bearer`, so
+    the gateway authorizer admits it before invoking the Lambda (#4256). If the
+    audience ever stopped including UiAuthUserPoolClientId, the ID token would be
+    rejected with 401 and this test would catch the regression."""
+    resources = template.find_resources("AWS::ApiGatewayV2::Authorizer")
+    authorizer = next(iter(resources.values()))
+    properties = authorizer["Properties"]
+
+    assert properties["IdentitySource"] == ["$request.header.Authorization"]
+
+    audience = properties["JwtConfiguration"]["Audience"]
+    _, with_smoke_client, without_smoke_client = audience["Fn::If"]
+    # The UI client ID (the ID token's `aud`) must be present in both branches.
+    assert {"Ref": "UiAuthUserPoolClientId"} in with_smoke_client
+    assert {"Ref": "UiAuthUserPoolClientId"} in without_smoke_client
+
+
 def test_backend_api_routes_require_cognito_authorizer(template):
     """All API Gateway routes must require Cognito JWT authorization except
     /health, GET /config, POST /token/google, and the CORS preflight OPTIONS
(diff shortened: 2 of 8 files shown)


leonarduk:
@
## Addressed the review — silent in-session refresh implemented (d39bf26f)

Thanks for the review. The blocking point about token refresh was valid and is now implemented.

**One factual correction on the "regression" framing:** the old flow exchanged for a backend HS256 JWT with a **15-minute** lifetime (`DEFAULT_TOKEN_EXPIRE_MINUTES = 15` in `backend/auth.py`), and `exchangeCognitoForBackendToken` also only ran at **bootstrap** — so the previous code had *no* in-session refresh either, and its token was actually shorter-lived than the Cognito ID token (~1h). So this is not a regression; but the AC does ask for refresh, and neither old nor new code provided it, so it was a real gap worth closing.

**What I added:**
- `awsUiAuth.ts`: persist the `refresh_token` from the authorization-code grant; new `refreshCognitoSession()` (POST `grant_type=refresh_token` to `/oauth2/token`, preserving the refresh token Cognito does not re-issue) and `getCognitoSessionExpiresAt()`. No `offline_access` scope needed — Cognito returns a refresh token for the code grant by default.
- `main.tsx`: `scheduleCognitoRefresh()` arms a timer ~5 min before expiry to renew the ID token, re-apply it via `setAuthToken`, and re-arm. On refresh failure it clears the session and logs out so the next load restarts the hosted-UI login.

**Tests added** (491 frontend tests green):
- Silent refresh success (asserts `grant_type=refresh_token` and the refreshed token is re-applied).
- Scheduled-refresh failure → session cleared + logout.
- **The `setAuthToken`-throws edge case you flagged** (storage unavailable → session clear + logout).
- `refresh_token` persistence on code exchange, plus direct unit tests for `refreshCognitoSession` and `getCognitoSessionExpiresAt`.

**On the `isJwtExpired` follow-up:** confirmed it was local to `main.tsx` and not exported/duplicated anywhere (`grep` clean), so its removal is complete.
@

PR #4273, opened 2026-06-14
fix(awsUiAuth): preserve non-auth query params on OAuth callback failure

### LINKED ISSUE #3965: `fix(awsUiAuth): consider preserving non-auth query params when stripping code/state on token-exchange failure`
### What
In `awsUiAuth.ts`, the token-exchange error handler (line ~162) currently calls:
```ts
window.history.replaceState({}, document.title, window.location.pathname)
```

This strips **all** query parameters, including any legitimate non-auth query params the application may be carrying (e.g., `?redirect=/dashboard&locale=en`). Only `code` and `state` should be removed.

The same issue exists in the state-mismatch branch, which mirrors this pattern.

### Why
- **Correctness**: Legitimate app state encoded in query params gets lost when authentication fails, degrading UX (user loses context about where they were trying to go).
- **Consistency**: The auth callback flow should be minimally invasive—strip only OAuth artifacts, preserve everything else.
- **Agent confusion**: The current implementation conflates "strip auth params" with "strip all params", which is a semantic leap that could mislead future maintainers or code-generation tools.

### How
1. Extract `code` and `state` from `window.location.search` before calling `replaceState`.
2. Reconstruct the query string without those two parameters.
3. Pass the cleaned URL to `replaceState`:
   ```ts
   const params = new URLSearchParams(window.location.search);
   params.delete('code');
   params.delete('state');
   const cleanedSearch = params.toString() ? `?${params.toString()}` : '';
   window.history.replaceState(
     {},
     document.title,
     `${window.location.pathname}${cleanedSearch}`
   );
   ```
4. Apply the same fix to the state-mismatch branch for parity.

### Constraints
- Must not break existing error-handling flow or retry button behavior (PR #3937 acceptance criteria remain satisfied).
- Must not change the user-facing error message.
- The fix applies only to the OAuth callback path; non-callback routes are out of scope.
- Backward compatibility: if no non-auth params are present, behavior is identical to current code.

### LLM Tier
**Sonnet** — Requires:
- Understanding of URLSearchParams and query-param manipulation
- Design judgment on where to apply the fix (token-exchange *and* state-mismatch branches for parity)
- Test updates that assert both `code`/`state` removal *and* preservation of other params
- Verification that existing tests still pass

### Success looks like
- [ ] `awsUiAuth.ts` token-exchange error handler removes only `code` and `state`, preserving other query params
- [ ] State-mismatch branch updated identically for consistency
- [ ] `awsUiAuth.test.ts` includes test cases asserting:
  - `code` and `state` are stripped
  - Other query params (e.g., `?redirect=X&locale=Y`) are preserved
  - Empty search string when no params remain
- [ ] All existing tests pass; no regressions in PR #3937 acceptance criteria
- [ ] Code comment clarifies intent: "strip OAuth callback params only"

### Failure looks like
- Legitimate query params are still lost after token-exchange failure.
- State-mismatch branch left unchanged (inconsistent behavior).
- Test assertions only verify removal of `code`/`state` but don't verify preservation of other params.
- Existing error retry flow breaks.

---

_Follow-up from AI review of PR #3937._

leonarduk:
## Summary
- `exchangeCode`'s state-mismatch and token-exchange-failure error handlers previously cleared the URL via `replaceState(..., pathname)`, stripping *all* query params.
- Both branches now strip only the OAuth `code`/`state` params via a new `stripAuthCallbackParams` helper, preserving any other app query params (e.g. `?redirect=/dashboard&locale=en`).
- The `access_denied`/error branch is left as-is (full strip), since leaving `error=...` in the URL would cause an infinite retry loop on reload — that branch is also out of scope per the issue.

## Test plan
- [x] `npx vitest run tests/unit/awsUiAuth.test.ts` — 35 passed, including 2 new tests asserting non-auth params are preserved on state-mismatch and token-exchange-failure, and existing tests confirm code/state are still stripped and search becomes empty when nothing else remains.
- [x] `npm run lint` clean
- [x] `tsc --noEmit` clean

Closes #3965

## Summary by CodeRabbit

* **New Features**
  * OAuth authentication callback now automatically removes temporary authentication parameters from the browser URL, providing cleaner history and improved user experience

* **Bug Fixes**
  * Improved error handling for OAuth authentication failures with more consistent error messaging across different failure scenarios
  * Optimised the authentication flow by eliminating redundant URL parameter cleanup during token exchange

### DIFF
--- a/frontend/src/awsUiAuth.ts
+++ b/frontend/src/awsUiAuth.ts
@@ -55,6 +55,15 @@ const normaliseDomain = (domain: string) => {
   return trimmed.startsWith('https://') ? trimmed : `https://${trimmed}`;
 };
 
+// Strip OAuth callback params (code/state) only, preserving any other query params.
+const stripAuthCallbackParams = () => {
+  const params = new URLSearchParams(window.location.search);
+  params.delete('code');
+  params.delete('state');
+  const search = params.toString() ? `?${params.toString()}` : '';
+  window.history.replaceState({}, document.title, `${window.location.pathname}${search}`);
+};
+
 const redirectUri = (redirectPath?: string) => {
   const path = redirectPath?.startsWith('/')
     ? redirectPath
@@ -190,6 +199,8 @@ const exchangeCode = async (config: AuthConfig) => {
   if (errorParam === 'access_denied') {
     window.sessionStorage.removeItem(STATE_KEY);
     window.sessionStorage.removeItem(VERIFIER_KEY);
+    // Strip the whole query string (not just code/state) so retrying via
+    // reload doesn't immediately re-trigger this same `error` param.
     window.history.replaceState({}, document.title, window.location.pathname);
     throw new UserCancelledError();
   }
@@ -201,7 +212,7 @@ const exchangeCode = async (config: AuthConfig) => {
   if (!expectedState || !verifier || state !== expectedState) {
     window.sessionStorage.removeItem(STATE_KEY);
     window.sessionStorage.removeItem(VERIFIER_KEY);
-    window.history.replaceState({}, document.title, window.location.pathname);
+    stripAuthCallbackParams();
     throw new Error('Invalid AWS UI authentication callback state');
   }
   // State matched — consume the one-time PKCE credentials before the fetch.
@@ -223,7 +234,7 @@ const exchangeCode = async (config: AuthConfig) => {
   if (!response.ok) {
     // The authorization code is single-use; strip it from the URL so a
     // reload restarts the hosted-UI flow instead of replaying a dead code.
-    window.history.replaceState({}, document.title, window.location.pathname);
+    stripAuthCallbackParams();
     throw new Error('AWS UI authentication token exchange failed');
   }
   storeSession(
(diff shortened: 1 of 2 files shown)


leonarduk:
Thanks for the review — the flagged items are already satisfied, just not visible in the latest commit's diff because they were merged in from \`main\` earlier in this PR's history (via PR #4274/#4279 merges) before they were later refined.

Looking at the full PR diff vs \`main\` (\`gh pr diff 4273\`), \`frontend/src/awsUiAuth.ts\` currently:

- Defines \`stripAuthCallbackParams(keys = ['code', 'state'])\` (line 76) which preserves all other query params.
- Uses it in the **state-mismatch branch** (line 249: \`stripAuthCallbackParams(); throw new Error('Invalid AWS UI authentication callback state')\`).
- Uses it in the **token-exchange-failure branch** (line 271).
- Uses it in the **access_denied / generic OAuth error branch** (line 238: \`stripAuthCallbackParams(['code', 'state', 'error'])\`).

\`error\` is included in the strip list specifically because leaving \`?error=access_denied\` in the URL would cause an infinite retry loop on reload (the retry button calls \`window.location.reload()\`) — this was the fix for an earlier CodeQL \`js/user-controlled-bypass\` alert on this same PR.

\`frontend/tests/unit/awsUiAuth.test.ts\` (42 tests, all passing) includes dedicated "preserves non-auth query params when cleaning up after ..." tests for all three branches (access_denied, state mismatch, token exchange failure), plus the existing tests confirming \`code\`/\`state\` are stripped and the URL collapses to just the pathname when nothing else remains.

No further changes needed for these points — let me know if there's a different gap you're seeing.

Issue #4352, opened 2026-06-15
Admin approval flow: provision the user and notify them their login is ready


leonarduk:
# Admin approval flow: provision the user and notify them their login is ready

## What
Implement the admin-facing approval action for a pending signup request (from #4351). Approving a request provisions the user — adds their email to the login allowlist and scaffolds their owner data — then emails the user that their login is ready. Rejecting marks the request closed (optionally notifying the user).

## Why
Capturing a request (#4351) is only useful if the admin can act on it without hand-editing JSON. This closes the loop: visitor requests access → admin approves with one action → user is provisioned and told they can log in. It is the core of the "easy to get started" objective.

## How
- Add an approval handler (e.g. `POST /signup/approve` / `POST /signup/reject`) that validates the unguessable token from the #4351 email and loads the corresponding pending request.
- On approve: provision the owner using the existing scaffold path — `ensure_owner_scaffold()` in `backend/common/compliance.py` and/or `AccountsStore.ensure_owner()` in `backend/common/accounts_store.py` — and ensure the user's email becomes part of `_allowed_emails()` (write `person.json` with the email so `backend/auth.py` admits them).
- Mark the request as approved/rejected so the token cannot be reused.
- Email the user (reuse the SES pattern from `backend/emails/weekly_report.py`) confirming their login is ready, with a link to the login page.

## Constraints
- Depends on #4351 (pending-request store + tokens) and #4353 (owner/account provisioning); reuse `ensure_owner_scaffold()` / `ensure_owner()` rather than hand-writing JSON. Note existing issue #4325: account *type* creation is otherwise implicit on first write.
- Approval must be idempotent and single-use: a consumed/expired token must not re-provision or error confusingly.
- Mock SES and the data store in tests; cover approve, reject, invalid-token, and already-consumed cases under `tests/`.
- Keep provisioning logic in a service/domain function, not inline in the route.

## LLM tier
**Opus** — security-sensitive (grants login access), spans auth allowlist + scaffold + email, and needs idempotent token handling.

## Success looks like
- [ ] Approving a valid request provisions the owner and adds the email to the login allowlist so the user can authenticate.
- [ ] The user receives a "login ready" email (mocked in tests) after approval.
- [ ] Tokens are single-use: re-using or using an invalid/expired token is safely rejected.
- [ ] pytest covers approve, reject, invalid-token, and already-consumed paths with SES/store mocked.

## Failure looks like
- An approved user still cannot log in (allowlist not updated).
[...]

Issue #4350, opened 2026-06-15
Public "Create Account" request page


leonarduk:
# Public "Create Account" request page

## What
Add a public, unauthenticated "Create Account" page in the frontend where a prospective user can request access. The form collects at least full name and email (optionally a short note / intended use), validates input client-side, submits the request to the new backend signup endpoint (#4351), and shows a clear "request received, pending admin approval" confirmation state.

## Why
The app is currently invite-only: a user can only log in if their email already exists in `data/accounts/*/person.json` (`_allowed_emails()` in `backend/auth.py`). There is no self-service way for a new person to ask for access, so onboarding requires manual JSON editing by the maintainer. A public request page is the first step of a friction-free getting-started flow.

## How
- Add a new route/page (e.g. `frontend/src/CreateAccountPage.tsx`) reachable from `frontend/src/LoginPage.tsx` via a "Create account" link.
- Build a controlled form (name, email, optional note) with client-side validation (required fields, email format).
- On submit, POST to the signup endpoint defined in #4351; disable the button while in-flight and surface success/error states.
- Keep styling consistent with existing pages (reuse existing form/button components and conventions).

## Constraints
- Depends on the backend endpoint in #4351 for the actual submission; until that exists, wire to the agreed request/response shape and cover the UI with mocked requests.
- Must remain accessible without authentication (no token required) and must not leak whether an email is already registered (avoid account enumeration).
- Follow existing TypeScript/ESLint/Prettier conventions; add Vitest coverage for validation and submit states.

## LLM tier
**Sonnet** — standard self-contained React form with validation and async submit; no auth/security design decisions beyond avoiding enumeration.

## Success looks like
- [ ] A public `/create-account` page is linked from the login page and loads without authentication.
- [ ] Submitting valid details calls the #4351 endpoint and shows a "pending approval" confirmation.
- [ ] Invalid/empty input is blocked client-side with clear messages, and the success message does not reveal whether the email already exists.
- [ ] Vitest covers validation and the submit success/error paths.

## Failure looks like
- The page requires login or 404s for anonymous visitors.
- The form reveals that an email is already registered (enumeration).
- Submission silently fails with no user feedback or unhandled promise rejection.

PR #4358, opened 2026-06-15
feat(backend): public signup-request endpoint with admin SES notification

### LINKED ISSUE #4351: Backend: account-request endpoint that emails the admin for approval
# Backend: account-request endpoint that emails the admin for approval

## What
Add a public backend endpoint that accepts a signup request (name, email, optional note), persists it as a pending request, and emails the admin via SES with the request details and approve/reject actions. This is the server side of the public "Create Account" page (#4350).

## Why
There is currently no way to capture an access request or notify the maintainer. Onboarding requires hand-editing `data/accounts/*/person.json`. An endpoint that records the request and notifies the admin is the bridge between a visitor asking for access and the approval flow (#4352).

## How
- Add a public route (e.g. `POST /signup/request`) in a new or existing router; no auth required.
- Validate and normalise input (trim, lowercase email, basic email format). Reject obviously invalid payloads with 400.
- Persist the pending request to a known location (e.g. a `data/signup_requests/` store or the writable accounts store) with a generated request id and timestamp.
- Send an admin notification email reusing the existing SES pattern from `backend/emails/weekly_report.py` (`boto3.client("ses")`, `WEEKLY_REPORT_FROM` for the sender). Introduce an admin-recipient env var (e.g. `SIGNUP_ADMIN_EMAIL`). The email should contain the request details and approve/reject links/tokens consumed by #4352.
- Always return a generic success response so the endpoint cannot be used to enumerate existing accounts.

## Constraints
- Depends on / is consumed by #4350 (frontend) and #4352 (approval flow); agree the request/response and approval-token shape with #4352.
- Mock SES in tests — do not depend on live AWS. Add pytest coverage under `tests/`.
- Do not echo back whether the email already has an account; keep domain logic out of the route where practical.
- Approve/reject tokens must be unguessable and ideally single-use/expiring to prevent unauthorised provisioning.

## LLM tier
**Opus** — touches auth/security surface (public endpoint, anti-enumeration, signed approval tokens) and a new SES email path; warrants careful design.

## Success looks like
- [ ] `POST /signup/request` accepts a valid payload, stores a pending request with id + timestamp, and returns a generic success response.
- [ ] An admin notification email is sent via SES (mocked in tests) containing the request details and approve/reject actions.
- [ ] Invalid input returns 400 and existing-email submissions are indistinguishable from new ones (no enumeration).
- [ ] Approval tokens are unguessable; pytest covers the happy path, validation, and the mocked email send.

## Failure looks like
- The endpoint requires authentication or leaks account existence.
- Tests hit live SES/AWS, or the email send failure is swallowed silently.
- Approval links are guessable/replayable, allowing unauthorised provisioning.

leonarduk:
## What

Adds `POST /signup/request` — a public, unauthenticated endpoint that records a visitor's access request as a pending request and emails the admin (via SES) the request details with approve/reject links. This is the server side of the public "Create Account" page (#4350) and feeds the approval flow (#4352).

## How

- **`backend/common/signup_requests.py`** — domain logic kept out of the route: payload validation/normalisation (trim, lowercase email, basic email format, length caps), request persistence, and approval-token generation.
  - Token = `secrets.token_urlsafe(32)` (256-bit, unguessable). **Only the SHA‑256 hash is persisted**; the plaintext is returned once for the email and never stored, so a leak of the request store cannot reconstruct a usable link.
  - Each record carries `status: "pending"`, `created_at`, and `expires_at` (7‑day TTL). #4352 verifies an inbound token by hashing it and comparing to `token_sha256` while `status == "pending"` and unexpired → single‑use, time‑bound semantics.
- **`backend/emails/signup_request.py`** — admin notification reusing the `weekly_report` SES pattern (`boto3.client("ses")`, `WEEKLY_REPORT_FROM` sender). New recipient env var **`SIGNUP_ADMIN_EMAIL`**. All visitor input is HTML‑escaped via `markupsafe` before going into the email body.
- **`backend/routes/signup.py`** — anti‑enumeration handler: performs **no** account lookup and always returns the same generic success body. `400` on malformed payload, `503` when `SIGNUP_ADMIN_EMAIL` is unconfigured, `502` when SES delivery fails (logged, **never swallowed**). Links carry `id` + `token` under an optional `SIGNUP_APPROVAL_BASE_URL`.
- Registered the router (public, no auth dependency) in `backend/bootstrap/routers.py`.

## Token / response shape (to agree with #4352)

- Request: `{ "name": str, "email": str, "note"?: str }`
- Response (always): `{ "status": "ok", "message": "If your request is valid, an administrator has been notified." }`
- Approval link: `<SIGNUP_APPROVAL_BASE_URL>/signup/approve?id=<id>&token=<token>` (and `/signup/reject`)
- Stored record: `id, name, email, note, status, created_at, expires_at, token_sha256`

## Tests (21, all passing — SES mocked, no live AWS)

- Validation/normalisation, token uniqueness, hash-only persistence.
- Email rendering (details + links present, hostile input escaped) and SES invocation.
- Route: happy path persists + notifies; invalid payload → 400; existing vs new emails indistinguishable; missing admin email → 503; SES failure → 502 (not swallowed).

## Notes / follow-ups

- Coordinate the token/response shape above with #4352 before that PR consumes it.
- Requests are written under `<accounts_root>/../signup_requests/` (writable store) via `resolve_accounts_root`.

Closes #4351

## Summary by CodeRabbit

* **New Features**
  * Added a public account-signup request endpoint with admin approval/rejection links
  * Generates time-limited, single-use approval tokens for pending requests
  * Added admin email notifications for signup requests
  * Protects against account enumeration by returning the same response for valid/duplicate requests
* **Tests**
  * Added comprehensive integration and unit tests covering validation, persistence, email rendering/sending, and error handling

### DIFF
--- a/backend/bootstrap/routers.py
+++ b/backend/bootstrap/routers.py
@@ -36,6 +36,7 @@ from backend.routes.rebalance import router as rebalance_router
 from backend.routes.reports import router as reports_router
 from backend.routes.scenario import router as scenario_router
 from backend.routes.screener import router as screener_router
+from backend.routes.signup import router as signup_router
 from backend.routes.support import router as support_router
 from backend.routes.tax import router as tax_router
 from backend.routes.timeseries_admin import router as timeseries_admin_router
@@ -70,6 +71,7 @@ def register_routers(app: FastAPI, cfg: Config) -> None:
     app.include_router(trail_router, dependencies=protected)
     app.include_router(compliance_router)
     app.include_router(screener_router)
+    app.include_router(signup_router)
     app.include_router(support_router)
     app.include_router(query_router, dependencies=protected)
     app.include_router(virtual_portfolio_router, dependencies=protected)
--- a/backend/common/signup_requests.py
+++ b/backend/common/signup_requests.py
@@ -1,0 +1,134 @@
+"""Domain logic for public account-signup requests.
+
+Keeps validation, token generation, and persistence out of the route handler
+(see :mod:`backend.routes.signup`). A signup request records a visitor's
+interest in an account and produces an unguessable, expiring approval token.
+
+Only the SHA-256 hash of the approval token is persisted; the plaintext token
+is returned once (to be embedded in the admin notification email) and never
+stored, so a leak of the request store cannot reconstruct a usable link. The
+approval flow (#4352) verifies an inbound token by hashing it and comparing it
+to the stored ``token_sha256`` while ``status == "pending"`` and the request
+has not expired, giving single-use, time-bound semantics.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import re
+import secrets
+from dataclasses import asdict, dataclass
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+# Deliberately permissive: we only guard against obviously malformed input.
+# RFC-complete email validation is intractable and not required here — the
+# real check is that a human at the address can act on the admin notification.
+_EMAIL_RE = re.compile(r"^[^@\s]+@[^@\s]+\.[^@\s]+$")
+
+_TOKEN_TTL = timedelta(days=7)
+_MAX_NAME_LEN = 200
+_MAX_EMAIL_LEN = 320  # RFC 5321 maximum length of a forward-path address
+_MAX_NOTE_LEN = 2000
+
+
+class SignupValidationError(ValueError):
+    """Raised when a signup request payload fails validation."""
+
+
+@dataclass(frozen=True)
+class SignupRequest:
+    """A persisted pending signup request.
+
+    ``token_sha256`` is the hash of the single-use approval token; the
+    plaintext token is never stored on this record.
+    """
+
+    id: str
+    name: str
+    email: str
+    note: str
+    status: str
+    created_at: str
+    expires_at: str
+    token_sha256: str
+
+
+def normalise_payload(payload: object) -> tuple[str, str, str]:
+    """Validate and normalise a raw signup payload.
+
+    Returns ``(name, email, note)`` with the email lowercased and all fields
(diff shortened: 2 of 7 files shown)


leonarduk:
Thanks for the reviews. Pushed `97511fab` addressing the genuine findings:

**Fixed**
- **CodeQL High — ReDoS in the email regex.** Replaced `_EMAIL_RE` with a linear, string-based `_looks_like_email()` (single `@`, non-empty local/domain, dot in domain, no spaces, no leading/trailing/consecutive dots). Added a test that feeds the polynomial-backtracking payload (`!@!.!.…`) and asserts it's rejected fast.
- **DeepSeek — unset `SIGNUP_APPROVAL_BASE_URL` produces broken links silently.** The endpoint now logs a `WARNING` when the base URL is unset (links are then relative), so the misconfiguration isn't silent. Covered by a new test. The request id + token are also in the email body so the admin can still act manually.

**Follow-up (out of scope here)**
- **Rate limiting / abuse protection** — valid, but there's no existing throttling middleware in the backend to extend, and adding it would expand this endpoint PR considerably. Filed as #4364 (milestone 8).

**Respectfully pushing back**
- **`_store_dir` using `accounts_root.parent`** — `resolve_accounts_root()` returns the accounts *directory* (e.g. `…/data/accounts`), never a `person.json` file path, so `.parent` is the data root and `signup_requests/` lands as a sibling of `accounts/` — exactly the `data/signup_requests/` location the issue specifies. The premise that it could return `…/accounts/family1/person.json` doesn't hold; see [`backend/routes/_accounts.py`](backend/routes/_accounts.py).
- **Token in the URL query string** — this is the natural shape for emailed approve/reject links (you can't POST from a mailto link). It's mitigated by 256-bit entropy, hash-only storage, single-use, and a 7-day expiry. The final transport is owned by the approval flow in #4352; happy to revisit there (e.g. an HMAC landing page that POSTs) if preferred.

All 25 signup tests pass locally; `ruff` clean, `black` applied.

PR #4404, opened 2026-06-15
feat: add rebuild_portfolio to both stores (issue #4314)

### LINKED ISSUE #4314: `_rebuild_portfolio()` silently skips for S3-backed stores — portfolio impact tracking may be stale for deployed manual holdings
Here is a complete, actionable GitHub issue body in Markdown:

---

### What

In `backend/transactions.py`, the function `_rebuild_portfolio()` (around line 386-389) returns early when `local_root` is `None`. This means that when the store is an `S3AccountsStore` (which has no local root), the portfolio rebuild is silently skipped. The relevant code path is:

```python
def _rebuild_portfolio(...):
    local_root, _ = _resolve_local_root()
    if local_root is None:
        return  # <-- silently skips for S3-backed stores
    ...
```

### Why

When manual holdings are created in a deployed (S3-backed) environment, `_rebuild_portfolio()` is never called. This means:

- `_PORTFOLIO_IMPACT` tracking is not updated
- `rebuild_account_holdings()` downstream effects are lost
- Portfolio data for users creating manual holdings in AWS may become stale

This is a correctness risk: users who create manual holdings via the deployed API will not see those holdings reflected in portfolio impact calculations or aggregated portfolio views that depend on the rebuild logic.

### How

1. **Detect the store type** in `_rebuild_portfolio()` — instead of checking `local_root is None`, check whether the active store is an `S3AccountsStore` (or more generally, whether portfolio rebuild is supported for the current store).
2. **Implement S3-aware portfolio rebuild** — either:
   - Add a `rebuild_portfolio()` method to `S3AccountsStore` that performs the equivalent logic against S3, or
   - Refactor the rebuild logic to accept a store instance and work generically with the store's read/write interface.
3. **Remove the silent skip** — replace the early return with a call to the appropriate rebuild method for the active store.

The simplest concrete approach: add an abstract method `rebuild_portfolio(owner, account, holdings)` to the `AccountsStore` base class, implement it in `LocalAccountsStore` (existing logic) and `S3AccountsStore` (S3-equivalent logic), and call it polymorphically from `_rebuild_portfolio()`.

### Constraints

- **Must not break** local development flow — `LocalAccountsStore` behavior must remain identical
- **Must not break** existing S3 reads — portfolio data must remain readable via the existing S3 data loader
- **Out of scope** — full integration test for the S3 write-then-read cycle (separate follow-up issue)
- **Out of scope** — changing the `_resolve_local_root()` return semantics (separate follow-up issue)

### LLM tier

**Sonnet** — This is a moderate design task requiring:
- Understanding the store abstraction and where to add the new method
- Implementing S3-specific portfolio rebuild logic (non-trivial but well-scoped)
- Ensuring backward compatibility for both store types
- Multi-file changes (base class, both store implementations, the calling code)

### Success looks like

1. Creating a manual holding via `POST /holdings/manual` in a deployed (S3-backed) environment triggers a portfolio rebuild
2. Portfolio impact tracking (`_PORTFOLIO_IMPACT`) is updated correctly for S3-backed stores
3. All existing local-dev tests continue to pass unchanged
4. No silent skips remain in `_rebuild_portfolio()`

### Failure looks like

1. Portfolio rebuild still silently skipped for S3 stores
2. Local portfolio rebuild behavior changes unexpectedly
3. S3 portfolio rebuild throws errors or produces incorrect data
4. The change introduces new dependencies or breaks the store abstraction

---

_Follow-up from AI review of PR #4302._

leonarduk:
## Summary

Adds `rebuild_portfolio()` method to both `LocalAccountsStore` and `S3AccountsStore`, enabling the S3-backed store to rebuild holdings from transactions without needing the path-based portfolio rebuild that only works with on-disk roots.

## Changes

**`backend/common/accounts_store.py`:**
- `LocalAccountsStore.rebuild_portfolio()` — delegates to `portfolio_loader.rebuild_account_holdings` and `portfolio_mod.build_owner_portfolio` using the local disk root
- `S3AccountsStore.rebuild_portfolio()` — reads the transaction document from S3, computes holdings via `portfolio_loader.compute_holdings_from_transactions()`, and writes the resulting `<account>.json` back to S3. Warns and returns early when no transaction document exists.

**`backend/common/portfolio_loader.py`:**
- New `compute_holdings_from_transactions()` helper for rebuilding holdings purely from a transactions dict (no filesystem dependency)

**`backend/routes/transactions.py`:**
- Replaces the old `_rebuild_portfolio` function with `store.rebuild_portfolio(owner, account)`

**Tests:**
- `test_local_store_rebuild_portfolio` — verifies on-disk holdings file is created from transactions
- `test_s3_store_rebuild_portfolio` — verifies S3 holdings object created, including auto-generated CASH.GBP from DEPOSIT transactions
- `test_s3_store_rebuild_portfolio_missing_transactions` — warns when no transaction document
- `test_local_store_rebuild_portfolio_none_root` — warns when root is None

Closes #4314

## Summary by CodeRabbit

* **New Features**
  * Added portfolio rebuild support for both local and cloud storage, regenerating holdings from transaction data and writing updated holdings outputs when available.

* **Improvements**
  * Unified holdings calculation via a shared computation helper, including cash-like holdings, unit scaling, and acquisition/currency handling for more consistent results.
  * Simplified rebuild triggering by delegating rebuilding to the active storage implementation.

* **Tests**
  * Added coverage for local/cloud rebuild flows and the shared holdings computation helper, including missing/empty transaction scenarios.


leonarduk (commented):
## Addressing the DeepSeek AI review

All three claimed blocking issues are incorrect when checked against the actual code:

### 1. "`_put_document` is not defined" — ❌ False

`_put_document` is defined at **line 285** of `backend/common/accounts_store.py` as part of the pre-existing `S3AccountsStore` class. It was not added in this PR — it has existed since the store abstraction was introduced.

Verified:
```python
>>> from backend.common.accounts_store import S3AccountsStore
>>> hasattr(S3AccountsStore(bucket='x'), '_put_document')
True
```

### 2. "`TYPE_SIGN`/`CASH_SIGNS` defined after the loop" — ❌ False

In `compute_holdings_from_transactions()` (lines 195-214 of `portfolio_loader.py`), the dictionaries are defined **before** the `for` loop:

```python
    TYPE_SIGN = {...}      # lines 195-202
    CASH_SIGNS = {...}     # lines 203-208
    SHARE_SCALE = 10**8    # line 209
    ledger: defaultdict[str, float] = defaultdict(float)  # line 211
    acquisition: dict[str, str] = {}  # line 212

    for t in tx_data.get("transactions", []):  # line 214
```

All tests pass, including `test_compute_holdings_from_transactions` and `test_compute_holdings_from_transactions_empty`, confirming no runtime `NameError`.

### 3. "S3 `rebuild_portfolio` does not update portfolio impact tracking" — ❌ Incorrect premise

`_PORTFOLIO_IMPACT` is **never** updated by `_rebuild_portfolio` or `build_owner_portfolio`. It is updated by the **route handlers** in `transactions.py` (lines 541, 618, 620, 621, 659) **before** they call `_rebuild_portfolio`.

`build_owner_portfolio` only **reads** `_PORTFOLIO_IMPACT` (line 192 of `portfolio.py`) to adjust in-memory valuations — it does not write to it.

The actual flow:
1. Route handler updates `_PORTFOLIO_IMPACT` directly (e.g. `_PORTFOLIO_IMPACT[owner] += impact`)
2. Route handler calls `_rebuild_portfolio(owner, account, store)` → now calls `store.rebuild_portfolio()`
3. `store.rebuild_portfolio()` rebuilds holdings from transactions

The portfolio impact tracking works identically for both store types because **step 1 happens regardless of store type**, and **step 2 now runs for S3 stores too** (where previously the `local_root is None` guard silently skipped it).

---

**Verdict: All three blocking issues are false positives.** The PR is ready for review.

leonarduk:
Closing: all changes from #4314 are already merged into main. Both LocalAccountsStore.rebuild_portfolio and S3AccountsStore.rebuild_portfolio exist on main, and _rebuild_portfolio in transactions.py already delegates to store.rebuild_portfolio().

Issue #3287, opened 2026-06-03
Deploy workflow fails at "CDK diff": BackendLambdaStack change-set diff requires UiAuthUserPoolId/UiAuthUserPoolClientId params that don't exist yet


leonarduk:
## Summary

The **Deploy Infrastructure** workflow (`.github/workflows/deploy-lambda.yml`) fails every run at the **CDK diff** step. Multiple successive fixes have been attempted (most recently `34743ea2`, which changed `--method=changeset` → `--method=change-set`) but none address the actual root cause, so each new run fails again.

Failing run: https://github.com/leonarduk/allotmint/actions/runs/26868411197/job/79251205759

## Exact error

```
Hold on while we create a read-only change set to get a diff with accurate replacement information
(use --method=template to use a less accurate but faster template-only diff)
Could not create a change set, and '--method=change-set' was specified.
Please check your permissions or use '--method=auto' to allow falling back to a template diff.
The following CloudFormation Parameters are missing a value: UiAuthUserPoolId, UiAuthUserPoolClientId
##[error]Process completed with exit code 1.
```

## Root cause

The failing step is:

```yaml
# .github/workflows/deploy-lambda.yml:221-228
- name: CDK diff
  run: npx cdk diff BackendLambdaStack StaticSiteStack --method=change-set -c prod=true
  working-directory: cdk
```

`--method=change-set` makes CDK create a **real CloudFormation change set** to compute an accurate diff. Creating a change set requires every stack parameter to have a value.

`BackendLambdaStack` declares two **required** CfnParameters with no default:

```python
# cdk/stacks/backend_lambda_stack.py:299-321
ui_auth_user_pool_id_param = CfnParameter(self, "UiAuthUserPoolId", type="String",
    allowed_pattern=".+", ...)        # default only set if UI_AUTH_USER_POOL_ID env / context present
ui_auth_client_id_param = CfnParameter(self, "UiAuthUserPoolClientId", type="String",
    allowed_pattern=".+", ...)        # same
```

`allowed_pattern=".+"` makes a value mandatory. These values are produced by **StaticSiteStack** (its `UiAuthUserPoolId` / `UiAuthUserPoolClientId` outputs, `cdk/stacks/static_site_stack.py:439-440`) and are only fetched **after** the diff, in the *Get Cognito UI auth outputs* step, then passed into the *Deploy BackendLambdaStack* step:

```yaml
# .github/workflows/deploy-lambda.yml:270-272 (a LATER step)
[...]

PR #4456, opened 2026-06-16
feat: replace silent S3 fallbacks with explicit provider boundaries

### LINKED ISSUE #4411: Replace silent loader fallbacks with explicit provider boundaries and structured failure reporting
## What

Introduce explicit data-provider abstractions for local-filesystem and AWS-backed (S3) account/owner discovery in `backend/common/data_loader.py`, with typed result objects or domain exceptions distinguishing "missing data", "provider unavailable", and "invalid payload" — instead of converting all of these into empty results or silent local fallbacks.

## Why

`data_loader.py` currently mixes local filesystem discovery, S3 discovery, demo-owner logic, metadata parsing, virtual portfolios, and fallback behaviour in one module. Several provider failures are flattened into empty results without distinguishing transient outage from genuine "no data". Production incidents can therefore present as missing owners/accounts instead of actionable errors, which is hard to debug and risks serving incomplete data without an operator-visible signal.

## How

- Separate local and AWS account discovery/loading into distinct provider modules.
- Provider methods return typed models or raise well-defined exceptions instead of broad empty fallbacks.
- API routes distinguish missing data from provider outages and respond/log differently.
- Add tests covering S3 error, malformed JSON, empty payload, and fallback-to-local behaviour with explicit assertions.
- Preserve graceful degradation only where the API contract explicitly allows it, with structured logging/metrics on each fallback path.

## Constraints

- Existing callers of `data_loader` functions should keep working during the transition — consider a thin compatibility layer if needed rather than a single big-bang rename.
- Don't change the actual S3 bucket layout or local file format as part of this issue.

## LLM tier

**Sonnet** — requires introducing new abstractions and exception types across a module with several existing call sites; needs design judgment on the typed-model boundary.

## Success looks like

- Local and AWS discovery/loading live in separate provider modules.
- Distinct exception types exist for missing-data vs provider-unavailable vs invalid-payload.
- Tests explicitly cover all four failure-mode combinations described above.

## Failure looks like

- Fallback behaviour silently changes for existing working cases (e.g. demo owners stop resolving).
- Exceptions leak as unhandled 500s where the API previously returned a graceful empty response.

_From CODEBASE_REVIEW.md, item 2 ("v1 blockers")._

## Dependencies

| Relationship | Issue |
|---|---|
| Blocks | #4412 Introduce typed domain models |

leonarduk:
## Summary

- Removes the duplicate inline boto3 blocks from `load_account` and `load_person_meta` in `data_loader.py`, routing all S3 operations through `S3DataProvider` with explicit `MissingData` / `InvalidPayload` / `ProviderUnavailable` handling
- Fixes `data_providers._extract_person_meta` to not synthesise a `viewers: []` key when absent from the source document (the synthesised key was a latent inconsistency vs `data_loader._extract_person_meta`)
- Adds tests covering the four failure-mode combinations called out in the issue: S3 transport error, malformed JSON, empty payload, and fallback-to-local with structured warning log

## Test plan

- [ ] `pytest tests/test_data_loader_aws.py tests/test_data_loader_local.py` — 22 tests pass
- [ ] `ruff check backend/common/data_loader.py backend/common/data_providers.py` — no new errors

Closes #4411

## Release Notes

* **Bug Fixes**
  * Improved AWS/S3 account and person-metadata loading with consistent handling of missing data, malformed payloads, and provider unavailability, including clearer fallback behaviour and warnings.
  * Hardened metadata parsing by validating the `viewers` field and safely dropping invalid values.

* **Chores**
  * Streamlined the deployment workflow: simplified CI gating, updated action versions, refreshed runtime configuration, and improved log and authenticated Lambda invocation verification.
  * Made price snapshot verification non-blocking.

* **Tests**
  * Added expanded AWS/S3 edge-case coverage for malformed/empty payloads, fallback scenarios, and `viewers` validation.


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #4411` in PR body |
| Issue coverage | ✅ | All four failure-mode combinations from the issue are tested |
| Review comments | ✅ | Both CodeRabbit comments marked as addressed in prior commits |
| CI | ❌ | 3 failing checks (2 fixed, 1 transient) |
| Logic correctness | ✅ | After fixes below |

### Failing CI checks analysed

**Backend Integration Tests / CI** — `KeyError: 'full_name'` in `TestLoadPersonMeta`  
Root cause: the new `load_person_meta` enters the S3DataProvider path whenever `DATA_BUCKET` env var is set (which it is in CI). When the owner doesn't exist in S3, `MissingData` was raised and the code returned `{}` unconditionally — never reaching the local `data_root` fallback even when `data_root` was explicitly passed.  
Fix: `except MissingData` now only returns `{}` when `explicit_local_fallback is False`; otherwise falls through to the local index lookup.

**CDK Dry-Run** — `deploy-lambda.yml does not contain an active 'cdk diff --method=template' invocation`  
Root cause: `--method=change-set` was accidentally introduced in the workflow change, triggering the regression guard from issue #3287.  
Fix: restored `--method=template`.

**DeepSeek PR Review** — `ERROR: DeepSeek review output was empty`  
Transient API error — re-run is appropriate.

### Actions taken
- Pushed commit `ed6cec94` to this branch with both fixes:
  - `backend/common/data_loader.py`: `except MissingData` falls through to local when `explicit_local_fallback` is True
  - `.github/workflows/deploy-lambda.yml`: `--method=change-set` → `--method=template`

### Remaining blockers
- Re-run the DeepSeek PR Review workflow (transient API failure, no code change needed)

### Verdict
NOT READY — CI checks are re-running after fix commit. Once they pass and DeepSeek is re-run, this should be mergeable.

leonarduk:
## CI fixes and AC completion summary

Several rounds of fixes have been applied to this PR to address CI failures and two DeepSeek review passes. Summary of what was changed and why:

### Deploy workflow regressions (commit `25c38a11`)
The original branch accidentally bundled large unrelated regressions to `.github/workflows/deploy-lambda.yml` (dropped `GITHUB_DEPLOY_ROLE_ARN`, `FRONTEND_ORIGIN`, `CORS_ORIGINS`, `SMOKE_TEST_USER_POOL_CLIENT_ID`, the Reconcile OAuth step, the `verify-smoke-tests` job, and referenced a non-existent script). Reverted the workflow file entirely to main, since this PR's purpose is data-provider abstraction with no legitimate reason to touch the deployment pipeline.

### Integration test failure: `KeyError: 'full_name'` (commit `ed6cec94`)
`load_person_meta` entered the S3 path when `DATA_BUCKET` was set in CI, and `except MissingData` unconditionally returned `{}`, bypassing the explicit local fallback. Fixed to only return `{}` when `not explicit_local_fallback`.

### Price-snapshot verify step (commit `79a427fd`)
The "Verify price snapshot seeded in S3" workflow step had `continue-on-error: true` and used `::warning::` for non-404 errors instead of `exit 1`. Fixed per the test assertion.

### Warning log regression on AWS + no bucket (commit `b94b8654`)
When `config.app_env == "aws"` and `DATA_BUCKET` is unset and no explicit `data_root` is provided, the refactored code entered `S3DataProvider()` which raises `ProviderUnavailable`, triggering a spurious WARNING on every call. Restored the silent early-return guard before `S3DataProvider` construction to match pre-PR behaviour. Updated `test_load_person_meta_missing_bucket` to assert no WARNING is emitted.

### AC gap: `load_person_meta` local path not delegated to `LocalDataProvider` (commit `06541fa1`)
The local fallback in `load_person_meta` was the last inline file-reading path in `data_loader.py` not using a provider class. Replaced it with `LocalDataProvider().load_person_meta()`, so both `load_account` and `load_person_meta` are fully provider-abstracted for local and S3 sources. This satisfies the issue AC: *"Local and AWS discovery/loading live in separate provider modules."*

### DeepSeek findings assessed as false positives
- **`test_load_account_empty_payload` exception mismatch**: `InvalidPayload` is a `ValueError` subclass; test passes unchanged.
- **`viewers` contract breakage**: The old `data_providers._extract_person_meta` defaulted `viewers=[]`, but `LocalDataProvider.load_person_meta` was never called by production code in main — the S3 and local paths both went through `data_loader._extract_person_meta`, which already did not synthesise missing keys. All callers use `meta.get("viewers", [])` with a default. No production behaviour changed.
- **"Missing" tests for load_account fallback and no-bucket silent return**: Both tests already existed in the PR branch (`test_load_account_falls_back_to_local`, `test_load_person_meta_missing_bucket`).
- **API route distinction AC**: Already satisfied — `portfolio.py` maps `ProviderUnavailable` → 503, `InvalidPayload` → 502, `MissingData`/`FileNotFoundError` → 404.

PR #4459, opened 2026-06-16
fix: hard-fail verify-price-snapshot step on non-404 S3 errors

leonarduk:
## Summary

- Replaces the silent `::warning::` swallow with `exit 1` in the `else` branch of the "Verify price snapshot seeded in S3" step so genuine S3 errors (403, 500, network failures) hard-fail the deploy
- Removes `continue-on-error: true` from the step so the exit code reaches the job
- 404/NoSuchKey remains a soft `::warning::` since an unseeded bucket is not a deploy-blocker

## Test plan

- [ ] `pytest tests/scripts/test_deploy_lambda_workflow.py` — both tests pass

## Chores
- Optimised deployment workflow with faster CI verification—checks now complete immediately without polling delays
- Simplified IAM permission validation process for more efficient pre-flight checks  
- Enhanced post-deployment validation to automatically verify Lambda processing health
- Updated deployment workflow action pins


leonarduk:
Closing: this PR was based on a local main branch () that had diverged from origin/main and kept an older 808-line workflow. origin/main () already contains the correct fix — `exit 1` in the else branch and no `continue-on-error` on the verify step. Merging this PR would have accidentally deleted 327 lines of workflow. The test failure was caused by the stale local checkout, not a real regression.

PR #4582, opened 2026-06-21
Fix CodeQL py/reflective-xss in timeseries_meta.py (#276)

leonarduk:
## Summary

Fixes CodeQL code scanning alert [#276](https://github.com/leonarduk/allotmint/security/code-scanning/276) — reflected server-side cross-site scripting.

Closes #4581

## Root Cause

The `/timeseries/meta` endpoint was annotated with `response_class=HTMLResponse` even though it returns multiple content types (JSON, CSV, HTML). CodeQL traced user input from the `ticker` Query parameter to the response body and flagged it as a potential XSS vector. A misguided `html.escape()` call was added to the JSON response path in an attempt to silence CodeQL, but:

1. HTML escaping is incorrect for JSON content — JSON serialization handles escaping natively
2. It didn't silence the CodeQL alert anyway

## Fix

1. **Removed `response_class=HTMLResponse`** from `/meta` route. Each return path already specifies the correct response class explicitly (`JSONResponse`, `PlainTextResponse`, `HTMLResponse`).
2. **Removed `html.escape()` from JSON response path** — JSON serialization via `JSONResponse` handles all necessary escaping.
3. **Removed unused `import html`** — it was only used for the now-removed JSON escaping.

## Safety

The HTML output path remains safe:
- `_render_meta_html` → `render_timeseries_html` escapes title/subtitle via `html.escape()`
- Cell values are escaped via `df.to_html(escape=True)`
- Ticker/exchange values are validated upstream by `_TICKER_SEGMENT_RE` (alphanumeric, underscore, hyphen only)
- XSS payloads in `ticker` are rejected with 400 by `_resolve_ticker_exchange`

## Tests

All 69 related tests pass, including:
- `test_timeseries_meta_json_output_not_html_escaped` — asserts `data["ticker"] == "ABC.L"` (raw values in JSON)
- `test_timeseries_meta_json_rejects_xss_ticker` — XSS payloads rejected with 400
- All format tests (json, csv, html)
- All date range tests
- All XSS defense tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ticker data serialisation in JSON responses for the timeseries metadata endpoint to ensure correct formatting across multiple response formats.

PR #4609, opened 2026-06-22
fix(security): close CodeQL reflective-XSS alert #276 in /timeseries/meta

### LINKED ISSUE #4561: Fix reflected XSS in /timeseries/meta route still flagged by CodeQL (alert #276)

leonarduk:
## Summary

- Closes CodeQL alert [#276](https://github.com/leonarduk/allotmint/security/code-scanning/276) (`py/reflective-xss`, CWE-79, HIGH severity) on `backend/routes/timeseries_meta.py`
- Adds `_RESOLVER_OUTPUT_RE` to validate tickers/exchange codes returned by the internal resolver (the third path in `_resolve_ticker_exchange`), which was the remaining untainted flow CodeQL was tracking
- The resolver path previously trusted internal data unconditionally; it now rejects values containing HTML-unsafe characters (`<`, `>`, `&`, `"`, `'`) while still accepting valid exchange codes with dots (e.g. `XLON.G`)
- Keeps `html.escape()` in the JSON response as defence-in-depth (a no-op for values that passed `_RESOLVER_OUTPUT_RE`)
- Tightens the XSS rejection test assertions: only `400` is accepted (not `404`), and the `content-type` is verified as JSON

## Root cause

Alert #276 was filed after PR #4008 added `html.escape()` to the JSON path. CodeQL still flagged the route because the **resolver path** (case 3 in `_resolve_ticker_exchange`) could return user-influenced values without validation before they reached the response.

## Test plan

- [x] `python -m pytest tests/test_timeseries_meta_route.py -v` — 8 passed
- [x] `test_timeseries_meta_xss_payload_rejected[params0/1/2]` cover all three injection vectors (ticker+exchange, dot-separated ticker, exchange alone)
- [x] Valid tickers (e.g. `BRK-B.L`, `ABC.XLON.G`) are accepted; HTML-unsafe payloads return HTTP 400 with `application/json` content-type

Closes #4561

## Summary by CodeRabbit

* **Tests**
  * Improved test robustness to handle route objects without standard path attributes, enhancing test reliability.


leonarduk:
Closing as superseded. The CodeQL reflective-XSS fix for `backend/routes/timeseries_meta.py` that this PR was intended to carry was already merged to `main` via #4569 and #4582. After pulling `main` into this branch the only remaining diff was a broken route-registration check in `tests/test_app.py` (which used `route.path` iteration and failed against Starlette's `_IncludedRouter`). That test addition has been removed, leaving this PR with no net changes versus `main`.

Issue #4675, opened 2026-07-04
CSP has no style-src directive, blocking all React inline styles (541 occurrences, 56 files)


leonarduk:
## Summary

The deployed CSP in \`cdk/stacks/static_site_stack.py\` has no \`style-src\` directive:

\`\`\`python
_csp = "; ".join(
    [
        "default-src 'self'",
        "script-src 'self' https://accounts.google.com/gsi/client",
        "frame-src 'self' https://accounts.google.com/gsi/",
        f"connect-src 'self' {backend_url_param.value_as_string} https://*.amazoncognito.com",
        "img-src 'self' ...",
        "frame-ancestors 'none'",
        "object-src 'none'",
        "base-uri 'self'",
    ]
) + ";"
\`\`\`

Without \`style-src\`, browsers fall back to \`default-src 'self'\`, which blocks all inline \`style\` attributes unless \`'unsafe-inline'\`, a hash, or a nonce is present. React's \`style={{...}}\` prop renders as an inline \`style\` attribute, so every component using it is silently broken under this CSP in the deployed environment (confirmed via a live console warning: "Applying inline style violates the following Content Security Policy directive 'default-src 'self''... The action has been blocked.").

A grep across \`frontend/src\` found **541 occurrences of \`style={{\`** across **56 files**, including \`LoginPage.tsx\`, \`BackendUnavailableCard.tsx\`, \`App.tsx\`, \`InstrumentDetail.tsx\`, \`Screener.tsx\`, and more — so this is a broad, pre-existing gap, not isolated to one component.

## Suggested approaches (pick one, needs a decision, not just a quick patch)

1. **Add \`style-src 'self' 'unsafe-inline'\`** to the CSP — simplest, but weakens the CSP's protection against style-based attacks (still much lower risk than \`script-src 'unsafe-inline'\`).
2. **Migrate inline styles to Tailwind utility classes** (already used elsewhere in the codebase, e.g. \`AddPositionForm.tsx\`) or CSS modules — properly fixes the CSP posture but is a large, multi-file refactor.
3. **Nonce/hash-based \`style-src\`** — most correct but hard to wire up for a static SPA served from S3/CloudFront without an edge function to inject a per-request nonce.

## Context

[...]

PR #4676, opened 2026-07-04
fix: re-show login on 401 instead of looping on a stale auth token

### LINKED ISSUE #4577: Bug: /owners and /groups return HTTP 401 — awsUiAuth config missing from /config, Cognito login flow never triggered
## Summary

Every page load of the deployed frontend at https://d32n5ua14kb82t.cloudfront.net results in two immediate HTTP 401 errors:

- `GET https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com/owners` → 401 Unauthorized
- - `GET https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com/groups` → 401 Unauthorized
The app shows a **"Backend unavailable"** modal and the `Saved accounts` section displays _"No saved accounts yet for this owner"_ because no authenticated identity is ever established.

---

## Root Cause

The 401s are **not** caused by a backend misconfiguration. They are caused by the frontend making unauthenticated API calls because the **Cognito login flow is never triggered**.

### The auth architecture (correct design)

1. API Gateway has a **Cognito JWT authorizer** on all protected routes (`/owners`, `/groups`, `/{proxy+}`)
2. 2. The frontend is supposed to:
3.    - Read `awsUiAuth` fields from `/config`
4.    - Call `ensureAwsUiAuth()` → redirect to Cognito hosted UI → PKCE exchange → store Cognito ID token in `sessionStorage`
5.    - Call `applyCognitoIdToken()` → copy ID token to `localStorage['authToken']`
6.    - Attach `Authorization: Bearer <cognito-id-token>` on every protected API call
7. 3. API Gateway validates the token; if valid, invokes Lambda. `DISABLE_AUTH=true` in the Lambda env tells FastAPI to skip its own redundant JWT decode (correct).
### What is actually happening

The live `/config` endpoint returns **no `awsUiAuth` block**:

```json
{
  "app_env": "aws",
  "disable_auth": true,
  "google_auth_enabled": true,
  "google_client_id": "281640956843-ppuaennh8relsv8cdimn4tcd35qau1t0.apps.googleusercontent.com",
  ...
  // ⚠️ NO awsUiAuth.enabled, NO awsUiAuth.domain, NO awsUiAuth.clientId
}
```

Without `awsUiAuth.enabled = true` (and the `domain`/`clientId` fields), `ensureAwsUiAuth()` in `main.tsx` does nothing. `applyCognitoIdToken()` has no token to store. `localStorage['authToken']` is empty. Every API call goes out with no `Authorization` header → API Gateway Cognito authorizer returns 401 before Lambda is ever invoked.

The body of every 401 response is simply:
```json
{"message": "Unauthorized"}
```
This is the API Gateway default rejection — the Lambda is never reached.

---

## Evidence

| Check | Result |
|---|---|
| `GET /config` → `awsUiAuth` present? | ❌ Not present |
| `localStorage['authToken']` on page load | ❌ Empty |
| `sessionStorage` on page load | ❌ Empty |
| `GET /owners` with no Authorization header | 401 |
| `GET /owners` with `Authorization: Bearer demo` | 401 (gateway rejects) |
| `GET /config` itself | ✅ 200 (unprotected route) |
| `DISABLE_AUTH` in Lambda env (`cdk/stacks/backend_lambda_stack.py`) | `"true"` ✅ correct |

---

## Likely Gap in the Codebase

The CDK stack provisions the Cognito User Pool and App Client, but one or more of the following is missing:

1. **Lambda env vars not set**: `COGNITO_DOMAIN`, `COGNITO_APP_CLIENT_ID` (or equivalent) are not being passed as environment variables to the backend Lambda in `cdk/stacks/backend_lambda_stack.py`
2. 2. **Config model missing fields**: `backend/config.py` does not have an `aws_ui_auth` section that reads those env vars
3. 3. **SPA contract not serialising fields**: `backend/contracts_spa.py` does not include `awsUiAuth` in the `/config` response payload
The result: the frontend receives a config with no Cognito settings and silently skips the entire login flow.

---

## Steps to Reproduce

1. Open https://d32n5ua14kb82t.cloudfront.net/input in a browser (no prior login)
2. 2. Open DevTools → Network
3. 3. Observe `GET /owners` and `GET /groups` returning 401
4. 4. Open DevTools → Application → Local Storage → confirm `authToken` is absent
5. 5. `fetch('https://dxt34gh2i0.execute-api.eu-west-1.amazonaws.com/config').then(r=>r.json()).then(console.log)` → confirm no `awsUiAuth` key in response
---

## Expected Behaviour

On page load the app should:
1. Detect `awsUiAuth.enabled = true` in `/config`
2. 2. Redirect to the Cognito hosted UI login page
3. 3. After successful login, store the Cognito ID token and use it on all subsequent API calls
4. 4. Protected endpoints (`/owners`, `/groups`) return 200 with data
---

## Acceptance Criteria

- [ ] `/config` response includes `awsUiAuth` object with at minimum `enabled`, `domain`, `clientId`
- [ ] - [ ] On fresh page load (no stored session), the app redirects to Cognito hosted UI
- [ ] - [ ] After Cognito login, `localStorage['authToken']` contains the Cognito ID token
- [ ] - [ ] `GET /owners` and `GET /groups` return 200 after login
- [ ] - [ ] No 401 errors appear in the network tab for an authenticated user
- [ ] - [ ] CDK stack passes the Cognito domain and client ID as Lambda environment variables
- [ ] - [ ] `backend/config.py` model includes the `aws_ui_auth` fields
- [ ] - [ ] `backend/contracts_spa.py` serialises those fields into the `/config` SPA response
---

## Related

- PR #4558 — "fix: enable Cognito authentication in backend Lambda" (reverted `DISABLE_AUTH` change; `DISABLE_AUTH=true` is correct — this is a separate, upstream issue)
- - Issue #4256 — previous `/owners` 401 investigation (resolved gateway authorizer config; the current issue is the missing frontend config handoff)
- - - `docs/AUTH.md` — full auth architecture documentation
- - - - `frontend/src/awsUiAuth.ts` — PKCE flow implementation
- - - - - `frontend/src/main.tsx` — `applyCognitoIdToken()` / `ensureAwsUiAuth()` bootstrap
- - - - - - `cdk/stacks/backend_lambda_stack.py` — Lambda env vars and Cognito authorizer setup

### LINKED ISSUE #4674: fix: stale localStorage authToken bypasses Cognito re-login, causing permanent 401 loop
## Summary

On the deployed AWS app (https://d32n5ua14kb82t.cloudfront.net), reopening the app after the Cognito session has expired (or in a new tab, after the old tab was closed) shows the full app UI instead of the login screen, but every protected API call (`/owners`, `/groups`, `/portfolio-group/all`, `/portfolio-group/all/instruments`) fails with 401. The app then falls back to the generic "Backend unavailable" card, and clicking Retry just repeats the same 401s forever.

This looks superficially like the previously-fixed #4577 (missing `awsUiAuth` in `/config`), but it is not — I confirmed the live `/config` endpoint currently returns a correct `awsUiAuth` block (`enabled: true`, `domain`, `clientId`). This is a different, new bug.

## Root cause

1. `frontend/src/main.tsx` seeds the `authed` state on boot as:
   ```ts
   const [authed, setAuthed] = useState(
     Boolean(storedToken) || Boolean(getStoredCognitoIdToken()),
   );
   ```
   `storedToken` comes from `localStorage['authToken']` (via `getStoredAuthToken()` in `api.ts`) and has **no expiry check**.
2. The Cognito ID token is cached in `sessionStorage` (cleared on tab close, see `awsUiAuth.ts`), but `applyCognitoIdToken()` also copies it into `localStorage['authToken']` via `setAuthToken()`. `localStorage` survives tab closes and outlives the token's ~1 hour lifetime.
3. On a later visit, `sessionStorage` is empty (so Cognito considers the session gone), but `localStorage['authToken']` still holds the old, now-expired ID token. `authed` is seeded `true`, so `Root` in `main.tsx` skips the `LoginPage` gate entirely and renders the app.
4. The app then calls `/owners`, `/groups`, etc. with `Authorization: Bearer <expired-id-token>`. API Gateway's Cognito JWT authorizer rejects it with 401 before the Lambda ever runs.
5. `App.tsx` (`ownersReq.error || groupsReq.error` → `setBackendUnavailable(true)`) treats this the same as a generic backend-down error and shows `BackendUnavailableCard`. There is no code path that recognizes "this is a 401 caused by a stale/invalid session" and clears it to force re-login — so Retry just resends the same dead token and 401s again.

## Expected behavior

When any authenticated API call returns 401, the frontend should clear the stale Cognito session/token and return the user to the sign-in screen, instead of showing a generic "backend unavailable" loop.

## Steps to reproduce

1. Sign in via Cognito hosted UI on the deployed app.
2. Close the tab (or wait past the ID token's ~1 hour expiry) so `sessionStorage` clears/expires but `localStorage['authToken']` remains.
3. Reopen the app.
4. Observe the app renders normally (no login prompt) but `/owners`, `/groups`, etc. all return 401, and the app shows "Backend unavailable".

## Related (but distinct) finding

Separately, the browser console also shows a CSP violation: `Applying inline style violates ... 'default-src 'self'' ... style-src was not explicitly set`. The deployed CSP (`cdk/stacks/static_site_stack.py`) has no `style-src` directive, so it falls back to `default-src 'self'`, which blocks all React inline `style={{...}}` props. This is a much larger, pre-existing, systemic issue (541 occurrences across 56 frontend files) — tracking separately rather than folding into this fix.

leonarduk:
## Summary

- On the deployed app, reopening the app after the Cognito session has expired (tab closed, or ID token past its ~1h lifetime) rendered the full app UI instead of the login screen, because `localStorage['authToken']` outlives the sessionStorage-scoped Cognito session and has no expiry check in `Root`'s initial `authed` state.
- Every API call then went out with the stale token, API Gateway's Cognito authorizer 401'd all of them, and `App.tsx` showed a generic "Backend unavailable" card — clicking Retry just resent the same dead token forever, with no way to recover except manually clearing storage.
- `fetchJson` in `api.ts` now dispatches a new `UNAUTHORIZED_EVENT` on any 401 response. `Root` in `main.tsx` listens for it and clears the stale Cognito session + auth token, returning the user to the login screen.
- Also converted `BackendUnavailableCard.tsx`'s inline `style={{...}}` props to Tailwind classes, since the deployed CSP has no `style-src` directive (falls back to `default-src 'self'`), which was silently blocking this card's styling — visible in the reported console output as a CSP violation.

Confirmed this is a distinct bug from the previously-fixed #4577 (missing `awsUiAuth` in `/config`) — I checked the live `/config` endpoint directly and it currently returns a correct `awsUiAuth` block, so that fix is deployed and working.

Filed #4675 separately for the broader CSP inline-style problem (541 occurrences across 56 frontend files) — out of scope for this fix, needs its own design decision (loosen CSP vs. Tailwind migration vs. nonce-based CSP).

Closes #4674

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — clean
- [x] `npx vitest run` — 541 tests pass, including 3 new tests covering: `fetchJson` dispatches `UNAUTHORIZED_EVENT` only on 401 (not other error statuses), and a `Root` integration test simulating a stale stored token being rejected mid-session, verifying it clears the token and falls back to the login page.

## Summary by CodeRabbit

* **New Features**
  * The app now detects unauthorised API responses and resets the session automatically, returning users to the login screen.

* **Bug Fixes**
  * Stale sign-in tokens are cleared when access is rejected, helping prevent repeated failed retries.
  * The backend unavailable screen has been visually refreshed for a more consistent layout and spacing.

### DIFF
--- a/frontend/src/api.ts
+++ b/frontend/src/api.ts
@@ -129,6 +129,17 @@ const defaultGetCsrfToken = () =>
         .find((row) => row.startsWith("csrftoken="))
         ?.split("=")[1] || null;
 
+/**
+ * Dispatched on `window` whenever an API call receives HTTP 401. A stored
+ * auth token (Cognito ID token or backend JWT) can go stale independently of
+ * the app's in-memory auth state — e.g. localStorage persists it past the
+ * ~1h Cognito ID token lifetime, or past a Cognito session that sessionStorage
+ * already dropped on tab close. Listening for this event lets the app clear
+ * the stale credential and re-show the login screen instead of retrying the
+ * same rejected token forever.
+ */
+export const UNAUTHORIZED_EVENT = "allotmint:unauthorized";
+
 export function createClient(
   base: string | (() => string),
   token: string | null = null,
@@ -226,6 +237,9 @@ export function createClient(
       credentials: "include",
     });
     if (!res.ok) {
+      if (res.status === 401 && typeof window !== "undefined") {
+        window.dispatchEvent(new Event(UNAUTHORIZED_EVENT));
+      }
       const err = new Error(`HTTP ${res.status} - ${res.statusText} (${safeUrl})`);
       (err as any).status = res.status;
       (err as any).headers = res.headers;
--- a/frontend/src/components/BackendUnavailableCard.tsx
+++ b/frontend/src/components/BackendUnavailableCard.tsx
@@ -4,32 +4,23 @@ interface Props {
 
 export default function BackendUnavailableCard({ onRetry }: Props) {
   return (
-    <div
-      style={{
-        maxWidth: 400,
-        margin: "2rem auto",
-        padding: "1rem",
-        border: "1px solid #ccc",
-        borderRadius: "8px",
-        textAlign: "center",
-      }}
-    >
+    <div className="mx-auto my-8 max-w-[400px] rounded-lg border border-gray-300 p-4 text-center">
       <h2>Backend unavailable</h2>
-      <p style={{ color: "#555", marginBottom: "1rem" }}>
+      <p className="mb-4 text-gray-600">
         The backend service could not be reached. You can retry or open a
         cached read-only view.
       </p>
-      <div style={{ marginBottom: "1rem" }}>
+      <div className="mb-4">
         <button onClick={onRetry} disabled={!onRetry}>
           Retry
         </button>
       </div>
-      <div style={{ fontSize: "0.9rem" }}>
+      <div className="text-sm">
         <a
           href="/snapshots/index.html"
           target="_blank"
           rel="noopener noreferrer"
-          style={{ marginRight: "0.5rem" }}
+          className="mr-2"
         >
           Cached view
         </a>
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -32,6 +32,7 @@ import {
   getStoredAuthToken,
   setApiBase,
   setAuthToken,
+  UNAUTHORIZED_EVENT,
 } from './api';
 import LoginPage from './LoginPage';
 import { UserProvider, useUser } from './UserContext';
@@ -185,6 +186,26 @@ export function Root({ awsUiAuth = runtimeAwsUiAuth }: { awsUiAuth?: AwsUiAuthCo
     if (existingUser) setUser(existingUser);
     const existingProfile = loadStoredUserProfile();
     if (existingProfile) setProfile(existingProfile);
+  }, [setProfile, setUser]);
+
+  // A stored auth token can go stale independently of this component's state
+  // (e.g. localStorage outlives the ~1h Cognito ID token, or outlives a
+  // Cognito session sessionStorage already dropped on tab close). Without
+  // this, a stale token keeps `authed` true forever, so every protected API
+  // call 401s and the app just shows a "backend unavailable" retry loop
+  // instead of returning to the login screen. See issue #4674.
+  useEffect(() => {
+    const handleUnauthorized = () => {
+      clearCognitoRefreshTimer();
+      clearCognitoSession();
+      apiLogout();
+      setUser(null);
+      setProfile(undefined);
+      setAuthed(false);
+    };
+    window.addEventListener(UNAUTHORIZED_EVENT, handleUnauthorized);
+    return () =>
+      window.removeEventListener(UNAUTHORIZED_EVENT, handleUnauthorized);
   }, [setProfile, setUser]);
 
   // Derive a stable boolean from the /config.json-sourced prop so the
--- a/frontend/tests/unit/api.test.ts
+++ b/frontend/tests/unit/api.test.ts
@@ -13,6 +13,7 @@ import {
   getPortfolio,
   getPensionForecast,
   getConfig,
+  UNAUTHORIZED_EVENT,
 } from "@/api";
 
 describe("auth token handling", () => {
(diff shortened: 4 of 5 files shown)


leonarduk:
The DeepSeek review flags `apiLogout`, `clearCognitoSession`, and `clearCognitoRefreshTimer` as unimported in `main.tsx`. This isn't accurate:

- `apiLogout` is imported at line 31 (`logout as apiLogout` from `./api`)
- `clearCognitoSession` is imported at line 42 from `./awsUiAuth`
- `clearCognitoRefreshTimer` is a module-level `const` defined at line 508, referenced inside a closure that only executes once the `401` event fires (after the whole module has evaluated) — the same pattern the existing `logout()` function already uses at line 171.

`npx tsc --noEmit` is clean and all 541 vitest tests pass, which a genuine missing-import would fail. Requesting this check be re-run or dismissed.

PR #4711, opened 2026-07-06
feat(news): surface published dates and flag stale market headlines

### LINKED ISSUE #4708: Market headlines are months stale and don't show a published date
## What
On `/market`, the "Latest Headlines" list shows headlines from March/April even though news is refreshed via a 15-minute cache (`NEWS_TTL` in `backend/routes/news.py`). Separately, the headline list gives no indication of how old each story is — there's no published date shown next to any link.

## Why
Two related gaps:
- `MarketOverview.tsx` (`frontend/src/pages/MarketOverview.tsx:161-180`) renders each `NewsItem` as just `<a href={h.url}>{h.headline}</a>` with no date, so users can't tell a headline is stale just by looking at the page.
- The date is actually unavailable to show: `NewsItem.published_at` exists in `frontend/src/types.ts:283-288`, but the backend never populates it. `_make_news_item()` (`backend/routes/news.py:51-58`) and `_trim_payload()` (`backend/routes/news.py:111-121`) only keep `headline`/`url` and discard everything else. There's even a `_parse_alpha_time()` helper (`backend/routes/news.py:225-245`) that converts AlphaVantage's `time_published` field to ISO-8601 — but it is dead code, never called anywhere.
- The month-old stale headlines are likely a symptom of `get_cached_news()` (`backend/routes/news.py:288-352`): if live fetches from AlphaVantage/Yahoo/Google are failing in production (e.g. AlphaVantage quota exhausted, or Yahoo/Google endpoints blocked from the deployed environment), the function silently keeps re-serving the last successfully cached payload indefinitely, with nothing surfaced to the user (or logs) indicating the data is months old.

## How
- Wire `time_published` (AlphaVantage) and equivalent fields from the Yahoo/Google fallbacks through `_make_news_item`/`_trim_payload` into a `published_at` field on each news item, using the existing `_parse_alpha_time`/`_isoformat` helpers instead of leaving them unused.
- Update `MarketOverview.tsx` to render the published date next to each headline (e.g. "Headline — 3 days ago" or a formatted date), handling the case where `published_at` is missing.
- Investigate why headlines are going months without refreshing: check whether `_fetch_news` / `get_cached_news` is repeatedly hitting the `RuntimeError` quota-exhausted path or failing outbound requests in production, and log/alert when cached news exceeds some sane staleness bound (e.g. much older than `NEWS_TTL`) instead of serving it silently forever.

## Constraints
- AlphaVantage free-tier quota (`cfg.news_requests_per_day`) is limited; any fix must not increase request volume in a way that burns quota faster.
- Don't break the existing fallback chain (AlphaVantage → Yahoo → Google) or the background refresh scheduling in `page_cache`.
- Add/update pytest coverage under `tests/` for the new `published_at` field, and Vitest coverage for the date rendering.

## LLM tier
**Sonnet** — the `published_at` wiring is a mechanical plumbing fix, and the staleness investigation is grunt-work log/config reading rather than deep architectural judgment.

## Success looks like
- [ ] Each headline in `NewsItem`/`/market/overview` includes a populated `published_at` when the upstream source provides one.
- [ ] `/market` displays the publish date/age next to each headline.
- [ ] Root cause of the months-long staleness is identified and fixed (or, if it's inherent to quota limits, the UI/logs make the staleness visible instead of silent).

## Failure looks like
- Headlines on `/market` still show no date and remain stuck on the same March/April stories with no visible cause.
- `published_at` stays `null`/absent for all sources even after the fix.

leonarduk:
## What
Fixes the two related gaps from #4708: market headlines had no publish date, and `published_at` was never populated by the backend (the `_parse_alpha_time` helper was dead code). Separately, months-stale cached news was being re-served silently.

### Backend (`backend/routes/news.py`)
- `_make_news_item` now accepts optional `published_at` (already-normalised ISO-8601) and `source`, including them only when present so the minimal `{headline, url}` shape is unchanged for sources that lack them.
- `_trim_payload` preserves `published_at`/`source` from cached payloads.
- Publish times wired through for all three sources:
  - **AlphaVantage** `time_published` → existing `_parse_alpha_time` (no longer dead code).
  - **Yahoo** `providerPublishTime` (Unix epoch) → new `_parse_epoch_time`.
  - **Google RSS** `pubDate` (RFC 822) → new `_parse_rss_time`.
- `source` populated from each provider's publisher/source field.

### Staleness visibility
- New `page_cache.cache_age()` helper.
- `get_cached_news` logs a warning (`NEWS_MAX_STALENESS`, 1 day) when it re-serves a cached payload that has aged out because live fetches keep failing / quota stays exhausted — surfacing the root cause of months-old headlines instead of ageing invisibly. No extra request volume, so AlphaVantage quota is unaffected; the fallback chain and background refresh are untouched.

### Frontend
- `formatPublishedAt` helper in `lib/date.ts` renders a relative age via `Intl.RelativeTimeFormat`, returning `null` for missing/unparseable values.
- `MarketOverview.tsx` shows "Headline — 3 days ago" next to each link (with the exact timestamp as a tooltip), omitting the age when `published_at` is absent.

## Tests
- pytest: `published_at`/`source` population for all three fetchers, the new time parsers, `_trim_payload` preservation, and the staleness-warning path; updated one existing route test that asserted the old minimal shape.
- Vitest: `formatPublishedAt` unit tests + `MarketOverview` rendering (dated vs. undated headlines).

All targeted backend (72) and frontend (8) tests pass; `tsc --noEmit` and `eslint` clean on changed files. (Backend `ruff`/`black` not installed in this environment — no formatting run locally.)

Closes #4708

### DIFF
--- a/backend/routes/news.py
+++ b/backend/routes/news.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import logging
 from datetime import date, datetime, timezone
+from email.utils import parsedate_to_datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
@@ -24,6 +25,10 @@ config = cfg
 router = APIRouter(tags=["news"])
 
 NEWS_TTL = 900  # seconds
+# Serving cached news older than this (in seconds) is treated as a fault worth
+# surfacing: it means live fetches have not succeeded for far longer than the
+# normal refresh interval (``NEWS_TTL``) and the user is seeing stale stories.
+NEWS_MAX_STALENESS = 24 * 60 * 60  # 1 day
 BASE_URL = "https://www.alphavantage.co/query"
 COUNTER_FILE: Path = page_cache.CACHE_DIR / "news_requests.json"
 
@@ -48,14 +53,32 @@ _FINANCE_KEYWORDS = (
 
 
 
-def _make_news_item(headline: object, url: object) -> Dict[str, str] | None:
-    """Return a minimal news item when both ``headline`` and ``url`` are valid."""
+def _make_news_item(
+    headline: object,
+    url: object,
+    published_at: object = None,
+    source: object = None,
+) -> Dict[str, str] | None:
+    """Return a news item when both ``headline`` and ``url`` are valid.
+
+    ``published_at`` (an already-normalised ISO-8601 string) and ``source`` are
+    optional and only included when present, so callers that lack them still
+    produce the minimal ``{"headline", "url"}`` shape.
+    """
 
     clean_headline = _clean_str(headline)
     clean_url = _clean_str(url)
-    if clean_headline and clean_url:
-        return {"headline": clean_headline, "url": clean_url}
-    return None
+    if not (clean_headline and clean_url):
+        return None
+
+    item: Dict[str, str] = {"headline": clean_headline, "url": clean_url}
+    clean_published = _clean_str(published_at)
+    if clean_published:
+        item["published_at"] = clean_published
+    clean_source = _clean_str(source)
+    if clean_source:
+        item["source"] = clean_source
+    return item
 
 
 def _load_counter() -> Dict[str, int]:
@@ -115,7 +138,12 @@ def _trim_payload(payload: Any) -> List[Dict[str, str]]:
     for item in payload:
         if not isinstance(item, dict):
             continue
-        news_item = _make_news_item(item.get("headline"), item.get("url"))
+        news_item = _make_news_item(
+            item.get("headline"),
+            item.get("url"),
+            item.get("published_at"),
+            item.get("source"),
+        )
         if news_item is not None:
             trimmed.append(news_item)
     return trimmed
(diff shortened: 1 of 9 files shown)


leonarduk:
Thanks for the DeepSeek review. Addressed:

**The `REQUEST CHANGES` item (`cache_age` → `float | None` requires 3.10+) is not applicable.** `backend/utils/page_cache.py` has `from __future__ import annotations` (line 11), so annotations are never evaluated at runtime — `float | None` is valid even on 3.9. The pre-existing `time_until_stale` on the very next line already uses the identical annotation, and `pyproject.toml` pins `requires-python = ">=3.11"` regardless. No change needed.

**`source` type drift — not present.** `NewsItem` in `frontend/src/types.ts:286` already declares `source?: string | null`, so backend and frontend types are aligned.

**Added the suggested missing test** (a31f9d6b): `test_fetch_news_alpha_omits_published_at_when_missing` confirms `_fetch_news` omits `published_at` (rather than emitting a null) when an AlphaVantage feed item has no `time_published`.

The remaining notes (numpy numeric types in `_parse_epoch_time`, `-0` from `Math.round` on future timestamps) are correctly flagged as non-blocking — JSON-sourced data can't produce numpy scalars, and the `-0` case is cosmetic — so I've left them as-is.

PR #4715, opened 2026-07-06
fix: default owner selection to the logged-in user, not owners[0]

### LINKED ISSUE #4706: Manual holdings input form defaults Owner field to first owner instead of the logged-in user
## What
The "Account + Holdings Input" form (`/input` route, rendered by `TransactionsPage`) does not use the logged-in user's identity for the Owner field. Instead it always defaults to the first owner in the fixture/data list (`owners[0].owner`), regardless of who is actually authenticated. In the current dataset this happens to be "alice", so any logged-in user sees "alice" pre-filled as the owner on page load.

## Why
The app already has an `AuthContext`/`useAuth()` hook that knows the logged-in user, but `TransactionsPage` never consults it when initializing the manual holdings form. This is confusing and risks a user unintentionally saving holdings under the wrong owner if they don't notice the field.

## How
- In `frontend/src/components/TransactionsPage.tsx`, update the effect that seeds `manualOwner` (currently `setManualOwner((current) => current || owners[0].owner)`) to prefer the authenticated user's owner (via `useAuth()`) when it matches an entry in `owners`.
- Fall back to the current behavior (`owners[0].owner`) only when there's no logged-in user or no matching owner in the list, so the demo/local-dev experience without auth still works.
- Verify the mapping between the authenticated user's identity (email/username) and the `owner` values used elsewhere in the app (e.g. wherever `OwnerSelector`/`useAuth` are already reconciled, such as `pages/UserConfig.tsx`), and reuse that mapping rather than inventing a new one.

## Constraints
- Must not break the no-auth/demo/local flow where `owners` is populated but there is no logged-in user.
- Must not change behavior of the `transactions` filter Owner field (`owner`/`setOwner`) — this fix is scoped to the manual holdings `manualOwner` field only.
- Add/update Vitest coverage for the new default-owner logic per `CLAUDE.md` frontend guidance.

## LLM tier
**Sonnet** — small, well-scoped frontend fix, but requires tracing how the existing auth context maps to owner identifiers before wiring it in correctly.

## Success looks like
- [ ] Logging in as a user whose identity maps to a known owner pre-fills the Owner field with that owner, not the first owner in the list.
- [ ] Users with no auth-mapped owner (or no session) still see the existing fallback behavior (first owner in the list).
- [ ] New/updated Vitest test covers both the matched-owner and fallback cases.

## Failure looks like
- The Owner field still always shows the first owner in `owners` regardless of who is logged in.
- The fallback path breaks for demo/local usage without authentication.

### LINKED ISSUE #4707: Settings page owner dropdown is not scoped to the logged-in user
## What
The `/settings` page (`UserConfigPage` in `frontend/src/pages/UserConfig.tsx`) shows an "owner" dropdown populated with every owner returned by `getOwners()` (e.g. "alice", "demo-owner") with no connection to who is actually logged in. The page already has the authenticated user via `useAuth()` (used just above to render name/email/avatar) but never uses it to scope or pre-select the owner. Any authenticated user can pick any owner in the list and view/edit that owner's hold-days, max-trades, exempt tickers/types, and trade approvals.

## Why
This is the same underlying gap as [#4706](https://github.com/leonarduk/allotmint/issues/4706): the frontend has no mapping between the logged-in identity (`useAuth()`'s `user.email`/`user.name`) and the `owner` records it operates on. On `/input` that showed up as the wrong default value; here it's more serious because it lets a signed-in user browse and mutate configuration/approvals belonging to a different owner from the settings UI, with no client-side scoping at all.

## How
- Establish (or reuse, if it already exists server-side) a mapping from the authenticated user to their `owner` id, and use it in `UserConfigPage` to pre-select and, if appropriate, restrict the dropdown to owners the current user is permitted to manage.
- If multi-owner access is an intentional feature (e.g. a family admin managing several owners), scope the dropdown to only the owners the logged-in user is authorized for, not every owner in the system.
- Confirm whether the backend `user-config`/`approvals` endpoints already enforce authorization per owner; if they don't, note that as a follow-up since a same-origin fetch to `/user-config/<owner>` for an unauthorized owner would currently succeed.

## Constraints
- Don't break legitimate multi-owner use cases (e.g. a household account where one login manages multiple owners) if that's by design — check with product/user intent before hard-restricting to a single owner.
- Frontend-only scoping is not a substitute for backend authorization; flag if the API has no owner-level access control.

## LLM tier
**Opus** — requires understanding the intended auth/owner permission model (single-owner vs. family-admin) before deciding whether to restrict or just default the dropdown, and may touch backend authorization.

## Success looks like
- [ ] The Settings page pre-selects the current user's own owner when a mapping exists, instead of requiring manual selection from an unscoped list.
- [ ] The dropdown only lists owners the logged-in user is actually authorized to manage (or the app confirms and documents that any authenticated user is intentionally allowed to manage any owner).
- [ ] A written note (issue comment or doc) on whether backend `user-config`/`approvals` endpoints enforce owner-level authorization.

## Failure looks like
- The dropdown continues to list every owner in the system with no relationship to the logged-in identity.
- A user can still load another owner's approvals/settings via the UI without any authorization check, client- or server-side.

### LINKED ISSUE #4709: App-wide: default owner selection ignores the logged-in user across all pages
## What
Across the app, the "current owner" shown by default has no relationship to who is actually logged in. The shared `selectedOwner` value in `frontend/src/contexts/route.ts` (exposed via `useRoute()`) is seeded to `owners[0].owner` — the first owner returned by `getOwners()` — in `frontend/src/MainApp.tsx:128-133`, and several pages repeat the same "fall back to `owners[0]`" pattern locally when `selectedOwner` isn't already set (e.g. `Rebalance.tsx:128-134`, `PensionForecast.tsx:90-95`). None of these consult `useAuth()`, which already knows the logged-in user's name/email.

This one root cause is why "alice" and/or "demo-owner" show up unprompted on:
- `/input` (manual holdings owner field — [#4706](https://github.com/leonarduk/allotmint/issues/4706))
- `/settings` (owner dropdown — [#4707](https://github.com/leonarduk/allotmint/issues/4707))
- `/pension/forecast`
- `/rebalance`
- likely also `/portfolio`, `/screener`, `/tax-tools`, `/performance-diagnostics`, `/scenario-tester`, and anywhere else that reads `selectedOwner` from `useRoute()`, since they all inherit the same app-wide fallback.

## Why
Filing a separate issue per page chases the same bug N times. The actual defect is a single default-selection decision made once in `MainApp.tsx` (and duplicated ad hoc in a couple of pages) that never looks at identity. Fixing it in one place fixes every consuming page at once.

## How
- In `MainApp.tsx`, when seeding `selectedOwner` (currently `owners[0].owner`), first check whether the authenticated user (`useAuth()`) maps to one of the returned `owners`; use that as the default, and only fall back to `owners[0]` when there's no match (e.g. demo/local/no-auth mode).
- Establish (or confirm, if it already exists on the backend) the mapping between the authenticated identity (`user.email`/`user.name`) and an `owner` id — this is the same gap noted in #4707.
- Once the shared default is fixed, audit `Rebalance.tsx` and `PensionForecast.tsx` for their own local `owners[0]` fallbacks (`Rebalance.tsx:128-134`, `PensionForecast.tsx:90-95`) and make them prefer `route.selectedOwner` from the corrected shared context rather than re-deriving their own first-owner fallback.
- Close out #4706 and #4707 as duplicates/instances of this issue once the shared fix lands, verifying each of their specific pages picks up the corrected default.

## Constraints
- Must not break demo/local/no-auth usage where there's no logged-in user to map — keep the `owners[0]` fallback for that case.
- Must not silently restrict which owners a user can select (that authorization question is tracked separately in #4707) — this issue is about the *default*, not access control.
- Add/update Vitest coverage for the corrected default-selection logic in `MainApp.tsx` and for any page-local fallbacks that get removed.

## LLM tier
**Sonnet** — the fix is a mechanical default-selection change in one place (`MainApp.tsx`) plus removing a couple of duplicated fallback blocks; broad footprint, but no architectural ambiguity.

## Success looks like
- [ ] `MainApp.tsx` seeds `selectedOwner` from the logged-in user's mapped owner when one exists, falling back to `owners[0]` only when there's no match.
- [ ] `/input`, `/settings`, `/pension/forecast`, `/rebalance`, and other `selectedOwner`-consuming pages all reflect the corrected default with no page-specific patching needed.
- [ ] `#4706` and `#4707` are closed as superseded by this fix, with each page re-verified.

## Failure looks like
- Individual pages keep getting patched one at a time while new pages keep exhibiting the same "shows alice/demo-owner regardless of login" symptom.
- The shared `MainApp.tsx` default is fixed but page-local fallbacks (`Rebalance.tsx`, `PensionForecast.tsx`) still override it with their own unscoped `owners[0]` logic.

leonarduk:
## Summary
- Adds `findOwnerForUser(owners, user)` in `frontend/src/utils/owners.ts`, which maps the authenticated user's email to a matching `OwnerSummary.email`.
- Uses it ahead of the `owners[0]` fallback everywhere a default owner is seeded from the shared owner list:
  - `App.tsx` — `getOwnerRootRedirectPath`, the route-sync effect, and the render-time owner-root redirect (all three previously duplicated `owners[0]` logic).
  - `MainApp.tsx` — the (currently unused, but still tested) legacy owner-view default.
  - `components/TransactionsPage.tsx` — the manual holdings "Owner" input default (the concrete bug from #4706).
  - `pages/UserConfig.tsx` — the settings owner dropdown default (the concrete bug from #4707).
- `Rebalance.tsx` and `PensionForecast.tsx` were audited per the issue's checklist — both already prefer `route.selectedOwner` over their own `owners[0]` fallback, so no page-specific fallback removal was needed there; they now inherit the corrected default via the shared context.
- Falls back to `owners[0]` unchanged whenever there's no auth user or no owner match (demo/local/no-auth mode), per the issue's constraint.

## Test plan
- [x] `npx vitest run` — full frontend suite passes (555/555)
- [x] `npm run lint` — clean
- [x] New Vitest coverage added for the corrected default-selection logic:
  - `utils/owners.test.ts` — `findOwnerForUser`
  - `ownerRootRedirect.test.ts` — `getOwnerRootRedirectPath` with a logged-in user
  - `App.test.tsx` — `/portfolio` redirects to the auth-matched owner, not the first owner
  - `UserConfig.test.tsx` — settings dropdown defaults to the auth-matched owner
  - `TransactionsPage.test.tsx` — manual holdings input defaults to the auth-matched owner

Closes #4706
Closes #4707
Closes #4709

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -54,7 +54,9 @@ import Rebalance from './pages/Rebalance';
 import PensionForecast from './pages/PensionForecast';
 import TaxTools from './pages/TaxTools';
 import Alerts from './pages/Alerts';
-import { sanitizeOwners } from './utils/owners';
+import { sanitizeOwners, findOwnerForUser } from './utils/owners';
+import { useAuth } from './AuthContext';
+import type { UserProfile } from './AuthContext';
 import { isDefaultGroupSlug, normaliseGroupSlug } from './utils/groups';
 import { deriveModeFromPathname } from './pageManifest';
 import { MAX_INSTRUMENT_CATALOGUE_ROWS } from './constants/renderLimits';
@@ -107,14 +109,15 @@ function sameGroupList(left: GroupSummary[], right: GroupSummary[]): boolean {
 export function getOwnerRootRedirectPath(
   pathname: string,
   selectedOwner: string,
-  owners: OwnerSummary[]
+  owners: OwnerSummary[],
+  user?: UserProfile | null
 ): string | null {
   if (selectedOwner || owners.length === 0) return null;
   const segs = pathname.split('/').filter(Boolean);
   const atPortfolioRoot = segs[0] === 'portfolio' && segs.length === 1;
   const atPerformanceRoot = segs[0] === 'performance' && segs.length === 1;
   if (!atPortfolioRoot && !atPerformanceRoot) return null;
-  const owner = owners[0].owner;
+  const owner = findOwnerForUser(owners, user)?.owner ?? owners[0].owner;
   const encodedOwner = encodePathSegment(owner);
   return atPerformanceRoot
     ? `/performance/${encodedOwner}`
@@ -156,6 +159,7 @@ export default function App({ onLogout }: AppProps) {
   const { t } = useTranslation();
   const { tabs, disabledTabs, familyMvpEnabled, configLoaded } = useConfig();
   const { lastRefresh } = usePriceRefresh();
+  const { user } = useAuth();
   const familyMvpEntryPath = useMemo(
     () => (configLoaded ? getFamilyMvpEntryPath(tabs, disabledTabs) : null),
     [configLoaded, tabs, disabledTabs]
@@ -321,7 +325,7 @@ export default function App({ onLogout }: AppProps) {
         setSelectedOwner(decodePathSegment(segs[1]));
       } else if (owners.length > 0) {
         // URL redirect is handled by the render-time <Navigate> in renderMainContent.
-        setSelectedOwner(owners[0].owner);
+        setSelectedOwner(findOwnerForUser(owners, user)?.owner ?? owners[0].owner);
       }
     } else if (newMode === 'instrument') {
       setSelectedGroup(segs[1] ?? '');
@@ -344,6 +348,7 @@ export default function App({ onLogout }: AppProps) {
     disabledTabs,
     owners,
     navigate,
+    user,
   ]);
 
   useEffect(() => {
@@ -407,7 +412,8 @@ export default function App({ onLogout }: AppProps) {
     const nextPath = getOwnerRootRedirectPath(
       location.pathname,
       selectedOwner,
-      owners
+      owners,
+      user
     );
     if (nextPath) {
       navigate(nextPath, { replace: true });
@@ -440,6 +446,7 @@ export default function App({ onLogout }: AppProps) {
     navigate,
     location.pathname,
     location.search,
+    user,
   ]);
 
   // data fetching based on route
@@ -526,7 +533,7 @@ export default function App({ onLogout }: AppProps) {
       !redirectSegs[1] &&
       owners.length > 0
     ) {
-      const owner = owners[0].owner;
+      const owner = findOwnerForUser(owners, user)?.owner ?? owners[0].owner;
       const destPath = redirectMode === 'performance'
         ? `/performance/${encodePathSegment(owner)}`
         : `/portfolio/${encodePathSegment(owner)}`;
--- a/frontend/src/MainApp.tsx
+++ b/frontend/src/MainApp.tsx
@@ -27,8 +27,9 @@ import InstallPwaPrompt from "./components/InstallPwaPrompt";
 import BackendUnavailableCard from "./components/BackendUnavailableCard";
 import lazyWithDelay from "./utils/lazyWithDelay";
 import PortfolioDashboardSkeleton from "./components/skeletons/PortfolioDashboardSkeleton";
-import { sanitizeOwners } from "./utils/owners";
+import { sanitizeOwners, findOwnerForUser } from "./utils/owners";
 import { isDefaultGroupSlug } from "./utils/groups";
+import { useAuth } from "./AuthContext";
 
 const ScreenerQuery = lazy(() => import("./pages/ScreenerQuery"));
 const TimeseriesEdit = lazy(() =>
@@ -50,6 +51,7 @@ export default function MainApp() {
   const location = useLocation();
   const { t } = useTranslation();
   const { mode, setMode, selectedOwner, setSelectedOwner, selectedGroup, setSelectedGroup } = useRoute();
+  const { user } = useAuth();
 
   const [owners, setOwners] = useState<OwnerSummary[]>([]);
   const [groups, setGroups] = useState<GroupSummary[]>([]);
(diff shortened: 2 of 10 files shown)


leonarduk:
@deepseek — thanks for the review. One clarification on the blocking item, plus a change I made based on your non-blocking suggestion.

**Re: the alleged race in `UserConfig.tsx` — this is a false positive.**

The reasoning inverts React's effect model. The config-fetch effect's dependency array is `[owner]`, and it's guarded by `if (owner)`:

```tsx
useEffect(() => {
  if (!owners.length) return;
  setOwner((current) => current || findOwnerForUser(owners, user)?.owner || '');
}, [owners, user]);

useEffect(() => {
  if (owner) { getUserConfig(owner); getApprovals(owner); }
}, [owner]);
```

On initial load:
1. `owners` resolves → the defaulting effect runs and calls `setOwner('jamie')`.
2. The config-fetch effect's initial run had `owner === ''`, so the `if (owner)` guard made it a no-op — no fetch for `""`.
3. `owner` changing from `''` → `'jamie'` is **exactly** what re-triggers the config-fetch effect, which now fetches for `'jamie'`.

The claim that "no re-fetch occurs because `owner` changed after that effect already ran" is backwards: an effect re-runs *because* a value in its dependency array changed. `owner` is that value.

**Re: your non-blocking test suggestion — taken (commit `5c95bf78`).**

You correctly spotted that the test masked behavior by returning `{}` from `getUserConfig` for any argument. I strengthened it to assert `getUserConfig` is called with the mapped owner (`"jamie"`) and never with `""` or the first owner (`"alex"`). It passes — which empirically demonstrates the effects sequence correctly, rather than just arguing it.

No code change to `UserConfig.tsx` since the underlying concern doesn't hold, but the new assertion makes that verifiable in CI.

PR #4717, opened 2026-07-06
feat(auth): enforce per-owner authorization on settings & approvals (#4710)

### LINKED ISSUE #4710: Settings page lets any logged-in user select and edit any owner's config (no authorization scoping)
## What
Split out from [#4707](https://github.com/leonarduk/allotmint/issues/4707) (closed as superseded by [#4709](https://github.com/leonarduk/allotmint/issues/4709) for the default-selection part). The remaining, still-open question: the `/settings` owner dropdown (`UserConfigPage` in `frontend/src/pages/UserConfig.tsx:146-157`) lists every owner in the system and lets the logged-in user select and edit any of them (hold-days, max-trades, exempt tickers/types, trade approvals) with no authorization check, client- or server-side.

Fixing the *default* selection (#4709) does not address this: even with a correct default, the dropdown still lets a user manually switch to and modify another owner's settings.

## Why
Whether this is acceptable depends on the intended access model:
- If AllotMint is single-owner-per-login, any owner other than the user's own shouldn't be selectable at all.
- If it's a family/household model where one login manages several owners, the dropdown should be scoped to the owners that login is authorized for, not every owner in the system.
Right now neither is enforced — the list is unscoped and, as far as could be determined from the frontend, the backend `user-config`/`approvals` endpoints have no visible per-owner authorization check either.

## How
- Determine the intended access model (single-owner vs. family-admin) — check with product intent/docs, or existing backend auth code, before changing behavior.
- If single-owner: restrict the dropdown to just the user's mapped owner (reusing whatever mapping #4709 introduces).
- If family-admin: scope the dropdown to only the owners the logged-in user is authorized to manage.
- Check `backend` routes for `/user-config/<owner>` and `/approvals/<owner>` (see `frontend/src/api.ts:1272-1276` for the client calls) for owner-level authorization; if none exists, that's a server-side gap independent of any frontend fix.

## Constraints
- Don't restrict functionality for an intentional multi-owner/family-admin use case without confirming that's not the design.
- A frontend-only restriction is not a substitute for backend authorization — both need to be addressed if the backend has no check.

## LLM tier
**Opus** — requires clarifying the intended access model before changing behavior, and may require backend authorization changes, not just a frontend dropdown filter.

## Success looks like
- [ ] The intended access model (single-owner vs. family-admin) is confirmed.
- [ ] The Settings dropdown only lists owners the logged-in user is authorized to manage under that model.
- [ ] `/user-config/<owner>` and `/approvals/<owner>` backend endpoints are confirmed to enforce (or are updated to enforce) owner-level authorization.

## Failure looks like
- The dropdown still lists every owner in the system regardless of who is logged in.
- A user can still load or mutate another owner's settings/approvals via a direct API call with no authorization check.

leonarduk:
## Summary

Fixes the server-side authorization gap in #4710: the `/user-config/<owner>` and `/accounts/<owner>/approvals` endpoints were **authenticated but not authorized**, so any logged-in user with a valid token could read or mutate **any** owner's config (hold-days, max-trades, exempt tickers/types) and trade approvals via a direct API call.

## Access model (confirmed from code, not assumed)

The intended model is already implemented in `backend/common/data_loader.py` (`_is_authorized` inside `_list_local_plots`, used to scope the `/owners` listing — and therefore the Settings dropdown). It is a **family/viewer model**: an identity may access an owner when it matches

- the owner id, **or**
- the owner's `email` in `person.json`, **or**
- an entry in the owner's `viewers` list.

This PR does **not** invent a new model — it applies the existing one to the endpoints that were missing it.

## What changed

- **New** `backend/common/authz.py` — reusable `identity_can_access_owner()` / `ensure_owner_access()` centralizing the owner-id / email / viewers rule.
- `backend/routes/user_config.py` — `GET`/`POST /user-config/<owner>` now call `ensure_owner_access`.
- `backend/routes/approvals.py` — `GET`/`POST`/`DELETE /accounts/<owner>/approvals` and `POST /accounts/<owner>/approval-requests` now call `ensure_owner_access`.
- Unauthorized access returns **403** (`PermissionDeniedError`, already wired to the app error handler).
- Authorization is a **no-op when `config.disable_auth` is set** (local/demo mode), matching how `/owners` exposes all owners there — so demo / no-auth flows are unaffected (issue constraint).

## Frontend / dropdown

No frontend change was required for the dropdown-scoping success criterion: `getOwners()` already calls the `/owners` endpoint (with the bearer token), which is **already scoped by the logged-in identity** via the same `_is_authorized` rule when auth is enabled. The remaining, real gap was the direct-API access on the two per-owner endpoints — now closed as defense at the API layer.

## Tests

- `tests/backend/common/test_authz.py` — unit tests for the helper (owner-id / email / viewers match, case-insensitivity, blank/None identity, malformed metadata, disable_auth no-op).
- `tests/test_user_config_route.py` — auth-enabled test: authorized owner → 200, other owner → 403 on read and write (and no mutation occurs).
- `tests/backend/test_approvals_route.py` — auth-enabled test: authorized owner → 200; other owner → 403 on GET/POST/DELETE (and no approvals file is created).

Existing tests run with `disable_auth: true` (repo default `config.yaml`) and are unaffected. `ruff`, `black`, and `isort` pass clean on the changed files.

## Scope note

Limited to the two endpoint groups named in #4710 (settings + approvals). Broadening owner-level authorization to other per-owner endpoints (e.g. `/portfolio/<owner>`) would be a larger, separate change and is out of scope here.

Closes #4710

### DIFF
--- a/backend/common/authz.py
+++ b/backend/common/authz.py
@@ -1,0 +1,72 @@
+"""Per-owner authorization helpers.
+
+AllotMint uses a family/viewer access model: an authenticated identity may
+access an owner's data when the identity matches the owner id, the owner's
+configured ``email``, or appears in the owner's ``viewers`` list. This mirrors
+the scoping already applied by :func:`backend.common.data_loader.list_plots`
+for the ``/owners`` listing (see ``_is_authorized`` in ``_list_local_plots``),
+and is applied here to per-owner endpoints that were previously authenticated
+but not authorized (e.g. ``/user-config/<owner>`` and
+``/accounts/<owner>/approvals``).
+
+When ``config.disable_auth`` is truthy the API runs in local/demo mode where
+the ``/owners`` listing exposes every owner and no identity is enforced; these
+helpers deliberately no-op in that mode so demo and no-auth flows are not
+restricted.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any, Mapping, Optional, Set
+
+from backend.common.data_loader import load_person_meta
+from backend.common.errors import PermissionDeniedError
+from backend.config import config
+
+_FORBIDDEN_DETAIL = "Not authorized for this owner"
+
+
+def _allowed_identities(owner: str, meta: Mapping[str, Any]) -> Set[str]:
+    """Return the lower-cased identities permitted to access ``owner``."""
+
+    allowed: Set[str] = set()
+    if isinstance(owner, str) and owner.strip():
+        allowed.add(owner.strip().lower())
+
+    email = meta.get("email") if isinstance(meta, Mapping) else None
+    if isinstance(email, str) and email.strip():
+        allowed.add(email.strip().lower())
+
+    viewers = meta.get("viewers") if isinstance(meta, Mapping) else None
+    if isinstance(viewers, list):
+        allowed.update(viewer.strip().lower() for viewer in viewers if isinstance(viewer, str) and viewer.strip())
+    return allowed
+
+
+def identity_can_access_owner(identity: Optional[str], owner: str, meta: Mapping[str, Any]) -> bool:
+    """Return ``True`` when ``identity`` may access ``owner`` given ``meta``."""
+
+    if not isinstance(identity, str) or not identity.strip():
+        return False
+    return identity.strip().lower() in _allowed_identities(owner, meta)
+
+
+def ensure_owner_access(
+    identity: Optional[str],
+    owner: str,
+    accounts_root: Optional[Path] = None,
+) -> None:
+    """Raise :class:`PermissionDeniedError` when ``identity`` lacks access.
+
+    No-op when authentication is disabled (local/demo mode). Otherwise the
+    owner's person metadata is consulted and the caller must match the owner
+    id, the owner's email, or a listed viewer.
+    """
+
+    if config.disable_auth:
+        return
+
+    meta = load_person_meta(owner, accounts_root)
+    if not identity_can_access_owner(identity, owner, meta):
+        raise PermissionDeniedError(_FORBIDDEN_DETAIL, safe_detail=_FORBIDDEN_DETAIL)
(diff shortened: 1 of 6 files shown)


leonarduk:
Thanks for the review, DeepSeek. Addressed in `5a7b8e1f`. Point-by-point:

**Blocking #1 — "`owner_dir.parent` is the wrong root for `load_person_meta`":** This one is incorrect. `_resolve_owner_dir(root, owner)` returns the owner's directory (`<accounts_root>/<owner>`), so `owner_dir.parent` *is* the accounts root that `load_person_meta(owner, data_root)` expects (it reads `data_root/<owner>/person.json`). The existing integration test authorized `alice` correctly with this exact call, which wouldn't happen if the metadata lookup were pointed at the wrong location. That said — see #2, I've reworked this code path anyway, so the routes now pass the resolved accounts root explicitly.

**Blocking #2 — "authorization runs after `_resolve_owner_dir`, leaking owner existence":** Agreed, this is a legitimate (if minor) hardening. Fixed: `ensure_owner_access` now runs immediately after `root = resolve_accounts_root(request)`, **before** `_resolve_owner_dir`, in all four approvals handlers. A non-existent owner now returns `403` rather than `404`, so existence is no longer distinguishable. Added a regression test (`GET /accounts/ghost/approvals → 403`).

**Blocking #3 — "`user_config` relies on `request.app.state.accounts_root` (AttributeError risk)":** `app.state.accounts_root` is always set at app creation (`backend/app.py:88`), so an `AttributeError` isn't actually reachable — but I agree the two route modules should resolve the root the same way. Both now use `resolve_accounts_root(request)` (user-config with `allow_missing=True` to preserve the prior "path may not exist yet" semantics).

**Test coverage:** Added the two you flagged as important — the **viewers-list** authorization path (`carol` lists our identity as a viewer → `200`) and the non-existent-owner case, both exercised through the real route handlers. The `disable_auth=True` route path is implicitly covered by the pre-existing route tests (repo default `config.yaml` has `disable_auth: true`, so those tests run through the no-op branch); the dedicated no-op assertion lives in `test_authz.py`.

**Follow-ups:** All four suggestions are reasonable. The frontend dropdown is already scoped server-side today (`getOwners()` → the identity-scoped `/owners` endpoint), so I don't think a frontend change is required for correctness, but a defense-in-depth filter + the authz-consolidation refactor (the `_is_authorized` rule is duplicated across `data_loader.py`/`portfolio.py`) are worth separate issues. I'll raise those.

`ruff`/`black`/`isort` clean on the changed files; full approvals/user-config suites green.

PR #4733, opened 2026-07-06
Verify Lambda invocations against authenticated smoke-test traffic

### LINKED ISSUE #3249: Lambda invocation metrics cannot be verified post-deployment
## Issue
Post-deployment verification cannot confirm Lambda invocations. While the absence of invocations is expected (due to Cognito JWT auth), this prevents truly verifying that the backend is operational.

## Details
From deployment logs:
```
No Lambda invocations in the last 5 minutes (expected: Cognito JWT auth blocks unauthenticated probes at API Gateway)
```

## Impact
- Cannot verify backend is actually receiving authenticated requests
- Post-deploy validation provides limited confidence
- No metrics-based verification of backend readiness

## Root Cause
The deployment validation relies on unauthenticated curl probes, which are blocked by API Gateway Cognito auth. While this is correct security-wise, it means we have no way to verify the backend is actually operational.

## Suggested Fix
- Provide authenticated test credentials in deployment pipeline
- Create a test endpoint that doesn't require auth (metrics only)
- Use backend logs instead of invocation metrics for validation

## Related Logs
- `deploy/28_Check recent BackendLambda CloudWatch errors.txt`

leonarduk:
## Summary
- The deploy job's `Check recent BackendLambda CloudWatch errors` step runs before the `smoke-test` job authenticates via Cognito and makes real requests, so it always saw 0 invocations regardless of backend health — it could never actually verify the backend was operational (#3249).
- Add a new step, `Verify authenticated invocations reached BackendLambda`, in the `smoke-test` job after the Cognito-authenticated frontend smoke tests (`frontend/tests/smoke.spec.ts`, `config-retry.spec.ts`) run. It checks CloudWatch invocation metrics over the smoke-test window and hard-fails if zero invocations were recorded despite real authenticated traffic — unlike the deploy-job check, where zero is expected and tolerated.
- The new check is skipped gracefully (not failed) when no smoke-test Cognito credentials are configured, matching the existing degraded-mode behavior of the token-fetch step.
- Clarified the comment on the existing deploy-job check to explain why it can never observe authenticated traffic and point to the new check.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-lambda.yml'))"` — valid YAML
- [x] `actionlint .github/workflows/deploy-lambda.yml` — no findings
- [ ] Observe the next production deploy run: `Verify authenticated invocations reached BackendLambda` should report a non-zero invocation count

Closes #3249

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -1098,8 +1098,10 @@ jobs:
         # Marks the start of the window checked by "Verify authenticated
         # invocations reached BackendLambda" below, so that window covers only
         # the smoke tests' own traffic rather than picking up unrelated
-        # invocations from earlier in the deploy.
-        run: echo "METRIC_START=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_ENV"
+        # invocations from earlier in the deploy. Subtracting a small buffer
+        # guards against clock skew between this runner and the CloudWatch
+        # datapoint boundary for an invocation made in the same second.
+        run: echo "METRIC_START=$(date -u -d '30 seconds ago' +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_ENV"
 
       - name: Run backend smoke tests
         run: npm run smoke:test
@@ -1112,6 +1114,7 @@ jobs:
         # VITE_ALLOTMINT_API_BASE so the built app calls the real deployed API.
         # SMOKE_AUTH_TOKEN and TEST_ID_TOKEN arrive via $GITHUB_ENV set by the
         # token-fetch step above; no explicit env: entries needed here.
+        id: frontend_smoke
         if: success() || failure()
         run: npm --prefix frontend run smoke:frontend
         env:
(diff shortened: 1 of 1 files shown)


leonarduk:
Thanks for the review — addressed the real issue and pushing back on two claims that don't hold up against the actual diff.

**Fixed (85b20969):**
- **`if: always()` false-positive risk** — legitimate. Changed the condition to `if: always() && steps.frontend_smoke.outcome == 'success'` (added `id: frontend_smoke` to the "Run frontend smoke tests" step). Now the invocation-count assertion only runs when we know authenticated traffic was actually sent successfully; a pre-existing test failure no longer risks being misreported as a broken JWT authorizer.
- **`METRIC_START` timing** — added a 30s buffer (`date -u -d '30 seconds ago' ...`) as cheap extra safety margin, even though CloudWatch's 60s metric period already makes exact-second exclusion unlikely.

**Not applicable — please re-check against the diff:**
- **"`SMOKE_AUTH_TOKEN` vs `TEST_ID_TOKEN` mismatch"**: the "Fetch Cognito smoke-test token" step writes *both* vars from the same `$TOKEN` in consecutive lines:
  ```
  printf 'SMOKE_AUTH_TOKEN=%s\n' "${TOKEN}" >> "$GITHUB_ENV"
  printf 'TEST_ID_TOKEN=%s\n' "${TOKEN}" >> "$GITHUB_ENV"
  ```
  `SMOKE_AUTH_TOKEN` is not dead — it's set whenever the token fetch succeeds, same as `TEST_ID_TOKEN`.
- **"Hardcoded `BackendLambdaStack` / should pass function name as deploy-job output"**: this matches an existing, deliberate pattern in this same job — the "Resolve backend URL for smoke tests" step re-queries CloudFormation directly rather than trusting the deploy job's output, specifically because cross-job outputs containing masked secret substrings arrive empty downstream (see that step's comment). Re-querying `BackendLambdaStack` for the function name follows the same reasoning; passing it as a job output would reintroduce the failure mode that step was written to avoid.

PR #4759, opened 2026-07-06
fix(news): warn when serving stale cache after empty _call() result

### LINKED ISSUE #4713: Warn when serving stale cache after successful but empty `_call()` result
# Warn when serving stale cache after successful but empty `_call()` result

## What

In `backend/routes/news.py`, the `_warn_if_stale` function is only called on the `RuntimeError` path (line 429). When `_call()` succeeds but returns an empty list (all three news sources return no items), the stale cache is served silently without any warning.

## Why

This creates a correctness risk where users receive stale news data without any indication. The staleness warning was specifically added to surface this issue, but the empty-result path bypasses it entirely. This could lead to agent confusion in downstream systems that rely on staleness signals, and makes debugging cache freshness issues harder.

## How

1. In `backend/routes/news.py`, after the `_call()` succeeds and returns a result, add a check for empty/falsy `payload` before returning the cached data
2. Call `_warn_if_stale(cached)` when `payload` is empty and `cached` is being returned
3. Add a test case `test_get_cached_news_warns_when_serving_stale_on_empty_result` following the pattern of the existing `test_get_cached_news_warns_when_serving_stale_on_quota` test

## Constraints

- Must not change the behavior when `_call()` returns non-empty results
- Must not affect the existing `RuntimeError` warning path
- Must not modify the `_warn_if_stale` function itself
- Out of scope: adding a `stale` flag to the API response (separate issue)

## LLM Tier

**Sonnet** — The change requires understanding the control flow between `_call()`, cache serving, and `_warn_if_stale`. While the code change is small, the test requires design judgment to properly mock the empty-result scenario and verify the warning is emitted.

## Success looks like

- `_warn_if_stale` is called when `_call()` returns empty and cached data exists
- New test passes: `test_get_cached_news_warns_when_serving_stale_on_empty_result`
- All existing tests continue to pass
- No change to behavior when `_call()` returns non-empty results

## Failure looks like

- Warning fires when `_call()` returns non-empty results (false positive)
- Warning fires when no cached data exists (unnecessary log noise)
- Existing `RuntimeError` warning path breaks
- Test doesn't actually verify the warning was emitted

_Follow-up from AI review of PR #4711._

leonarduk:
## What

In `backend/routes/news.py`, `_warn_if_stale` was only called on the `RuntimeError` path. When `_call()` succeeded but returned an empty list, stale cached data was silently overwritten with empty results _and_ the warning was never emitted.

### Backend (`backend/routes/news.py`)
- After `_trim_payload(payload)`, added a check: if `payload` is falsy _and_ `cached` exists, call `_warn_if_stale(page)`, schedule a refresh, and return `cached` instead of the empty result.

### Tests (`tests/routes/test_news_helpers.py`)
- Added `test_get_cached_news_warns_when_serving_stale_on_empty_result` following the pattern of the existing `test_get_cached_news_warns_when_serving_stale_on_quota` test: mocks `_try_consume_quota` to return `True` (quota available) and `_fetch_news` to return `[]`, then verifies the warning is logged and the cached payload is returned.

## Constraints verified

- ✅ No change when `_call()` returns non-empty results (false positive)
- ✅ No warning when no cached data exists (unnecessary log noise)
- ✅ Existing `RuntimeError` warning path unchanged
- ✅ `_warn_if_stale` function itself not modified
- ✅ New test passes: `test_get_cached_news_warns_when_serving_stale_on_empty_result`
- ✅ All 33 existing tests in `test_news_helpers.py` continue to pass
- ✅ All 24 existing tests in `test_news.py` + `test_news_route.py` continue to pass

Closes #4713

### DIFF
--- a/backend/routes/news.py
+++ b/backend/routes/news.py
@@ -436,8 +436,13 @@ def get_cached_news(
 
     payload = _trim_payload(payload)
     if not payload and cached is not None:
+        # Use NEWS_TTL as the initial delay rather than an immediate retry:
+        # with no delay the background loop's first call happens right away,
+        # and since quota wasn't exhausted (we got this far) it would likely
+        # fetch empty again and overwrite the still-good cached payload with
+        # an empty one via save_cache.
         _warn_if_stale(page)
-        _schedule_refresh()
+        _schedule_refresh(NEWS_TTL)
         return cached
 
     if cache_writer is not None:
--- a/backend/routes/news.py
+++ b/backend/routes/news.py
@@ -435,6 +435,11 @@ def get_cached_news(
         return []
 
     payload = _trim_payload(payload)
+    if not payload and cached is not None:
+        _warn_if_stale(page)
+        _schedule_refresh()
+        return cached
+
     if cache_writer is not None:
         cache_writer(page, payload)
     else:
--- a/tests/routes/test_news_helpers.py
+++ b/tests/routes/test_news_helpers.py
@@ -240,6 +240,29 @@ def test_get_cached_news_warns_when_serving_stale_on_quota(monkeypatch, caplog):
     assert any("Serving stale cached news" in rec.message for rec in caplog.records)
 
 
+def test_get_cached_news_warns_when_serving_stale_on_empty_result(monkeypatch, caplog):
+    """When _call() succeeds but returns empty and stale cache exists, warn and serve cache."""
+    cached_payload = [{"headline": "Old", "url": "https://example.com/old"}]
+
+    monkeypatch.setattr(news_module.page_cache, "load_cache", lambda page: cached_payload)
+    monkeypatch.setattr(news_module.page_cache, "is_stale", lambda page, ttl: True)
+    monkeypatch.setattr(news_module.page_cache, "time_until_stale", lambda page, ttl: 0)
+    monkeypatch.setattr(news_module.page_cache, "schedule_refresh", lambda *a, **k: None)
+    monkeypatch.setattr(
+        news_module.page_cache,
+        "cache_age",
+        lambda page: news_module.NEWS_MAX_STALENESS + 1,
+    )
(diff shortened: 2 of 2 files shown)


leonarduk:
Re: DeepSeek's "blocking" finding on `_schedule_refresh(NEWS_TTL)` (line ~442) — this is a false positive.

`_schedule_refresh` here is **not** `page_cache.schedule_refresh`. It's a local closure defined earlier in `get_cached_news`:

```python
def _schedule_refresh(initial_delay: float | None = None) -> None:
    page_cache.schedule_refresh(
        page,
        NEWS_TTL,
        _call,
        can_refresh=_can_request_news,
        initial_delay=initial_delay,
    )
```

It already closes over `page` and forwards it as the first positional arg to `page_cache.schedule_refresh`. `_schedule_refresh(NEWS_TTL)` fills the `initial_delay` parameter — exactly the intended behavior — not a "page key" argument. The same call pattern (`_schedule_refresh(delay)`) is already used earlier in the function for the fresh-cache path.

DeepSeek's suggested fix, `_schedule_refresh(page)`, would actually break things: `page` (a string) would be passed as `initial_delay`, and `page_cache.schedule_refresh`'s internal loop does `if initial_delay is not None and initial_delay > 0`, which raises `TypeError: '>' not supported between instances of 'str' and 'int'` on the first refresh cycle.

To close this out definitively, I pushed 4dc4af2a which strengthens `test_get_cached_news_warns_when_serving_stale_on_empty_result` to capture the actual `page_cache.schedule_refresh` call args and assert `page="news_CACHED"`, `initial_delay=NEWS_TTL` — confirming the call is correct. All 58 news-related tests pass.

(Separately, 7b9b0559 already fixed the real underlying issue: calling `_schedule_refresh()` with no delay would make the background loop retry immediately, likely re-fetching empty and overwriting the good cached payload on disk — that's why `NEWS_TTL` is passed as the delay now.)

PR #4780, opened 2026-07-06
fix(auth): resolve real caller identity when disable_auth trusts a Cognito token

### LINKED ISSUE #4750: No account mapping exists for steveleonard11@gmail.com and unmatched emails may silently fall back to the demo identity
## What
`steveleonard11@gmail.com` does not appear in any `data/accounts/<owner>/person.json` file, nor in `.env.example` — it is only referenced as a human escalation contact in `docs/TECHNICAL_SUPPORT.md:23-25`. Account resolution lives in `backend/auth.py`: `_allowed_emails` (lines 62-139) builds the allow-list from each account's `person.json`, and `get_active_user` (lines 433-475) falls back to `local_login_identity()` (`backend/config.py:533-541`, sourced from `LOCAL_LOGIN_EMAIL`) when auth is disabled. Separately, `smoke_identity` (`backend/config.py:127,416-420,544-552`) defaults to `"demo"` and is used by smoke-test flows (`scripts/update_smoke_endpoints.py:65-93`). With no `person.json` entry for this email, it's unclear from the code alone what account an actual login with this email resolves to — the original bug report's premise (that it silently maps to "demo") is plausible given the `smoke_identity` default but is not directly confirmed by the data present in this repo.

## Why
Correct account resolution is foundational — if it's possible for an unrecognized email to resolve to any account (demo or otherwise) instead of failing explicitly, that's a data-integrity and security concern, not just a UX one. This came up because the user (steveleonard11@gmail.com) is the account owner and this email should map to a real, named account.

## How
- Reproduce first: log in as `steveleonard11@gmail.com` against the deployed environment and capture what account/data actually loads and which code path (`_allowed_emails`, `get_active_user`, or the `local_login`/`smoke_identity` fallback) is responsible.
- If no `data/accounts/<owner>/person.json` currently owns this email, confirm with the user which account (e.g. a new or existing `steve` account) should own it, then add/update that `person.json` — treat this as a data change per the repo's caution around `data/` and smoke-test identities.
- Audit `get_active_user` in `backend/auth.py` to confirm whether an email with no matching account should reject the request rather than silently defaulting to `smoke_identity`/demo.

## Constraints
- `data/` changes are sensitive (per repo conventions) — confirm the intended owner/account name before editing any `person.json`.
- Must not break existing smoke-test flows that intentionally rely on `smoke_identity` defaulting to `"demo"`.

## LLM tier
**Sonnet** — requires reproducing against a live environment, tracing two related but distinct fallback paths in `auth.py`/`config.py`, and making a data change with user confirmation.

## Success looks like
- [ ] Reproduction confirms (or rules out) that this email currently resolves to the demo account
- [ ] `steveleonard11@gmail.com` maps deterministically to the confirmed correct account after the fix
- [ ] An email with no matching account entry fails explicitly instead of silently defaulting to a demo/smoke identity
- [ ] A backend test covers both a known-email resolution and an unmatched-email rejection

## Failure looks like
- The email still resolves to the demo account after the change
- Fixing this breaks the smoke-test flows that depend on the `"demo"` fallback

leonarduk:
## Summary

Reproduced and confirmed the issue's premise via code tracing: in the deployed Cognito/API Gateway path, `config.disable_auth=true` (API Gateway's Cognito JWT authorizer validates the token before the Lambda runs â€” see `docs/AUTH.md`). `get_current_user`/`get_active_user` tried to decode the Cognito ID token with the backend's own HS256 verifier, which always fails for an RS256 token, and fell back to a single shared `local_login_identity()`. That collapsed **every** real authenticated Cognito user into the same identity instead of resolving their own account â€” a real bug, not just a theoretical one.

- **`backend/auth.py`** â€” when a bearer token is present but not backend-decodable, extract its `email`/`sub` claim (safe: the signature was already verified upstream by the gateway) and authorize it against `_allowed_emails()`. An unrecognized email now gets an explicit 401/403 instead of silently becoming the shared local identity. Only a request with **no token at all** (bare local dev) still falls back to `local_login_identity()`.
- `smoke_identity()` / smoke-test flows are untouched â€” this only changes the token-present branch of the `disable_auth` path.

## Explicitly out of scope: account provisioning

This PR does **not** add `steveleonard11@gmail.com` to any `data/accounts/*/person.json`. An earlier revision of this branch did that, but `data/accounts/{alice,demo,demo-owner}` in this repo are synthetic fixtures for local dev/tests -- none of them carry a real email -- and committing a real user's email/name into a **public** git repository was a mistake, caught in review. Production account data belongs in the deployed data store (S3 in AWS, or an operator's own local data root), never in version control. Branch history was rewritten (squashed, force-pushed) to remove that data from the diff entirely.

Provisioning `steveleonard11@gmail.com` to a real account should go through the existing signup/admin-approval flow (`backend/common/signup_provision.py`) against the actual deployed environment, not a git commit. That's a separate, follow-up action for whoever operates the deployed instance -- not something this PR should do.

## Test plan
- [x] `tests/test_auth.py` / `tests/test_auth_get_active_user.py` â€” updated the two pre-existing tests that encoded the old (buggy) "any unparseable token â†’ shared local identity" behavior; added coverage for: resolving the real email claim from an unverified token, rejecting an unrecognized email claim (403), and rejecting a token with no readable claim (401), on both `get_current_user` and `get_active_user`.
- [x] `python -m pytest tests/test_auth.py tests/test_auth_get_active_user.py tests/backend/test_auth_module.py tests/test_google_auth.py tests/test_app.py tests/test_backend_api.py tests/test_accounts_api.py tests/test_main.py` â€” 152 passed, 4 skipped, 2 xfailed, 1 xpassed, no regressions.
- [x] `ruff check` / `black --check` clean on changed files (`backend/pyproject.toml` config).

Closes #4750

### DIFF
--- a/backend/auth.py
+++ b/backend/auth.py
@@ -244,24 +244,61 @@ def _user_from_token(token: str | None) -> str:
     return email
 
 
+def _identity_from_unverified_token(token: str) -> str | None:
+    """Return the ``email``/``sub`` claim from ``token`` without checking its signature.
+
+    Only called when ``config.disable_auth`` is true and ``decode_token`` (the
+    backend's own HS256 verifier) could not decode the token. In that
+    configuration the token is a Cognito ID token whose signature and
+    audience API Gateway's JWT authorizer already validated before this
+    Lambda ever ran (see docs/AUTH.md), so re-deriving the claims here without
+    re-checking the signature is safe — it is not the first check performed.
+    """
+    claims = _unverified_claims(token)
+    email = claims.get("email") or claims.get("sub")
+    return email if isinstance(email, str) and email else None
+
+
+def _resolve_identity_when_auth_disabled(token: str | None) -> str | None:
+    """Resolve the caller's identity when ``config.disable_auth`` is true.
+
+    A present token is trusted to have already been verified upstream: either
+    it is an app-signed backend JWT (Google flow), or a Cognito ID token
+    validated by API Gateway's JWT authorizer before invoking this Lambda
+    (Cognito flow). Either way its claimed email must belong to a provisioned
+    account — an unrecognized email is rejected explicitly rather than
+    silently collapsing every caller into the shared local/demo identity.
+    Only a request with no token at all (bare local dev, no client-side auth)
+    falls back to the configured local login identity.
+    """
+    if isinstance(token, str) and token:
+        user = decode_token(token)
+        if user:
+            return user
+        email = _identity_from_unverified_token(token)
+        if email is None:
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail="Invalid authentication credentials",
+            )
+        if email.lower() not in _allowed_emails():
+            logger.warning(
+                "Unauthorized identity on disable_auth path for %s",
+                sanitise_log_value(email),
+            )
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Unauthorized email")
+        return email
+    return local_login_identity()
+
+
 async def get_current_user(token: str | None = Depends(oauth2_scheme)) -> str:
     """Return the authenticated user extracted from the bearer token."""
 
     if config.disable_auth:
-        # When auth is disabled (e.g. DISABLE_AUTH=true with a Cognito JWT authorizer
-        # at API Gateway), try to extract identity from an app-signed JWT if present.
-        # Non-app tokens (e.g. Cognito ID tokens signed with RS256) will not decode
-        # with the HS256 app secret — decode_token returns None for those — so we fall
-        # back to the local identity rather than raising 401.
-        if isinstance(token, str) and token:
-            user = decode_token(token)
-            if user:
-                current_user.set(user)
-                return user
-        fallback = local_login_identity()
-        if fallback:
-            current_user.set(fallback)
-            return fallback
+        identity = _resolve_identity_when_auth_disabled(token)
+        if identity:
+            current_user.set(identity)
+            return identity
         current_user.set(None)
     token_str = token if isinstance(token, str) else None
     return _user_from_token(token_str)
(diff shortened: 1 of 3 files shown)


data/accounts/steveleonard11/person.json
```diff
@@ -0,0 +1,8 @@
+{
```
leonarduk: you are putting user data in the code base.  the data in s3 is not for git

leonarduk:
Thanks for the review. Responding to the two blocking points:

**"Missing data change (AC2 not met)"** — this is intentional, and the PR description now explains why (an earlier revision of this branch did add `data/accounts/steveleonard11/person.json`, but that was reverted after review caught it committing a real user's email and name into this **public** repo). `data/accounts/{alice,demo,demo-owner}` are synthetic test fixtures with no real emails; they are not "the repo's configured account registry" for real users — production account data belongs in the deployed data store (S3/operator's local data root), never in version control. Provisioning `steveleonard11@gmail.com` to a real account is a follow-up action for whoever operates the deployed environment, via the existing signup/admin-approval flow (`backend/common/signup_provision.py`) — not something this PR does. Branch history was squashed and force-pushed to remove that data from the diff entirely, so it won't show up in a re-review.

**"No reproduction evidence (AC1 not met)"** — reproduction here was via code tracing rather than hitting the live deployed environment (no access to it from this session): confirmed that `config.disable_auth=true` in the deployed Cognito/API Gateway path (per `docs/AUTH.md`), that a Cognito RS256 ID token always fails the backend's HS256 `decode_token`, and that the old code then fell back to a single shared `local_login_identity()` regardless of which real user authenticated. That's a confirmed code-level bug, not a data change needed for it to be true. I've added this framing to the PR description.

Also: thanks for flagging the missing no-token fallback regression test — `tests/test_auth_get_active_user.py::test_get_active_user_returns_local_identity` and `tests/test_auth.py::test_get_current_user_returns_local_identity_when_disabled` already cover that path (both pre-existing, unmodified by this PR) — no gap there.

PR #4783, opened 2026-07-06
feat(market): surface stale headline indicator on /market page

### LINKED ISSUE #4749: /market page has no visible staleness indicator for cached headlines
## What
`frontend/src/pages/MarketOverview.tsx` fetches headlines via `getMarketOverview()` (line 39) and renders each one's age using `formatPublishedAt(h.published_at)` (lines 6, 169), but nothing tells the user when a displayed headline is actually stale. On the backend, `backend/routes/market.py`'s `market_overview` endpoint (lines 282-295) builds the list via `_fetch_headlines()` (lines 237-272), which calls `get_cached_news(sym)` in `backend/routes/news.py:380-444`. That cache has `NEWS_TTL = 900` (15 min, `news.py:27`) and `NEWS_MAX_STALENESS = 24*60*60` (1 day, `news.py:28`, added in #4708). `_warn_if_stale()` (lines 360-377) only logs a server-side warning when served cache exceeds max staleness — `get_cached_news` will still serve that stale cache to the frontend (and schedule a background refresh) when quota is exhausted or a live fetch fails, with no flag reaching the UI.

## Why
PR #4708 (commit e2cd371c) added server-side staleness detection and `published_at` surfacing, but stopped at logging — there's still no user-facing signal, so a user has no way to tell whether the /market page reflects a quota outage or a fresh fetch. This is the concrete UI gap behind the "stale stories" report.

## How
- Have `get_cached_news`/`_fetch_headlines` pass through a staleness flag (derived from the existing `NEWS_MAX_STALENESS` check) alongside each headline or as part of the `market_overview` response.
- Surface that flag in `MarketOverview.tsx` as a visible badge/label per stale headline (or a page-level banner if the whole feed is stale), building on the existing `formatPublishedAt` rendering rather than duplicating staleness logic.
- Decide whether headlines older than `NEWS_MAX_STALENESS` should still render (flagged) or be filtered out entirely — confirm the desired behavior before implementing.

## Constraints
- Must not increase AlphaVantage API call volume — reuse the existing cache/TTL/staleness logic from #4708 rather than adding new fetch paths.
- Should degrade gracefully when quota is exhausted (already-cached data still shown, just flagged) rather than showing an empty page.

## LLM tier
**Sonnet** — requires threading a computed value from `news.py` through `market.py` into the frontend component and making a UX call on stale/filtered display.

## Success looks like
- [ ] `/market` visibly flags headlines older than `NEWS_MAX_STALENESS`
- [ ] Behavior when quota is exhausted or fetch fails is graceful (stale-but-flagged, not silent or blank)
- [ ] A frontend test covers rendering of a stale headline in `MarketOverview.tsx`

## Failure looks like
- Headlines older than the staleness threshold still render with no visible indication
- Quota exhaustion causes the page to render blank instead of falling back to flagged cached content

leonarduk:
## Summary
- `get_cached_news` (backend/routes/news.py) now tags every returned headline with a `stale` boolean, computed from the existing `NEWS_MAX_STALENESS` threshold introduced in #4708. No new fetch paths or extra AlphaVantage calls are added — this reuses the existing cache/TTL/staleness logic.
- `_fetch_headlines` (backend/routes/market.py) and the `/market/overview` response pass the flag through unchanged.
- `MarketOverview.tsx` renders a "Stale" badge next to any headline whose backing cache exceeds the staleness threshold, so a quota outage or repeated live-fetch failure is visible to users instead of only logged server-side.
- `getNews` (frontend/src/api.ts) also passes the flag through for consistency, since it shares the same `NewsItem` shape.
- Headlines are still rendered (flagged, not filtered) so quota exhaustion degrades gracefully rather than producing a blank page.

## Test plan
- [x] `python -m pytest tests/routes/test_news.py tests/routes/test_market.py tests/backend/routes/test_market.py -q` — 41 passed, including two new cases covering the stale/not-stale quota-exhaustion paths
- [x] `npx vitest run tests/unit/pages/MarketOverview.test.tsx` — 7 passed, including two new cases for the stale badge
- [x] `npx eslint` on the touched frontend files — clean
- [x] `ruff check` / `black --check` / `isort --check-only` (backend/pyproject.toml config) on the touched backend files — clean

Closes #4749

## Summary by CodeRabbit

* **New Features**
  * News headlines can now be marked as “stale” when cached data may be outdated, with a clear badge shown in the market view.
  * News responses now carry this status through to the app so the interface can reflect it consistently.

* **Bug Fixes**
  * Improved handling of cached news during quota limits so older items are still surfaced with the correct freshness indicator.
  * Updated coverage to ensure stale and non-stale headlines display correctly.

### DIFF
--- a/backend/routes/market.py
+++ b/backend/routes/market.py
@@ -234,16 +234,18 @@ def _fetch_uk_sectors() -> List[SectorPayload]:
     return out
 
 
-def _fetch_headlines() -> List[Dict[str, str]]:
+def _fetch_headlines() -> List[Dict[str, Any]]:
     """Fetch latest headlines for all known index symbols.
 
     Each index symbol is queried individually; results are aggregated and
-    de-duplicated by URL or headline.  If all requests fail, an error is logged
-    so callers have some visibility into the failure.
+    de-duplicated by URL or headline.  Each item carries the ``stale`` flag
+    from ``get_cached_news`` so the frontend can flag headlines served from
+    an aged cache.  If all requests fail, an error is logged so callers have
+    some visibility into the failure.
     """
 
     logger = logging.getLogger(__name__)
-    headlines: List[Dict[str, str]] = []
+    headlines: List[Dict[str, Any]] = []
     seen: set[str] = set()
     success = False
 
--- a/backend/routes/news.py
+++ b/backend/routes/news.py
@@ -52,7 +52,6 @@ _FINANCE_KEYWORDS = (
 )
 
 
-
 def _make_news_item(
     headline: object,
     url: object,
@@ -227,9 +226,7 @@ def fetch_news_yahoo(ticker: str) -> List[Dict[str, str]]:
             _parse_epoch_time(item.get("providerPublishTime")),
             item.get("publisher"),
         )
-        if news_item is not None and _is_finance_related(
-            news_item["headline"], clean_ticker, instrument_name
-        ):
+        if news_item is not None and _is_finance_related(news_item["headline"], clean_ticker, instrument_name):
             out.append(news_item)
     return out
 
@@ -253,9 +250,7 @@ def fetch_news_google(ticker: str) -> List[Dict[str, str]]:
             _parse_rss_time(item.findtext("pubDate")),
             item.findtext("source"),
         )
-        if news_item is not None and _is_finance_related(
-            news_item["headline"], clean_ticker, instrument_name
-        ):
+        if news_item is not None and _is_finance_related(news_item["headline"], clean_ticker, instrument_name):
             out.append(news_item)
     return out
 
@@ -357,6 +352,13 @@ def _fetch_news(ticker: str) -> List[Dict[str, str]]:
     return []
 
 
+def _is_cache_stale(page: str) -> bool:
+    """Return True when the cache for ``page`` exceeds ``NEWS_MAX_STALENESS``."""
+
+    age = page_cache.cache_age(page)
+    return age is not None and age > NEWS_MAX_STALENESS
+
+
 def _warn_if_stale(page: str) -> None:
     """Log a warning when the cache being served is dangerously old.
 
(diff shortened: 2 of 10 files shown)


leonarduk:
Thanks for the review — responding to the blocking finding and the two open questions.

**1. \"`_is_cache_stale` references undefined variable `page`\" — not accurate.**

`_is_cache_stale` is defined with `page` as an explicit parameter:

```python
def _is_cache_stale(page: str) -> bool:
    """Return True when the cache for ``page`` exceeds ``NEWS_MAX_STALENESS``."""
    age = page_cache.cache_age(page)
    return age is not None and age > NEWS_MAX_STALENESS
```

([backend/routes/news.py:355-359](backend/routes/news.py#L355-L359))

It's called from `get_cached_news` as `_is_cache_stale(page)`, passing in the `page = f"news_{tkr}"` local already in scope at that point ([backend/routes/news.py:417](backend/routes/news.py#L417), [:447](backend/routes/news.py#L447)). This is an ordinary function call with an argument, not a closure/free-variable reference — there's no `NameError`. `python -m pytest tests/routes/test_news.py -q` (41 tests, including the two new stale-flag tests that exercise this exact code path) passes clean, which wouldn't be possible if this raised at runtime.

**2. \"Verify `getMarketOverview` passes through `stale`\" — confirmed, it does.**

`getMarketOverview` in `frontend/src/api.ts` is a direct passthrough with no field mapping:

```ts
export const getMarketOverview = () =>
  fetchJson<MarketOverview>(`${API_BASE}/market/overview`);
```

Unlike `getNews`, it doesn't reshape the response, so whatever `market_overview` returns (including `stale` on each headline) reaches the frontend unchanged. Covered by the `MarketOverview.test.tsx` stale-badge tests, which mock `getMarketOverview` returning `stale: true`/`false` and assert on the rendered badge.

**3. Missing integration test for `market_overview` → `stale` propagation — fair point, added.**

I'll follow up with a test exercising `_fetch_headlines`/`market_overview` end-to-end with a stale `get_cached_news` result, since the existing tests only assert `get_cached_news` and `_fetch_headlines` independently.

No code changes needed for the blocking finding since it doesn't reproduce; will push the added integration test shortly.

leonarduk:
CI caught two test files I'd missed when grep'ing for direct assertions on the news payload shape: [tests/backend/test_news_route.py](tests/backend/test_news_route.py) and [tests/routes/test_news_helpers.py](tests/routes/test_news_helpers.py) both assert exact dicts returned from `/news` and `get_cached_news`, which now include the `stale` key.

Fixed in d079ed1:
- `test_get_news_falls_back` / `test_get_news_trims_payload_and_caches` — expect `stale: False` on the fresh-fetch path.
- `test_get_cached_news_warns_when_serving_stale_on_quota` — expects `stale: True`, since this test's `cache_age` mock returns `NEWS_MAX_STALENESS + 1` (this is the real bug-catching case: previously the test only checked the log warning, not that the flag correctly turns on).
- `test_get_cached_news_reuses_fresh_cache` — expects `stale: False` on the fresh-cache-hit path.

Full suite (`pytest tests/ -q`, deselecting the two pre-existing unrelated `test_review_comment.py` bash/pipefail failures caused by a Windows Git Bash quirk) is green: 2748 passed.

PR #4784, opened 2026-07-06
test: strengthen deploy-workflow assertions and add fetch-cloudwatch-logs.sh coverage

### LINKED ISSUE #4762: test: strengthen deploy-workflow assertions and add fetch-cloudwatch-logs.sh coverage
## What
Add and strengthen test coverage for the deploy tooling: bats tests for `fetch-cloudwatch-logs.sh`, a CI smoke/dry-run test for the same script, log-output verification tests, and several assertion strengthenings in existing Python deploy-workflow tests.

Covers: #3785, #3807, #4075, #4076, #4130, #4141, #4143

## Why
Several existing tests assert on loose conditions (string/list-membership checks) that can pass coincidentally without verifying the actual behaviour, and `fetch-cloudwatch-logs.sh` currently has no automated coverage at all for its AccessDenied/empty/success paths.

## How
- Add bats unit tests for `scripts/bash/fetch-cloudwatch-logs.sh` covering AccessDenied, empty-result, and success paths (#3807).
- Add an integration/dry-run smoke test for `fetch-cloudwatch-logs.sh` in CI so a real (mocked) invocation is exercised, not just unit-level assertions (#4075).
- Add tests verifying log output content/format for the deploy scripts (#4076).
- Strengthen `check_cdk_diff_iam_removals` assertions to use `' '.join(row)` instead of relying on list-membership coincidence (#3785).
- Strengthen `test_deploy_lambda_workflow`'s `2>&1` assertion to verify placement on the `aws s3api` line specifically, not just string presence (#4130).
- Assert `head_exit` is captured before the `head_output` subshell runs in the deploy-lambda workflow test (#4141).
- Strengthen the continue-on-error absence check to assert the key is not present in the step dict, rather than a string search (#4143).

## Constraints
- No live AWS/GitHub network calls — mock `aws`/`gh` CLI calls per CLAUDE.md.
- Bats tests must run in the same CI environment as the existing shell test suite (check for an existing bats runner/config before introducing a new one).
- Assertion strengthenings must not change what real-world conditions the tests catch — only tighten false-positive risk.

## LLM tier
**Sonnet** — spans two test frameworks (bats and pytest) and requires understanding existing mocking patterns well enough not to weaken coverage while tightening assertions.

## Success looks like
- [ ] bats tests exist for `fetch-cloudwatch-logs.sh` covering AccessDenied, empty-result, and success paths
- [ ] A CI smoke/dry-run test exercises `fetch-cloudwatch-logs.sh` end-to-end with mocked AWS calls
- [ ] New tests verify deploy script log output content, not just exit codes
- [ ] `check_cdk_diff_iam_removals` assertions use `' '.join(row)` instead of list-membership checks
- [ ] `test_deploy_lambda_workflow`'s `2>&1` assertion verifies placement on the `aws s3api` line
- [ ] `head_exit`-before-`head_output` ordering is explicitly asserted
- [ ] continue-on-error absence check asserts key-not-in-dict rather than string search
- [ ] `make lint` and the full pytest/bats suites pass clean

## Failure looks like
- New bats tests that don't actually invoke the mocked failure paths (AccessDenied/empty), giving false confidence
- Assertion strengthening that makes a test too strict and breaks on unrelated, valid script output changes
- Test changes that pass locally but aren't wired into the CI job that's supposed to run them

leonarduk:
## Summary
- Add bats tests (`tests/bash/fetch-cloudwatch-logs.bats`) for `scripts/bash/fetch-cloudwatch-logs.sh`, covering the AccessDenied, generic-error, empty-result, and success/filtering paths plus argument validation. The `aws` CLI is mocked, so no network access is needed.
- Wire a new `shell-tests` job into `.github/workflows/ci.yml` that runs the bats suite via `npx bats` on every push/PR — a dry-run smoke test with no live AWS calls.
- Fix `tests/scripts/test_check_cdk_diff_iam_removals.py`: four assertions checked `"x" in unmatched[0]` (list-membership, which only passed by coincidence) instead of substring containment. Changed to `"x" in " ".join(unmatched[0])`.
- Add an assertion in `tests/scripts/test_deploy_lambda_workflow.py` that `head_exit` is captured in the same statement as the `head_output` subshell in the deploy-lambda workflow (guards against a future refactor leaving `head_exit` stale).

Note: the `2>&1` placement and `continue-on-error` absence assertions called for in the issue (sub-issues #4130/#4143) were already present in `test_deploy_lambda_workflow.py` from a prior commit — verified rather than re-added.

## Test plan
- [x] `npx bats tests/bash/fetch-cloudwatch-logs.bats` — 6/6 pass locally
- [x] `pytest tests/scripts/test_check_cdk_diff_iam_removals.py tests/scripts/test_deploy_lambda_workflow.py` — 13/13 pass
- [x] Full `pytest tests/` suite — 2745 passed, 6 pre-existing failures in `tests/test_review_comment.py` unrelated to this change (confirmed failing identically on `main`, caused by a `bash`/WSL `pipefail` option mismatch in the local environment)
- [x] `ruff check` / `black --check` on the two modified test files — clean (pre-existing repo-wide lint/format debt in unrelated files left untouched)

Closes #4762

## Summary by CodeRabbit

* **Tests**
  * Added coverage for shell script log-fetching behaviour, including warnings, empty results, successful filtering, and input validation.
  * Tightened regression checks for deployment and IAM-related scripts to better match expected output and error handling.
* **Chores**
  * Expanded CI to run Bash test suites as part of automated checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,13 @@ jobs:
         with:
           node-version: '24'
       # bats-core is fetched on demand via npx rather than assuming it's
-      # preinstalled on the runner image; all aws CLI calls in these tests
-      # are mocked (see tests/bash/fetch-cloudwatch-logs.bats), so this is a
-      # pure dry-run with no live AWS network access.
+      # preinstalled on the runner image; pin the version so an upstream
+      # bats-core release can't silently change CI behaviour. All aws CLI
+      # calls in these tests are mocked (see
+      # tests/bash/fetch-cloudwatch-logs.bats), so this is a pure dry-run
+      # with no live AWS network access.
       - name: Run bats tests for deploy shell scripts
-        run: npx --yes bats tests/bash/*.bats
+        run: npx --yes bats@1.13.0 tests/bash/*.bats
 
   validate-backend-deps:
     name: Validate backend/requirements.txt (dry-run)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,21 @@ jobs:
       - name: CDK infrastructure tests
         run: pytest cdk/tests/ --no-cov
 
+  shell-tests:
+    name: Bash script tests (bats)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24'
+      # bats-core is fetched on demand via npx rather than assuming it's
+      # preinstalled on the runner image; all aws CLI calls in these tests
+      # are mocked (see tests/bash/fetch-cloudwatch-logs.bats), so this is a
+      # pure dry-run with no live AWS network access.
+      - name: Run bats tests for deploy shell scripts
+        run: npx --yes bats tests/bash/*.bats
+
   validate-backend-deps:
     name: Validate backend/requirements.txt (dry-run)
     runs-on: ubuntu-latest
--- a/tests/bash/fetch-cloudwatch-logs.bats
+++ b/tests/bash/fetch-cloudwatch-logs.bats
@@ -1,0 +1,83 @@
+#!/usr/bin/env bats
+# Unit tests for scripts/bash/fetch-cloudwatch-logs.sh covering the
+# AccessDenied, empty-result, and success paths documented at the top of
+# the script under test. See #3807, #4075, #4076.
+
+SCRIPT="$BATS_TEST_DIRNAME/../../scripts/bash/fetch-cloudwatch-logs.sh"
+
+setup() {
+  FAKE_BIN="$(mktemp -d)"
+  PATH="$FAKE_BIN:$PATH"
+}
+
+teardown() {
+  rm -rf "$FAKE_BIN"
+}
+
+# Writes a fake `aws` executable to $FAKE_BIN that stands in for
+# `aws logs filter-log-events ...`. The script under test only inspects
+# stdout (see its "Capture stdout ... separately from stderr" comment), so
+# STDOUT_TEXT is what fetch-cloudwatch-logs.sh actually sees.
+write_fake_aws() {
+  local stdout_text="$1"
+  local exit_code="$2"
+  cat > "$FAKE_BIN/aws" <<EOF
+#!/usr/bin/env bash
+printf '%s\n' "$stdout_text"
+exit $exit_code
+EOF
+  chmod +x "$FAKE_BIN/aws"
+}
+
+@test "exits 0 and emits a scoped warning on AccessDeniedException" {
+  write_fake_aws "An error occurred (AccessDeniedException) when calling the FilterLogEvents operation" 1
+
+  run "$SCRIPT" "/aws/lambda/my-fn" 600
+
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"::warning::logs:FilterLogEvents denied for /aws/lambda/my-fn; CloudWatch logs not available (see #3742)"* ]]
+}
+
+@test "exits 0 and emits a generic warning on a non-AccessDenied API error" {
+  write_fake_aws "An error occurred (ThrottlingException) when calling the FilterLogEvents operation" 1
+
+  run "$SCRIPT" "/aws/lambda/my-fn" 600
+
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"::warning::Failed to fetch CloudWatch logs for /aws/lambda/my-fn:"* ]]
+  [[ "$output" == *"ThrottlingException"* ]]
+}
+
+@test "prints a friendly message when the log group has no matching events" {
+  write_fake_aws "None" 0
+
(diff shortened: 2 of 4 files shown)


leonarduk:
## Response to DeepSeek review

Two legitimate points fixed in d998e20b:

- **Pinned `bats-core` version** — `npx --yes bats@1.13.0` instead of unpinned `bats`, so an upstream release can't silently change CI behaviour.
- **Made the `head_exit` ordering check explicit** — the assertion now verifies `head_output`'s subshell is fully closed (`)"`) *before* `head_exit` is captured, rather than only checking both names appear in the same statement.

The rest of the review doesn't hold up against the actual diff/repo state:

- **"Missing bats test file"** — `tests/bash/fetch-cloudwatch-logs.bats` is in the diff (`gh pr diff 4784 --name-only` lists it, and it was part of commit d665f3ad). It has 6 tests: AccessDenied, a generic non-AccessDenied API error, empty-result, success/`None`-filtering, plus the two arg-validation guards — all run against a mocked `aws` CLI.
- **"Missing log-output verification tests"** — the "filters out None entries and prints real log messages on success" test in that file *is* the log-output content/format check (asserts `None` is stripped and real messages pass through).
- **"Missing CI smoke/dry-run test beyond unit-level assertions"** — the new `shell-tests` CI job runs that same bats suite against a mocked `aws` end-to-end on every push/PR; that mocked, no-network run is the dry-run smoke test called for.
- **`actions/checkout@v7` / `setup-node@v6` + Node 24 "version mismatch"** — this is the repo-wide convention, not a deviation. Every other workflow in `.github/workflows/` (ci.yml's own existing jobs, deploy-lambda.yml, frontend-tests.yml, iac-validation.yml, cdk-dry-run.yml, etc.) uses exactly `checkout@v7` and `setup-node@v6` with `node-version: '24'`. Introducing `v6`/`22` here would be the actual inconsistency.
- **"No existing bats runner to reuse"** — confirmed by search; no `.bats` files or bats config existed anywhere in the repo before this PR, so introducing one is correct per the issue's own constraint ("check for an existing bats runner/config before introducing a new one").

Local verification after the fixes: `npx bats@1.13.0 tests/bash/*.bats` → 6/6 pass; `pytest tests/scripts/test_deploy_lambda_workflow.py tests/scripts/test_check_cdk_diff_iam_removals.py` → 13/13 pass; `ci.yml` YAML parses cleanly.

leonarduk:
This is a re-post of the same three claims from the previous DeepSeek pass, already addressed in [my earlier comment](#issuecomment-4902465775) — nothing new here, so no further changes:

1. **`.bats` file "missing"** — still present: [`tests/bash/fetch-cloudwatch-logs.bats`](https://github.com/leonarduk/allotmint/blob/2a9feda6/tests/bash/fetch-cloudwatch-logs.bats), 6 tests covering AccessDenied, a generic API error, empty-result, success/log-content filtering, and arg validation, run against a mocked `aws` CLI. It's likely not visible to this reviewer because the incremental diff for the latest commit (`2a9feda6`) only touches 6 lines inside that already-existing file (switching `run "$SCRIPT"` → `run bash "$SCRIPT"` to fix a Linux CI permission issue), not the file's full content.
2. **`checkout@v6`/Node 22** — still incorrect; `@v7` + Node `24` is the convention used by every other job in `ci.yml` and every other workflow file in this repo (verified again just now). `@v6`/`22` would be the actual regression.
3. Log-output verification — same bats file's "filters out None entries and prints real log messages on success" test covers this.

Separately (unrelated to this review): the bats suite was failing on the Linux CI runner due to a permission-bit mismatch (script is tracked as `100644`, invoked via `run "$SCRIPT"` directly rather than `bash "$SCRIPT"` as production does) — fixed in `2a9feda6`, 6/6 passing locally after the fix.

leonarduk:
Third pass, same claims — already rebutted with evidence in [comment 1](#issuecomment-4902465775) and [comment 2](#issuecomment-4902627733). No changes made:

- `tests/bash/fetch-cloudwatch-logs.bats` exists and is unchanged since d665f3ad: https://github.com/leonarduk/allotmint/blob/2a9feda6/tests/bash/fetch-cloudwatch-logs.bats
- `checkout@v7` + Node `24` is the actual repo-wide convention (every workflow file uses it); `@v6`/Node 22 would be the regression, not the fix.
- The `2>&1`-placement and continue-on-error key-absence assertions this review says are "missing" were already present in `test_deploy_lambda_workflow.py` on `main` before this PR branched (added in #4114) — this PR only added the `head_exit`-ordering assertion on top of them. They're still there in the current diff, lines 72–86.
- New claim this round — "continue-on-error check still uses string search, not key-not-in-dict" — is also incorrect: `"continue-on-error" not in verify_step` *is* a dict key-absence check (`verify_step` is a `dict` parsed from YAML by `yaml.safe_load`, not a string), which is exactly what AC #4143 asked for.

This reviewer appears to be evaluating a truncated/incremental diff view rather than the full accumulated PR + repo state each time, which is producing the same false negatives on every pass. Flagging for @leonarduk to review manually rather than continuing to reply to each repeat.

PR #4801, opened 2026-07-07
fix(frontend): make logout reachable regardless of Menu prop wiring

### LINKED ISSUE #4751: Logout button is hidden behind a menu category and depends on onLogout being wired
## What
The logout control in `frontend/src/components/Menu.tsx:280-291` only renders when two conditions both hold: the menu category is `'preferences'`, and an `onLogout` callback prop was actually passed into `Menu`. If whatever parent renders `Menu` for a given page/auth mode doesn't thread `onLogout` through, the button silently disappears with no fallback and no visible error.

## Why
Users need a reliable way to end their session from any page. Because visibility is gated on both menu navigation and a prop being correctly threaded, the button can go missing in specific auth modes or pages without anyone noticing until a user reports it — which is what triggered this issue.

## How
- Trace every place `Menu` is mounted (e.g. `frontend/src/App.tsx` / the top-level app shell) and confirm `onLogout` is passed consistently, including in Cognito-hosted-UI mode and local/disable_auth dev mode.
- Confirm the click handler correctly invokes `cognitoLogout()` in `frontend/src/awsUiAuth.ts:158-174`, which builds the Cognito hosted-UI logout redirect and calls `window.location.assign(...)`.
- Decide whether logout should stay nested under the `preferences` category or be promoted somewhere more discoverable (e.g. always-visible in the top-level menu) — confirm with the user before changing IA.
- Add a frontend test asserting the logout button renders for an authenticated session in every auth mode the app supports.

## Constraints
- Must continue to work under both Cognito-hosted-UI auth and local `disable_auth` dev mode — don't special-case one and break the other.
- Don't restructure unrelated menu categories while fixing this.

## LLM tier
**Sonnet** — requires tracing prop threading across `Menu.tsx` and its parent(s), reasoning about two distinct auth modes, and possibly an IA decision; not a single-line fix.

## Success looks like
- [ ] Logout is visibly reachable for every authenticated session, in both Cognito and local dev auth modes
- [ ] Clicking logout calls `cognitoLogout()` (or the local-mode equivalent) and the session is actually cleared
- [ ] A frontend test fails if `onLogout` is ever left unwired for a rendered `Menu` instance

## Failure looks like
- Logout button still disappears in one auth mode because `onLogout` isn't passed in that render path
- Logout click doesn't clear the Cognito session or leaves the user still authenticated after redirect

leonarduk:
## Summary
- `Menu` only rendered a logout control when both the `preferences` menu category applied **and** an `onLogout` callback was explicitly passed in. `AlertSettings` (`/alert-settings`) and `PerformanceDiagnostics` (`/performance/:owner/diagnostics`) are mounted as standalone routes in `frontend/src/main.tsx` (outside `App`'s tree), so they render `<Menu />` with no `onLogout` prop — the button silently disappeared on those pages with no error, which is exactly the bug reported.
- Rather than hand-threading `onLogout` through every standalone page (which would need to be repeated correctly for every future page that mounts `Menu`), the real logout implementation (already correct for both Cognito hosted-UI and local `disable_auth` dev mode, `frontend/src/awsUiAuth.ts:158-174`) is now registered into `AuthContext` once `main.tsx`'s `Root` resolves the auth mode.
- `Menu` (`frontend/src/components/Menu.tsx`) falls back to that context-provided logout whenever no `onLogout` prop is supplied, so the control is reachable from any page by construction — it can't silently go missing again because a parent forgot to wire a prop.
- The logout button's position (nested under `preferences`) is unchanged — no IA change, per the issue's constraint.

## Test plan
- [x] `frontend/tests/unit/components/Menu.test.tsx`: added a test asserting logout renders and works via `AuthContext` alone (no `onLogout` prop) — the regression this issue reports — plus a test confirming logout stays hidden when neither the prop nor context value is available (unauthenticated/no-provider case).
- [x] `npx vitest run tests/unit/components/Menu.test.tsx` — 11/11 pass.
- [x] `npx vitest run tests/unit` — 549/562 pass, remaining 13 lost to sandbox worker-fork crashes (no assertion failures); the previously-flaky files pass 48/48 in isolation.
- [x] `npx tsc --noEmit -p tsconfig.json` — clean.
- [x] `npm run lint` — clean.

Closes #4751

## Summary by CodeRabbit

* **New Features**
  * Logout is now exposed via the shared account context, enabling sign-out from multiple parts of the app.
  * The menu can automatically show a logout option when a logout action is available through the account context.

* **Bug Fixes**
  * Corrected when the logout item appears in the menu, including cases where no direct logout prop is provided.

* **Tests**
  * Added unit coverage for logout rendering and behaviour with and without account-context logout.

* **Chores**
  * Improved test mocking stability and simplified the test worker configuration.


leonarduk:
Thanks for the review — content-wise it's an approve, matching the acceptance criteria. The **DeepSeek AI code review** check itself failed for a different reason: the review body ended the verdict as a plain bullet (\`- APPROVE\`) instead of the bold \`**APPROVE**\` the prompt in [\`review_common.py\`](.github/scripts/review_common.py) asks for, so [\`extract_verdict.py\`](.github/scripts/extract_verdict.py) correctly reported "no valid verdict found" and exited 1 — a parsing false-negative, not a real requested-changes verdict. I've re-run that job.

This is a pre-existing reliability gap in the shared AI-review verdict parser (any provider that emits an unbolded verdict trips it), not something introduced by this PR, so I'm not folding a fix into this branch — flagging it as a follow-up instead rather than expanding scope here.

No other blocking findings from the review, so no code changes needed on this PR as a result.

PR #4857, opened 2026-07-07
Audit CfnOutput consumption in deploy-lambda.yml and document HTTP/2 review

### LINKED ISSUE #4830: Audit CfnOutput consumption in deploy-lambda.yml and review HTTP/2 response handling
## What
Audit every `aws cloudformation describe-stacks --query ...OutputValue` call in `.github/workflows/deploy-lambda.yml` (e.g. `BackendApiUrl`, `UserPoolId`-style outputs) for bare-value assumptions such as missing `--output text` or unstripped whitespace/newlines, and separately review whether the workflow's curl-based checks need explicit handling for HTTP/2 responses.

## Why
Several `CfnOutput` values (`BackendApiUrl`, `DataBucketName`, `UiAuthUserPoolId`, `UiAuthUserPoolClientId`, `SmokeTestUserPoolClientId`, `UiAuthDomain`, etc., defined across `cdk/stacks/backend_lambda_stack.py` and `cdk/stacks/static_site_stack.py`) are consumed downstream in the deploy workflow via `describe-stacks --query ... --output text`, sometimes with explicit `"None"`/empty-string guards, sometimes without. A silent bare-value assumption (missing `--output text`, an unstripped trailing newline, or a JSON-array output where a scalar was expected) would cause a downstream step to silently pass a malformed value (e.g. `"None\n"` as an origin) rather than fail loudly. Separately, if API Gateway or CloudFront ever responds over HTTP/2, curl's default header-name case-folding and status-line parsing could behave subtly differently in the CORS smoke-test steps — this should be confirmed as a non-issue or fixed. This consolidates sub-issues #4080 and #4098 from #4734.

## How
1. Grep `.github/workflows/deploy-lambda.yml` for every `describe-stacks --query` invocation and list which outputs are consumed, whether each uses `--output text`, and whether the consuming step validates for `None`/empty before use (the existing pattern at ~lines 434-435, 449, 685 already does this for some outputs — check all of them for consistency).
2. For any output missing the `--output text` flag or a None/empty guard, add it, following the existing guarded pattern.
3. Confirm what protocol version curl negotiates against the deployed API Gateway HTTP API / CloudFront distribution (`curl -v` shows the negotiated protocol) and whether the existing header-matching (`grep -i`) is robust regardless of HTTP/1.1 vs HTTP/2 — HTTP/2 always lowercases header names, and curl already normalises this for `-D`/header dumps, so this step may conclude "no change needed, document why."
4. Document the outcome of step 3 as a comment in the workflow near the smoke-test curl step, whichever the conclusion.

## Constraints
- Do not change deploy behavior for outputs that already have correct guards — this is an audit-and-patch-the-gaps task, not a rewrite of the CfnOutput consumption pattern.
- Keep changes localized to `.github/workflows/deploy-lambda.yml` (and CDK output definitions only if a genuinely missing output is found).

## LLM tier
**Haiku** — mechanical grep-and-check audit with a well-defined existing pattern to follow; the HTTP/2 review is a documentation-only conclusion in the common case.

## Success looks like
- [ ] Every `describe-stacks --query ...OutputValue` call in `deploy-lambda.yml` uses `--output text` and is guarded against `None`/empty values
- [ ] A documented conclusion on whether HTTP/2 response handling needs any change in the CORS smoke-test curl calls

## Failure looks like
- A guard is added that changes behavior for an output that was already working correctly, breaking an unrelated deploy step
- The HTTP/2 review is skipped entirely rather than concluded either way

leonarduk:
## Summary
- Audited every `describe-stacks`/`list-stack-resources --query ...OutputValue` call in `.github/workflows/deploy-lambda.yml` (25 call sites). All already use `--output text` and guard against `None`/empty, with one gap: the "Fetch Lambda logs from smoke test" step was missing the `2>/dev/null || echo ""` fallback its sibling deploy-job step uses, so a transient `describe-stacks` API error would hard-fail the step instead of degrading gracefully to "not found; skipping" — now fixed for consistency.
- Reviewed whether the CORS smoke-test curl checks need explicit HTTP/2 handling. Conclusion: none of the curl invocations request HTTP/2 (no `--http2`/`--http2-prior-knowledge`), so they negotiate HTTP/1.1 over TLS regardless of what API Gateway/CloudFront support; and the existing header-matching (`grep -i` + `tr -d '\r'`) is already case-insensitive and CR-tolerant, which is what HTTP/2's lowercase headers and LF-only lines would need anyway. Documented this conclusion as a comment near the CORS check so it doesn't need re-litigating later.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/deploy-lambda.yml'))"` — YAML parses cleanly
- [ ] Deploy workflow itself only runs on tag push / manual dispatch to production, so behavior is verified by inspection here; no local repro of the AWS-dependent steps

Closes #4830

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -587,6 +587,13 @@ jobs:
         # `|| true` on the curl assignment keeps set -e from short-circuiting
         # past the explicit status check below when --fail trips on a non-2xx
         # response.
+        # HTTP/2 review (issue #4830): none of the curl calls in this workflow pass
+        # --http2/--http2-prior-knowledge, so curl negotiates HTTP/1.1 over TLS here
+        # regardless of what API Gateway/CloudFront would otherwise support. Even if
+        # that changed, the header-matching below (`grep -i` + `tr -d '\r'`) is
+        # already case-insensitive and CR-tolerant, which is exactly what HTTP/2's
+        # mandatory-lowercase header names and LF-only line endings require. No
+        # change needed; this applies to the OPTIONS preflight check below too.
         run: |
           set -euo pipefail
           echo "Checking CORS headers on ${BACKEND_URL%/}/config for Origin: ${FRONTEND_ORIGIN}"
@@ -1302,7 +1309,7 @@ jobs:
           log_group="$(aws cloudformation describe-stacks \
             --stack-name BackendLambdaStack \
             --query "Stacks[0].Outputs[?OutputKey=='BackendLambdaLogGroupName'].OutputValue" \
-            --output text)"
+            --output text 2>/dev/null || echo "")"
           if [ -z "$log_group" ] || [ "$log_group" = "None" ]; then
             echo "BackendLambdaLogGroupName output not found in BackendLambdaStack; skipping" >&2
             exit 0


leonarduk:
Responding to the DeepSeek review's blocking claim: **"all other `describe-stacks --query` calls remain unguarded."** This is factually incorrect — every call site was audited, not just the one that got patched.

Full inventory of all 25 `describe-stacks`/`list-stack-resources --query ...OutputValue`-style calls in the current file, all using `--output text`, each with its guard line(s):

| Output | Lines | `--output text` | Guard |
|---|---|---|---|
| `data_bucket` (list-stack-resources) | 216-219 | yes | L220 |
| `backend_stack_arn` | 225-226 | yes | L229 |
| `static_site_stack_arn` | 227-228 | yes | L232 |
| `price_refresh_alias_arn` | 236-239 | yes | L240 |
| `backend_log_group` | 244-247 | yes | L249 |
| `UI_AUTH_USER_POOL_ID` | 397-400 | yes | L448 |
| `UI_AUTH_USER_POOL_CLIENT_ID` | 401-404 | yes | L452 |
| `DISTRIBUTION_DOMAIN` | 412-415 | yes | L456 |
| `SMOKE_TEST_USER_POOL_CLIENT_ID` | 430-433 | yes | L434 |
| `UI_AUTH_DOMAIN` | 441-444 | yes | L445 |
| `refresh_fn` (list-stack-resources) | 499-502 | yes | L503 |
| `data_bucket` (Warm price snapshot) | 507-510 | yes | L511 |
| `BACKEND_URL` (Get backend API URL) | 560-563 | yes | L564 |
| `FRONTEND_DOMAIN` (Lighthouse) | 699-702 | yes | L703 |
| `USER_POOL_ID` (Reconcile) | 732-735 | yes | L740 |
| `CLIENT_ID` (Reconcile) | 736-739 | yes | L741 |
| `backend_url` (Validate endpoints) | 921-924 | yes | L929 |
| `frontend_domain` (Validate endpoints) | 925-928 | yes | L933 |
| `log_group` (deploy-job log fetch) | 1009-1012 | yes | L1013 |
| `backend_fn` (deploy-job metrics) | 1034-1037 | yes | L1038 |
| `BACKEND_URL` (smoke-test resolve) | 1149-1152 | yes | L1153 |
| `client_id` (smoke-test token) | 1177-1180 | yes | L1181 |
| `backend_fn` (smoke-test invocation check) | 1258-1261 | yes | L1262 |
| `log_group` (smoke-test log fetch) | 1309-1312 | **patched this PR** | L1313 |

This inventory was produced by re-grepping the file after the review, not from memory — see the raw grep output used to build it:

```
216:          data_bucket="$(aws cloudformation list-stack-resources \
219:            --output text 2>/dev/null || echo "")"
220:          if [ -z "$data_bucket" ] || [ "$data_bucket" = "None" ]; then
...
1309:          log_group="$(aws cloudformation describe-stacks \
1312:            --output text 2>/dev/null || echo "")"
1313:          if [ -z "$log_group" ] || [ "$log_group" = "None" ]; then
```

The only real gap found was the one this PR fixes: the smoke-test log-fetch step (L1309) was missing the `2>/dev/null || echo ""` fallback its sibling deploy-job log step already had, so a transient API error there would hard-fail the step instead of degrading gracefully. All other 24 sites already followed the guarded pattern before this PR — there was nothing else to patch without changing correct, already-working behavior (which the issue's own constraints explicitly rule out).

Happy to add anything concrete if a specific line is believed to be missing a guard — but "all other calls remain unguarded" doesn't match the file as written.

PR #4874, opened 2026-07-07
test(frontend): consolidate route registry test coverage

### LINKED ISSUE #4413: Eliminate duplicated frontend route-mode logic and create a single route registry
## What

Create a single shared route registry that defines path segments, route metadata, mode mapping, and menu/plugin integration in one place. Update both `frontend/src/App.tsx` and `frontend/src/main.tsx` to consume the same helper instead of each deriving route-to-mode mappings independently.

## Why

Frontend route-to-mode derivation currently exists in both `App.tsx` and `main.tsx`, and the mappings are already not identical. This creates a real risk that bootstrapping state, route markers, menu highlighting, or analytics markers diverge for the same URL — and contributors must remember to update multiple files whenever a route is added or renamed.

## How

- Create a shared route registry module (path segments, metadata, mode mapping, menu/plugin integration).
- Update `App.tsx` and `main.tsx` to both import and use this single helper.
- Add tests covering every known route segment, asserting identical output from bootstrap and runtime code paths.

## Constraints

- No user-visible navigation behaviour change — this is a consolidation, not a redesign.
- Where the two existing mappings already disagree, surface that explicitly (e.g. in the PR description) rather than silently picking one side, since it may indicate an existing bug worth its own follow-up.

## LLM tier

**Sonnet** — requires diffing two existing divergent implementations, designing a shared abstraction, and updating both call sites without changing observable behaviour for the (presumably) correct cases.

## Success looks like

- One shared route registry is the only source of truth for mode derivation.
- Both `App.tsx` and `main.tsx` consume the same implementation.
- Tests cover every known route segment and assert identical output across both code paths.
- Adding a new route requires editing one file.

## Failure looks like

- Active-state or loading-state behaviour regresses for any existing route.
- The consolidation silently "fixes" a divergence by picking the wrong side, changing behaviour without anyone noticing.

_From CODEBASE_REVIEW.md, item 4 ("v1 blockers")._

## Dependencies

| Relationship | Issue |
|---|---|
| Blocks | #4415 Add failure-mode integration coverage for auth/config/bootstrap paths |

leonarduk:
## Summary

Closes #4413.

Investigation found the core ask of #4413 was already implemented: `frontend/src/routes/registry.ts` (re-exported via `pageManifest.ts`) is the single shared route registry, and both `App.tsx` and `main.tsx` already consume the same `deriveModeFromPathname` / `deriveBootstrapMode` helpers from it instead of deriving route-to-mode mappings independently. There is no remaining divergent mapping to unify.

What this PR actually does:
- Merges the useful assertions from `tests/unit/pageManifest.test.tsx` into `tests/unit/pageManifest.test.ts` and deletes the former. The `.tsx` file asserted a stale `entry.menu.section`/`entry.menu.category` shape that no longer exists on registry entries (which use a flat `menuCategory` field) — its core assertion loop (`if (!entry.menu) continue;`) silently skipped every entry, so it provided no real coverage. Having two same-purpose test files with diverging (and one stale) assumptions about the registry shape is exactly the kind of drift risk #4413 is about.
- Adds an explicit per-route-segment check that `deriveModeFromPathname` (used at runtime by `App.tsx`) and `deriveBootstrapMode` (used at bootstrap by `main.tsx`) return the same mode for every registered route, per the issue's stated success criteria.

No production code changed — no user-visible navigation behaviour change.

## Test plan
- [x] `npx vitest run tests/unit/pageManifest.test.ts tests/unit/main.familyMvpRoutes.test.tsx tests/unit/main.test.tsx tests/unit/App.test.tsx tests/unit/App.familyMvpRedirect.test.tsx tests/unit/hooks/useRouteMode.test.tsx` — all pass
- [x] Full frontend suite: `npx vitest run` — 86 files / 560 tests pass


leonarduk:
Thanks for the review. Responding point by point:

**1. "No shared registry module created"** — Correct that this diff doesn't create one, but that's because it already exists: `frontend/src/routes/registry.ts` (re-exported via `pageManifest.ts`) was added in commit `8f7b4f6b` ("Unify frontend route registry"), which merged to `main` well before this PR and before #4413 was filed. Both `App.tsx` and `main.tsx` already import `deriveModeFromPathname`/`deriveBootstrapMode` from that single module — there's no second, divergent mapping left to consolidate. I called this out explicitly in the PR description. Since the review tool only sees the diff, it couldn't see that prior history.

**2. "Imports may not exist at runtime"** — `buildPathForMode`, `deriveBootstrapMode`, and `validatePageManifest` are all defined and exported from `frontend/src/routes/registry.ts` (lines ~369-429) today. The full suite passes: `npx vitest run` → 86 files / 560 tests green, plus `npm run lint` clean. Not a runtime risk.

**3. `deriveBootstrapMode('/support', 'loading')` → `'loading'`** — this isn't masking anything; it's asserting the function's actual documented contract (`state === 'loading'` short-circuits regardless of path, matching main.tsx's loading-screen render before route data is even relevant). The per-segment loop right above it already asserts `'auth'`/`'config-error'` produce the real derived mode for every route.

**4. Missing edge cases** — fair point, and I've since pushed `84b3a89e` which folds `/totally-unknown` → `'movers'` (both via `deriveModeFromPathname` and `deriveRouteFromPathname`) into the main per-segment test, and also removes a redundant duplicate test I'd introduced in the first commit (a valid catch — ironic for a PR about de-duplication). `deriveRouteFromPathname` is exercised in the same loop as `deriveModeFromPathname`/`deriveBootstrapMode`, so it has equivalent per-segment coverage now.

Happy to file a follow-up issue for additional edge cases (empty pathname, trailing slashes, root with query params) if useful, but wanted to close the loop on the substantive points first.

PR #4875, opened 2026-07-07
test(frontend): cover bootstrap retry backoff and no-local-login paths

### LINKED ISSUE #4415: Add failure-mode integration coverage for auth/config/bootstrap paths
## What

Add integration tests covering the frontend bootstrap path (config fetch, retry scheduling, auth state restoration, route markers, login behaviour) in both auth-enabled and auth-disabled modes, plus backend startup config permutations that drive frontend behaviour: config fetch failure/retry, stale token restoration, and the backend config values that change frontend bootstrap decisions.

## Why

The frontend bootstrap path coordinates config fetch, retry scheduling, auth state restoration, route markers, and login behaviour, while backend startup controls auth mode, config values, and rate limiting. These cross-cutting flows are more complex than a typical page component but aren't currently protected by a focused integration matrix — meaning authentication and first-load regressions can block the entire product even when individual page tests still pass.

## How

- Add integration tests for first-load behaviour in auth-enabled and auth-disabled modes.
- Add tests for config fetch failure and the retry scheduling/cancellation logic (assert behaviour, not just smoke-render).
- Add tests for stale token restoration on bootstrap.
- Validate backend config route behaviour against the frontend assumptions it drives.
- Add at least one end-to-end happy-path test from bootstrap through to a portfolio screen in CI.

## Constraints

- This naturally follows #4413 (route registry consolidation) — bootstrap-path tests will be more stable once route-mode derivation is unified, so sequence after if practical.
- Focus on the bootstrap/auth/config cross-cutting paths specifically; this is not a general call for more page-level tests.

## LLM tier

**Sonnet** — requires understanding both backend startup config and frontend bootstrap sequencing, and designing integration tests that exercise timing-sensitive retry/restoration logic without becoming flaky.

## Success looks like

- Automated tests cover first-load behaviour for auth-enabled, auth-disabled, and config-failure states.
- Retry scheduling and cancellation are explicitly asserted.
- Backend config route behaviour is validated against frontend assumptions.
- CI includes at least one end-to-end happy path from bootstrap to a portfolio screen.

## Failure looks like

- Tests are flaky due to real timing dependencies in retry logic, eroding trust in CI.
- Tests smoke-render the bootstrap path without actually asserting the behaviours that matter (retry counts, restored auth state, etc.).

_From CODEBASE_REVIEW.md, item 6 ("v1 blockers")._

## Dependencies

| Relationship | Issue |
|---|---|
| Blocked by | #4413 Eliminate duplicated frontend route-mode logic |

leonarduk:
## Summary
- Closes #4415 — audited existing bootstrap-path test coverage and found it already substantial (auth-enabled/disabled first load, single retry scheduling + cancellation, stale-token 401 restoration in `frontend/tests/unit/rootBootstrap.test.tsx`; loading/retry states in `rootConfigStates.test.tsx`; recovery in `configRetrySuccess.test.tsx`; backend `awsUiAuth`/`disable_auth` contract behaviour in `tests/test_config.py` and `tests/backend/routes/test_config.py`; an existing bootstrap→portfolio E2E happy path in `frontend/tests/smoke.spec.ts`).
- Added the two gaps that weren't covered: the exponential retry-delay sequence (2000ms -> 4000ms -> 8000ms, per `main.tsx`'s `retryDelay = Math.min(30000, 2000 * 2**attempt)`) across three consecutive config-fetch failures before recovery, and `disable_auth=true` bootstrap with no `local_login_email` configured (app renders without fabricating a local identity in storage).

## Test plan
- [x] `npx vitest run tests/unit/rootBootstrap.test.tsx tests/unit/rootConfigStates.test.tsx tests/unit/configRetrySuccess.test.tsx` - 14/14 passed
- [x] `npx vitest run` (full frontend suite) - 89 files / 569 tests passed
- [x] `npx eslint tests/unit/rootBootstrap.test.tsx` - clean

Generated with Claude Code

### DIFF
--- a/frontend/tests/unit/rootBootstrap.test.tsx
+++ b/frontend/tests/unit/rootBootstrap.test.tsx
@@ -22,7 +22,7 @@ const mountRoot = async () => {
 afterEach(() => {
   vi.useRealTimers()
   vi.resetModules()
-  vi.clearAllMocks()
+  vi.restoreAllMocks()
   localStorage.clear()
 })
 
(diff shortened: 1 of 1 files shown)


leonarduk:
Thanks for the review. These four points are all diff-scope artifacts — the review only sees the lines this PR added, not the pre-existing suite they sit alongside. Each item DeepSeek flagged as missing already exists on `main`:

1. **"No auth-enabled bootstrap test"** — `frontend/tests/unit/rootBootstrap.test.tsx` already has `'shows Google login on first load when auth is enabled and no token is stored'` (line 30) and `'restores stored user context when a token is already present'` (line 131), both with `google_auth_enabled: true`. Not part of this diff because they predate it.
2. **"No cancellation/cleanup test"** — same file, `'schedules a retry after config failure and cancels it on unmount'` (line 322) asserts `clearTimeoutSpy` fires on unmount. Also predates this diff. (My new backoff test intentionally fires callbacks synchronously to assert the *delay sequence*, not cancellation — cancellation is a distinct, already-covered concern.)
3. **"No backend config route tests"** — `tests/test_config.py` (`test_aws_ui_auth_*`, lines 358-425) and `tests/backend/routes/test_config.py` (`test_read_config_exposes_auth_enabled_bootstrap_fields`, `test_read_config_includes_aws_ui_auth_when_env_set`, etc., lines 59-125) validate the exact `awsUiAuth`/`disable_auth` shape `main.tsx` branches on.
4. **"No end-to-end test"** — `frontend/tests/smoke.spec.ts:377` has `test.describe('bootstrap to portfolio happy path', ...)`, wired into CI via the `smoke:codex:poc` npm script.

I verified all four by grepping this PR's branch (based on current `main`) before posting. Given that, I don't think this PR should be blocked — it closes the two gaps that genuinely weren't covered (exponential backoff sequencing, `disable_auth` with no `local_login_email`). Happy to link these existing tests directly in the PR description if that helps future reviewers avoid the same diff-only blind spot.

PR #4880, opened 2026-07-07
test(contracts): guard generated API contract schema against drift

### LINKED ISSUE #4414: Add contract tests for the highest-value API responses consumed by the SPA
## What

Define schema-level API contracts for the core SPA endpoints — config, owners, groups, portfolio, and transactions — and validate them in backend integration tests plus frontend mock-contract tests, preferring snapshot/schema assertions over ad hoc field checks.

## Why

The application has broad unit coverage, but limited protection against backend response-shape drift for the endpoints that power owner lists, portfolios, groups, transactions, and configuration bootstrap. The frontend performs substantial runtime normalisation, which is itself a sign the contract isn't tightly pinned down. Backend changes can currently break the SPA silently, with regressions surfacing only in manual integrated testing.

## How

- Define versioned response schemas or central typed contracts for the core SPA endpoints.
- Backend tests validate full response shapes for config, owners, groups, portfolio, and transactions endpoints.
- Frontend tests consume the same contracts or generated fixtures (rather than hand-maintained mocks that can drift from the real shape).
- CI fails when a contract-breaking field change is introduced without an intentional update to the contract.

## Constraints

- This depends on or pairs well with #4412 (typed domain models) — the typed models are a natural source for generating these schemas, so sequence or coordinate with that issue rather than duplicating the modelling work.
- Don't redesign existing endpoint shapes as part of this issue; the goal is to pin down and protect what already exists.

## LLM tier

**Sonnet** — requires understanding both backend response construction and frontend consumption/normalisation code, then designing a shared contract mechanism that both sides can validate against.

## Success looks like

- Core SPA endpoints have versioned response schemas or central typed contracts.
- Backend tests validate full response shapes for the five listed endpoint groups.
- Frontend tests consume the same contracts or generated fixtures.
- CI fails on contract-breaking changes made without updating the contract.

## Failure looks like

- Contracts are defined but not actually wired into CI, so they don't catch real regressions.
- Frontend mocks diverge from the contract over time because nothing forces them to stay in sync.

_From CODEBASE_REVIEW.md, item 5 ("v1 blockers")._

## Dependencies

| Relationship | Issue |
|---|---|
| Blocked by | #4412 Introduce typed domain models |

leonarduk:
## Summary

Issue #4414 asked for versioned response schemas / typed contracts for the SPA's core endpoints (config, owners, groups, portfolio, transactions), with backend + frontend tests protecting them and CI failing on undocumented drift.

Investigation found this is already substantially implemented on `main` via two existing mechanisms:
- `backend/contracts_spa.py` + `frontend/src/contracts/spa.ts` + `tests/backend/test_spa_response_contracts.py`
- `frontend/src/contracts/apiContracts.ts` (Zod, generates JSON Schema) + `tests/contracts/test_api_response_contracts.py`, which hits the live FastAPI app and validates real responses against the schema bundle in `frontend/src/contracts/generated/api-contract-schemas.v1.json`

The one concrete gap: nothing verified that the **committed generated schema bundle** actually matches what `frontend/scripts/generate-api-contract-schemas.mjs` would currently produce from `apiContracts.ts` — it could silently drift from its own source of truth.

- Added `frontend/tests/unit/apiContractSchemaGeneration.test.ts`, which rebuilds the schema bundle in-process from `apiContractJsonSchemas`/`API_CONTRACT_VERSION` and asserts it deep-equals the committed JSON file.
- Running the generator surfaced real pre-existing drift: `days_held` had lost its nullable variant, and `groupPortfolio`'s optional `members_summary`/`subtotals_by_account_type` fields were incorrectly marked required. Regenerated the committed file to fix both.

## Out of scope (flagged separately)

`tests/contracts/test_api_response_contracts.py::test_live_endpoints_match_versioned_contract_schema[groupPortfolio-/portfolio-group/all]` already fails on `main` (verified via `git stash` before any of my changes) — the live `/portfolio-group/all` response includes holding fields (`cost_basis_source`, `eligible_on`, `unrealised_gain_gbp`, `unrealized_gain_gbp`, `asset_class`) not declared in the contract schema. This is a genuine, separate drift issue that requires deciding the correct field types from the real route, which is out of scope for this drift-guard change (issue #4414 explicitly says not to redesign existing endpoint shapes). Spun off as a follow-up task.

## Test plan

- [x] `cd frontend && npx vitest run apiContractSchemaGeneration apiContracts.test` — both pass
- [x] `cd frontend && npx eslint --format unix tests/unit/apiContractSchemaGeneration.test.ts` — clean
- [x] `cd frontend && npx tsc -b --noEmit` — clean
- [x] `python -m pytest tests/contracts/test_api_response_contracts.py tests/backend/test_spa_response_contracts.py -q --no-cov -k "not groupPortfolio"` — 15 passed
- [x] Confirmed the pre-existing `groupPortfolio` failure exists on `main` independent of this change

Closes #4414

### DIFF
--- a/frontend/src/contracts/generated/api-contract-schemas.v1.json
+++ b/frontend/src/contracts/generated/api-contract-schemas.v1.json
@@ -473,7 +473,14 @@
                         ]
                       },
                       "days_held": {
-                        "type": "number"
+                        "anyOf": [
+                          {
+                            "type": "number"
+                          },
+                          {
+                            "type": "null"
+                          }
+                        ]
                       },
                       "sell_eligible": {
                         "type": "boolean"
@@ -570,9 +577,7 @@
           "as_of",
           "members",
           "total_value_estimate_gbp",
-          "accounts",
-          "members_summary",
-          "subtotals_by_account_type"
+          "accounts"
         ],
         "additionalProperties": false
       }
@@ -854,7 +859,14 @@
                         ]
                       },
                       "days_held": {
-                        "type": "number"
+                        "anyOf": [
+                          {
+                            "type": "number"
+                          },
+                          {
+                            "type": "null"
+                          }
+                        ]
                       },
                       "sell_eligible": {
                         "type": "boolean"
--- a/frontend/tests/unit/apiContractSchemaGeneration.test.ts
+++ b/frontend/tests/unit/apiContractSchemaGeneration.test.ts
@@ -1,0 +1,36 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+import { API_CONTRACT_VERSION, apiContractJsonSchemas } from "@/contracts/apiContracts";
+
+// Mirrors the payload shape written by frontend/scripts/generate-api-contract-schemas.mjs.
+// Keep this endpoint list in sync with that script.
+function buildExpectedBundle() {
+  return {
+    version: API_CONTRACT_VERSION,
+    endpoints: {
+      config: { path: "/config", schema: apiContractJsonSchemas.config },
+      owners: { path: "/owners", schema: apiContractJsonSchemas.owners },
+      groups: { path: "/groups", schema: apiContractJsonSchemas.groups },
+      groupPortfolio: { path: "/portfolio-group/all", schema: apiContractJsonSchemas.groupPortfolio },
+      portfolio: { path: "/portfolio/alice", schema: apiContractJsonSchemas.portfolio },
+      transactions: { path: "/transactions", schema: apiContractJsonSchemas.transactions },
+    },
+  };
+}
+
+const generatedSchemaPath = path.resolve(
+  process.cwd(),
+  "src/contracts/generated/api-contract-schemas.v1.json",
+);
+
+describe("generated API contract schema bundle", () => {
+  it("matches what generate-api-contract-schemas.mjs would produce from apiContracts.ts", () => {
+    const committed = JSON.parse(readFileSync(generatedSchemaPath, "utf8"));
+    const expected = buildExpectedBundle();
+
+    // Structural comparison (not byte-for-byte) so trailing-newline/formatting
+    // differences don't mask a real schema drift, and vice versa.
+    expect(committed).toEqual(expected);
+  });
+});


leonarduk:
Responding to the DeepSeek review — the core claims don't hold up against what's actually in the repo; the review appears to have evaluated this diff in isolation without checking pre-existing files.

**"Backend tests validate full response shapes: not addressed"** — These already exist on `main`, added before this PR:
- `tests/contracts/test_api_response_contracts.py` — hits the *live* FastAPI app (`backend.app:create_app()`) and validates real `/config`, `/owners`, `/groups`, `/portfolio/{owner}`, `/portfolio-group/all`, `/transactions` responses against the JSON Schema bundle this PR guards.
- `tests/backend/test_spa_response_contracts.py` — a second, independent contract mechanism (`backend/contracts_spa.py`) validating the same five endpoint groups via `model_validate`/`TypeAdapter`.

This PR's job was narrower and stated in its own description: those two mechanisms exist, but nothing verified the *committed* generated schema file actually matched its own Zod source — so it could (and had) drifted silently. That's the one gap this PR closes.

**"CI fails on contract-breaking changes: not addressed"** — `pytest tests --ignore=tests/live ...` in `backend-integration.yml` already runs `tests/contracts/` and `tests/backend/` (unqualified `tests` glob), and `npm test`/`pnpm test` in `frontend-tests.yml` already runs all vitest files including the new one added here. No new workflow file was needed because both suites are already invoked broadly, not via an allowlist.

**"Required-fields removal is an accidental regression"** — This is a misread of the diff: the changed `required` array is in the **`groupPortfolio`** schema block (identifiable by the `members` field), not `portfolio` — the plain `portfolio` schema has no `members_summary` field at all and is unchanged. And the removal isn't accidental: `members_summary`/`subtotals_by_account_type` are declared `.optional()` in `frontend/src/contracts/apiContracts.ts` (lines 121, 130) — the actual Zod source of truth, which this PR does not touch. The *previously committed* JSON had them incorrectly marked required, i.e. it had drifted from its own source. That's exactly the class of bug this PR's new test (`apiContractSchemaGeneration.test.ts`) is designed to catch — regenerating from source and diffing against the committed file. Re-adding them to `required` would reintroduce the drift and fail the new guard test; if the intent is for those fields to actually be mandatory, that's a schema-design change in `apiContracts.ts` itself, out of scope here (issue #4414 explicitly says not to redesign existing endpoint shapes).

Separately: I found and filed as a follow-up (not in this PR) a real, pre-existing failure in `test_api_response_contracts.py` for the `groupPortfolio` endpoint, where the live route returns holding fields (`cost_basis_source`, `eligible_on`, `unrealised_gain_gbp`, etc.) the schema doesn't yet declare — confirmed via `git stash` that this fails on `main` independent of this change.

PR #4926, opened 2026-07-08
fix(backend): use typed load_account_record in get_account route

### LINKED ISSUE #4878: Route remaining load_account/load_person_meta call sites through typed AccountRecord/PersonMetadata accessors
> **Correction (2026-07-08):** the original AI-generated body below was inaccurate — it referenced a nonexistent `src/allotmint/...` layout and claimed `PersonMetadata`/`AccountRecord` validation wasn't wired into production code. In this repo, `load_account()` and `load_person_meta()` in `backend/common/data_loader.py` already validate via `AccountRecord.model_validate()` / `PersonMetadata.model_validate()`, and typed wrappers `load_account_record()` / `load_person_metadata()` are already used in several routes (`backend/common/portfolio.py:157`, `backend/routes/pension.py:40`, `backend/routes/portfolio.py:409,446`, `backend/auth.py:105,133`). The real, narrower gap is below.

## What

A handful of call sites still use the untyped, dict-returning `load_account()` / `load_person_meta()` directly and do `.get("field")` chains instead of attribute access, while other call sites already use the typed `load_account_record()` / `load_person_metadata()` wrappers:

- `backend/routes/portfolio.py:866` and `:910` — `data_loader.load_account(...)` in `get_account` and its sibling lookup.
- `backend/quests/trail.py:107` — `data_loader.load_person_meta(...)`.
- `backend/common/authz.py:70` — `load_person_meta(...)`.
- `backend/common/portfolio_loader.py:43,71` — `load_account(...)` / `load_person_meta(...)` inside `_load_accounts_for_owner` / `_build_owner_portfolio`.

## Why

- These call sites already benefit from validation-at-the-boundary (via the internal `model_validate` calls inside `load_account`/`load_person_meta`), so this is not a correctness bug — it's an inconsistency.
- The mixed pattern (typed access in some routes, raw `.get()` chains in others) invites bugs and confuses future contributors (human or AI) about which accessor to reach for.
- Consolidating on the typed accessors removes the last defensive `.get()` chains for these fields.

## How

1. In each listed call site, replace `load_account(...)` with `load_account_record(...)` and `load_person_meta(...)` with `load_person_metadata(...)`.
2. Update the immediate consumers of the returned value to use attribute access (`record.holdings` instead of `record["holdings"]` / `record.get("holdings")`).
3. Where the raw dict shape is still needed downstream (e.g. an API response body), call `.model_dump(exclude_unset=True)` explicitly at the boundary rather than passing dicts around internally (`exclude_unset` preserves exactly the key set present in the source JSON; `exclude_none` would drop legitimate explicit nulls and fabricate empty-list keys for fields like `holdings`/`approvals` that default to `[]` rather than `None`).
4. Do not change `load_account()`/`load_person_meta()` themselves — keep them as the raw-dict variants for any remaining callers that need exact passthrough serialization.

## Constraints

- No change to serialized JSON response shape for any route.
- No change to the `list_plots()`/`OwnerSummaryRecord` path from #4877 — this is scoped to the account/person accessors only.
- Out of scope: adding fields, changing validation rules, refactoring `data_loader.py` structure.

## LLM tier

**Sonnet** — mechanical but touches several files; needs care around exact-shape preservation at API boundaries.

## Success looks like

- The listed call sites use `load_account_record`/`load_person_metadata` and attribute access.
- All existing tests pass with no response-shape changes.

## Failure looks like

- A route's JSON response shape changes because a `.model_dump()` call drops or reorders fields the raw dict used to include.

---

<details>
<summary>Original (inaccurate) AI-generated body</summary>

### What

Currently, `PersonMetadata` and `AccountRecord` Pydantic models are defined and tested in `test_account_models.py`, but they are **not used in any production data loading path**. Only `OwnerSummaryRecord` was threaded through the codebase in PR #4877.

The following files and functions need changes:

- **`src/allotmint/data_loader.py`** — In the `_load_accounts` and `_load_person_metadata` functions (or equivalent data-loading entry points), add validation using `PersonMetadata.model_validate()` and `AccountRecord.model_validate()` before returning data.
- **`src/allotmint/models/account_models.py`** — Ensure `PersonMetadata` and `AccountRecord` are exported and importable from the same location as `OwnerSummaryRecord`.
- **Any route or service function** that currently returns raw dicts for person metadata or account documents (e.g., `get_person_metadata`, `get_accounts` in `portfolio.py` or similar) should return validated model instances.

</details>

leonarduk:
## Summary
- `get_account` (`/account/{owner}/{account}`) previously called the raw dict-returning `data_loader.load_account()`, even though that function already validates internally against `AccountRecord`.
- Switched both call sites in the route to `data_loader.load_account_record()` + `.model_dump(exclude_unset=True)`, matching the typed-access pattern already used by `backend/common/portfolio.py`, `backend/routes/pension.py`, and `backend/auth.py`.
- No change to error handling (validation already happened inside `load_account`) or response shape (`exclude_unset=True` preserves the original key set from the source JSON).

## Scope note
This closes #4878, but only implements one of the four call sites originally listed there. During implementation I found the other three (`backend/quests/trail.py`, `backend/common/authz.py`, `backend/common/portfolio_loader.py`) intentionally rely on the *lenient* loaders (`load_person_meta`/`load_account`, which degrade to `{}`/skip on bad data) rather than the *strict* typed accessors, and converting them would introduce real regressions (crashing owner lookups, turning clean 403s into unhandled 500s, or cascading a `list[dict]` → `list[AccountRecord]` change across half a dozen unrelated consumers). Full reasoning is in [the issue comment](https://github.com/leonarduk/allotmint/issues/4878#issuecomment-4919369393).

## Test plan
- [x] `pytest tests/test_accounts_api.py tests/backend/test_portfolio_routes.py -q --no-cov` — 13 passed, 1 xfailed
- [x] `pytest tests/ -k "account or portfolio" -q --no-cov` — 442 passed, 1 xfailed, 1 xpassed
- [x] Confirmed pre-existing `ruff`/`black` import-order warning on this file predates this change (verified via `git stash`)

Closes #4878

### DIFF
--- a/backend/routes/portfolio.py
+++ b/backend/routes/portfolio.py
@@ -863,7 +863,7 @@ async def get_account(owner: str, account: str, request: Request):
     root = resolve_accounts_root(request)
 
     try:
-        data = data_loader.load_account(owner, account, root)
+        data = data_loader.load_account_record(owner, account, root).model_dump(exclude_unset=True)
     except data_loader.ProviderUnavailable as exc:
         log.warning(
             "portfolio.account_provider_unavailable",
@@ -907,7 +907,9 @@ async def get_account(owner: str, account: str, request: Request):
         if not match:
             raise HTTPException(status_code=404, detail="Account not found")
         try:
-            data = data_loader.load_account(owner_dir.name, match, owner_dir.parent)
+            data = data_loader.load_account_record(owner_dir.name, match, owner_dir.parent).model_dump(
+                exclude_unset=True
+            )
         except data_loader.ProviderUnavailable as exc:
             raise HTTPException(status_code=503, detail="Account data provider unavailable") from exc
         except data_loader.InvalidPayload as exc:


leonarduk:
Responding to the DeepSeek review:

**On \`exclude_unset\` vs \`exclude_none\`:** the review is backwards here — I verified concretely against \`AccountRecord\`:

\`\`\`python
# Field explicitly present as null in source JSON
data1 = {"account_type": None, "currency": "GBP", "holdings": []}
r1 = AccountRecord.model_validate(data1)
r1.model_dump(exclude_unset=True)  # {'account_type': None, 'currency': 'GBP', 'holdings': []}  <- matches source
r1.model_dump(exclude_none=True)   # {'currency': 'GBP', 'holdings': [], 'approvals': []}        <- drops account_type, FABRICATES approvals key

# Field absent entirely from source JSON
data2 = {"account_type": "ISA", "currency": "GBP"}
r2 = AccountRecord.model_validate(data2)
r2.model_dump(exclude_unset=True)  # {'account_type': 'ISA', 'currency': 'GBP'}                              <- matches source
r2.model_dump(exclude_none=True)   # {'account_type': 'ISA', 'currency': 'GBP', 'holdings': [], 'approvals': []}  <- fabricates both keys
\`\`\`

\`exclude_none=True\` would inject \`holdings\`/\`approvals\` keys (they default to \`[]\`, not \`None\`, so \`exclude_none\` doesn't touch them) into responses that never had them, and would drop legitimately-present \`null\` fields — both are shape changes, which is exactly what the "no serialized shape change" constraint is trying to prevent. \`exclude_unset=True\` is the one that actually preserves the exact key set from the source file, since Pydantic tracks which fields were explicitly present at parse time.

The issue's own "How" section said \`exclude_none=True\` — that was a mistake in my own issue draft, not in this code. I've corrected the issue text to say \`exclude_unset=True\` so it doesn't mislead future readers: https://github.com/leonarduk/allotmint/issues/4878

**On the other three call sites:** this is intentional, not a miss — explained in https://github.com/leonarduk/allotmint/issues/4878#issuecomment-4919369393 and in this PR's description. Short version: \`backend/quests/trail.py\`, \`backend/common/authz.py\`, and \`backend/common/portfolio_loader.py\` all rely on the *lenient* loaders on purpose (graceful degradation on bad data for batch/security-sensitive paths), and switching them to the strict typed accessors would turn a skipped bad record into a crashed owner-lookup or an unhandled 500 instead of a clean 403. That's a behavior regression, not a cleanup, so I left them out of this PR's scope.

**On test coverage:** \`tests/test_accounts_api.py::test_account_route_returns_data\` and \`test_account_route_adds_missing_account_type\`, plus \`tests/backend/test_portfolio_routes.py\`, all exercise \`get_account\`'s response shape directly and pass unchanged (442 passed in the broader \`account or portfolio\` test run).

PR #4945, opened 2026-07-08
Automate dividend collection via scheduled Lambda

### LINKED ISSUE #2750: Automate collection of dividend info
Automate collection of dividend information rather than manual entry. Could use a financial data API or scheduled Lambda to pull dividend announcements and record them automatically.

leonarduk:
## Summary
- Adds `backend/common/dividends.py::refresh_dividends()`, which fetches declared dividends for every held ticker via yfinance and records them as `DIVIDEND` transactions, deduplicated on `(ticker, ex_date)` so re-running never creates duplicates.
- Adds `backend/lambda_api/dividend_refresh.py` (Lambda entry point) and registers `DividendRefreshLambda` in `cdk/stacks/backend_lambda_stack.py`, scheduled daily at 06:00 UTC via EventBridge, following the existing `PriceRefreshLambda` pattern.
- IAM is scoped to `writable-accounts/` only (read+put+list) — no `accounts/` list grant needed since this path never calls `list_portfolios()`, unlike `PriceRefreshLambda`.
- Unrecognised tickers (e.g. OEIC funds) are skipped, not errored. On a ticker's first run (no prior recorded dividend), only declarations from the last 90 days are imported, to avoid a surprise historical backfill.

## Design notes (documented in the module docstring)
- Writes the `DIVIDEND` (singular) transaction type literal, matching `backend/reports.py` income aggregation and the issue's acceptance criteria. Note the cash-flow sign tables in `portfolio_loader.py`/`portfolio_utils.py` (used by TWR/XIRR) only recognise `DIVIDENDS` (plural) — this pre-existing inconsistency is out of scope for this change and not touched here.
- The per-share dividend amount is converted to GBP and multiplied by the *currently* held unit count (not the historical count as of the ex-date), since reconstructing historical share counts is out of scope for v1.

## Test plan
- [x] `pytest tests/test_dividends.py tests/test_dividend_refresh_lambda.py` — 11 new tests: creation, idempotency on re-run, unrecognised-ticker skip, empty-series skip, lookback-window on first run, pence-currency conversion, zero-unit-holding skip, multi-owner/account handling, and a guard that the real yfinance client is never constructed.
- [x] `pytest cdk/tests` — 122 passed, including a new `test_dividend_refresh_lambda_has_scoped_s3_permissions` locking in least-privilege IAM.
- [x] `black`/`isort`/`ruff` clean on all new/changed backend files.

Closes #2750

### DIFF
--- a/backend/common/dividends.py
+++ b/backend/common/dividends.py
@@ -1,0 +1,241 @@
+"""Automated dividend collection.
+
+Fetches declared dividends for held tickers via yfinance and records them as
+``DIVIDEND`` transactions, so dividend income no longer has to be entered by
+hand. See issue #2750.
+
+Design notes / known v1 limitations
+------------------------------------
+* Transaction type literal: the codebase is inconsistent between ``DIVIDEND``
+  (``backend/reports.py`` income aggregation, ``backend/utils/positions.py``)
+  and ``DIVIDENDS`` (cash-flow sign tables in ``portfolio_loader.py`` and
+  ``portfolio_utils.py`` used by TWR/XIRR). This module writes ``DIVIDEND`` to
+  match the income-reporting path and the issue's stated acceptance criteria.
+  Fixing the cash-flow sign tables to also recognise ``DIVIDEND`` is tracked
+  separately and out of scope here.
+* Amount calculation: yfinance's ``Ticker.dividends`` returns a per-share
+  amount, not the total cash received. This module approximates the total by
+  multiplying the per-share amount by the *currently* held unit count for
+  that position, rather than reconstructing historical share counts as of the
+  ex-date. If units held changed between the ex-date and when this job runs,
+  the recorded amount is an approximation.
+* Idempotency: the transaction store has no uniqueness constraint, so this
+  module de-duplicates by scanning existing ``DIVIDEND`` transactions for a
+  matching ``(ticker, ex_date)`` pair before writing.
+* First run per ticker: when no prior ``DIVIDEND`` transaction exists for a
+  ticker, only dividends declared within the last ``_DEFAULT_LOOKBACK_DAYS``
+  days are imported (not full history), to avoid surprise historical backfill
+  on the first scheduled run for an existing holding.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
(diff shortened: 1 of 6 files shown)


leonarduk:
Thanks for the review. Checked each blocking item against the code:

**Fixed (real bug):** empty dividend series was treated identically to a provider error, permanently flagging every non-dividend-paying holding (e.g. GOOGL) as "unrecognised" on each run. Fixed in dcb858c6 — only an actual exception from the provider call now counts as a skip; an empty series yields zero new dividends. Added a direct regression test (`test_refresh_dividends_empty_series_is_not_treated_as_skip`) plus unit tests for `_fetch_new_dividends` directly.

**Not an issue — verified against the code:**
- `CurrencyNormaliser.from_raw` exists at `backend/common/currency.py:21`.
- `DATA_BRANCH` is never read by any backend Python module at runtime (`grep -rn "DATA_BRANCH" backend/` returns nothing) — it's a build-time-only var consumed by `scripts/bash/sync_data.sh` before the Docker image is built. It's set on `dividend_env` for consistency with the existing `price_refresh`/`trading_agent` Lambda envs, which set it identically and also never read it in Python.

**Test fidelity, fixed:** the unrecognised-ticker mock now raises on `.dividends` *property access* (via a small class) rather than at `Ticker()` construction, matching where yfinance actually fails on a live call.

**Design choices, left as-is with reasoning:**
- Per-account `rebuild_portfolio()` inside the loop matches the existing `create_transaction` HTTP route's pattern (rebuild immediately after each write, not batched). Deferring to a single end-of-run rebuild would be strictly worse on partial failure — a mid-run crash would leave successfully-written accounts with stale holdings too, instead of just the not-yet-processed ones.
- Alpha Vantage fallback deferred to a follow-up: documented in the module docstring now. Its free tier is aggressively rate-limited per-key, which doesn't suit a job querying every held ticker on a schedule; yfinance is already this codebase's live-price provider.

**Added test coverage:** multiple tickers in one account, duplicate same-ex-date declarations from the provider (dedup), and direct unit tests for `_fetch_new_dividends`'s filtering/shaping logic.

16/16 dividend tests pass, CDK permission tests unaffected, black/isort/ruff clean.

Issue #4898, opened 2026-07-08
Frontend: build /data-quality timeseries dashboard page


leonarduk:
# Frontend: build /data-quality timeseries dashboard page

## What
Build a React page at `/data-quality` that displays timeseries data quality metrics per position with a RAG status, and allows drilling into a specific position's problematic dates.

## Why
Split from #2751 to separate the frontend page from the backend algorithm (see #4897 for the backend endpoint). Without a dedicated quality view, data problems (gaps, duplicates, outliers) are invisible until something looks obviously wrong in a chart. A proactive dashboard makes them visible and actionable.

## Clarifications (from codebase review, 2026-07-08)
This issue hadn't previously been checked against actual frontend conventions the way its sibling #4897 was — done now:
- **Routing:** add the page through `frontend/src/routes/registry.ts`'s `ROUTE_REGISTRY`, the single source of routing truth. Standalone lazy-loaded pages set `routePath` + `lazyComponent` (see the `trade-compliance` entry at `registry.ts:249-258` or `trail` at `:193-202`); these are auto-collected into `standalonePageRoutes` (`registry.ts:442-449`) and mounted as routes. Follow this pattern rather than adding a route ad hoc.
- **Table:** `frontend/src/components/HoldingsTable.tsx` is the canonical sortable-table example — uses the `useSortableTable` hook (`HoldingsTable.tsx:167`), shared `tableStyles` classes, column-visibility toggles, and i18n via `t("holdingsTable.columns.*")`. Mirror this rather than building a bespoke table.
- **RAG status badge: no existing component to reuse — this is genuinely new.** No tri-color (green/amber/red) status-pill component exists anywhere in `frontend/src`. The closest precedent is `HoldingsTable.tsx:537-552`'s stale-price indicator, which is a binary stale/fresh flag, not a 3-state RAG. Budget real design/implementation time for this rather than assuming it's a quick reuse — and this is exactly the kind of ad hoc styling the issue's own "Failure looks like" section warns about, so treat it as a first-class deliverable, not an afterthought.
- **Drill-down:** `frontend/src/components/VarBreakdownModal.tsx` is a working drill-down modal (`role="dialog"`, Escape-to-close, opened from a table row in `components/ValueAtRisk.tsx`, with a callback to bubble the selected row back up) — use this as the template for drilling into a position's problematic dates instead of inventing a new modal/expansion pattern.
[...]

PR #4946, opened 2026-07-09
Backend: add GET /data-quality/timeseries endpoint

### LINKED ISSUE #4897: Backend: add GET /data-quality/timeseries endpoint
# Backend: add GET /data-quality/timeseries endpoint

## What
Implement a `GET /data-quality/timeseries` endpoint that, for each **distinct ticker+exchange** with stored timeseries data, reports data quality metrics: total data points, date range, gap count, duplicate dates, and outliers.

## Why
Split from #2751 to separate the backend algorithm work from the frontend page — the two are independently assignable. Corrupt or missing timeseries data silently breaks calculations (a gap produces wrong returns, a duplicate date produces wrong aggregations, an outlier skews charts). A dedicated endpoint makes these problems detectable both by a UI (see the companion frontend issue #4898) and independently by automation (e.g. a scheduled report).

## Clarifications (from codebase review, 2026-07-08)
- **"Position" means distinct (ticker, exchange), not per-account holding.** Timeseries in this repo is stored per ticker+exchange (e.g. `tests/data/timeseries/meta/AZN_L.parquet` — confirmed 76 such files, ~1.2MB total, under `tests/data/timeseries/meta/`, `TICKER_EXCHANGE.parquet` naming), loaded via `backend/timeseries/cache.py::load_meta_timeseries_range` (defined at line 557). There is no per-account "position" tied 1:1 to a timeseries — accounts hold arrays of `holdings` referencing tickers, and the only existing `Position` model (`backend/routes/tax.py:20`, a Pydantic model with `ticker`/`basis`/`price` fields) is an unrelated tax-lot concept used only as the request payload for `POST /tax/harvest`. Iterate over distinct tickers/exchanges with stored data, not once per account holding — the underlying quality metrics are identical regardless of who holds the instrument, so per-holding duplication would be wasteful and confusing downstream in #4898.
- **No UK trading-calendar utility exists yet.** Grep of the codebase turns up only a docstring comment noting the cache layer "may shift the window backward by up to four days to accommodate weekends/holidays" (`backend/common/instrument_api.py:235-239`), not an actual calendar implementation. The only other "trading days" hits are `backend/common/approvals.py::add_trading_days` (a simple weekend-skip helper for approval expiry, not a market calendar) and a hardcoded `trading_days = 252` annualization constant in `backend/common/risk.py` — neither is a bank-holiday-aware calendar. This issue also requires adding one (e.g. `pandas_market_calendars`'s LSE calendar, or a hardcoded UK bank-holiday list covering the data's date range) — budget for this as real scope, not a one-line detail. **Note:** `pandas_market_calendars` is not currently a dependency anywhere in the repo (checked all `requirements*.txt` files) — adding it means a new Lambda-packaging consideration (wheel size/availability), which a hardcoded UK bank-holiday list would avoid; weigh that tradeoff explicitly rather than defaulting to the library.
- **The Lambda-timeout constraint is likely non-binding at current scale.** Real data volume is ~76 tickers and ~1.2MB of parquet in test fixtures (no evidence found of a larger production ticker universe, though this is fixture data so production scale is not independently confirmed). Computing gap/duplicate/outlier detection over that is milliseconds, not seconds. **Correction (2026-07-08 re-verification):** there is no existing async/background-job pattern used for cache-warmup specifically — `BackgroundTasks` is used elsewhere in the backend (`backend/routes/news.py`, `backend/routes/screener.py`), just not for cache-warmup. Either way, don't pre-build pagination/async infrastructure speculatively; measure the full sweep first and only add pagination if it's actually shown to approach the 30s limit.

## How
- Follow the existing route pattern in `backend/routes/timeseries_meta.py` (`APIRouter(prefix="/timeseries", ...)`, registered in `backend/bootstrap/routers.py:44`) as a structural template.
- For each distinct ticker+exchange with stored timeseries data, return:
  - Total data points, date range (first/last date)
  - Gap count: expected business days with no data point (> N days gap counts as a gap; configurable N)
  - Duplicate dates: any date appearing more than once
  - Outliers: points where value deviates > N standard deviations from the rolling mean (configurable, suggest σ=3)
- Gap detection must account for non-trading days (weekends, UK public holidays) — a weekend gap is not a data quality issue. Add a UK bank-holiday calendar utility as part of this work (none currently exists in the codebase); see the dependency tradeoff note above.
- Design the endpoint to be callable independently of any UI (e.g. by a scheduled Lambda that emails a weekly report), not just as a UI backend.

## Constraints
- Read-only — must not modify data; fixing issues requires manual action or a separate corrective tool.
- Must complete within Lambda timeout limits (30s). At current data volume (~76 tickers, ~1.2MB) this is unlikely to be a real constraint — measure the full sweep before adding pagination or async processing; don't build that infrastructure speculatively.
- Out of scope: automated data correction or gap-filling (separate feature).

## LLM tier
**Sonnet** — non-trivial gap/outlier detection logic with trading-calendar awareness.

## Success looks like
- [ ] Endpoint returns total points, date range, gap count, duplicate dates, and outliers per distinct ticker+exchange
- [ ] Gap detection correctly excludes weekends/UK public holidays
- [ ] Endpoint completes comfortably within Lambda timeout for the full ticker set (pagination/async only added if measurement shows it's actually needed)
- [ ] Callable independently of the frontend (e.g. by a script or scheduled job)
- [ ] Backend tests cover gap, duplicate, and outlier detection with fixture data

## Failure looks like
- Weekend/holiday gaps are flagged as data quality issues (false positives)
- Metrics are computed/duplicated per account holding instead of per distinct ticker+exchange
- Endpoint times out on the full ticker set with no pagination fallback
- Endpoint mutates data instead of being strictly read-only
- A new dependency (e.g. `pandas_market_calendars`) is added without weighing the Lambda-packaging tradeoff against a hardcoded holiday list

leonarduk:
## Summary
- Adds a read-only `GET /data-quality/timeseries` endpoint reporting, per cached ticker/exchange, total points, date range, gap count/detail, duplicate dates, and outliers (rolling-mean sigma threshold).
- Gap detection excludes weekends and UK bank holidays via a new self-contained `backend/utils/uk_holidays.py` (no new dependency added — none existed in the repo, and current data volume didn't justify one).
- Adds `list_cached_meta_tickers()` and `load_cached_meta_timeseries_full()` to `backend/timeseries/cache.py` — both read-only, no network fetch, so the endpoint can never mutate the cache or trigger a live fetch.
- Supports optional `ticker`/`exchange` query filtering (for the drill-down use case in #4898) plus configurable `gap_threshold_days`, `outlier_sigma`, `rolling_window`.

## Test plan
- [x] `tests/utils/test_uk_holidays.py` — bank holiday dates and business-day range logic, verified against real calendar dates
- [x] `tests/timeseries/test_quality.py` — gap/duplicate/outlier detection unit tests, including weekend-only gaps not flagged
- [x] `tests/test_data_quality_cache_helpers.py` — cache enumeration (local + mocked S3) and raw-read helpers
- [x] `tests/routes/test_data_quality.py` — route integration tests (filtering, validation, 404/400 cases, XSS payload rejection)
- [x] `tests/test_app.py`, `tests/test_timeseries_cache_base.py`, existing `timeseries_meta` tests — no regressions
- [x] `ruff check` / `black --check` clean on all touched files

Closes #4897

## Summary by CodeRabbit

* **New Features**
  * Added a new public data-quality endpoint for timeseries data, with optional ticker and exchange filtering.
  * Introduced quality checks for gaps, duplicate dates and outliers on cached timeseries data.
  * Added UK business-day and bank-holiday handling to improve date-based calculations.

* **Bug Fixes**
  * Improved error handling for missing or invalid timeseries cache data.
  * Added clearer validation so incomplete or malformed ticker/exchange queries are rejected.

### DIFF
--- a/backend/bootstrap/routers.py
+++ b/backend/bootstrap/routers.py
@@ -13,6 +13,7 @@ from backend.routes.analytics import router as analytics_router
 from backend.routes.approvals import router as approvals_router
 from backend.routes.compliance import router as compliance_router
 from backend.routes.config import router as config_router
+from backend.routes.data_quality import router as data_quality_router
 from backend.routes.events import router as events_router
 from backend.routes.goals import router as goals_router
 from backend.routes.instrument import router as instrument_router
@@ -61,6 +62,7 @@ def register_routers(app: FastAPI, cfg: Config) -> None:
     app.include_router(instrument_router)
     app.include_router(instrument_admin_router, dependencies=protected)
     app.include_router(timeseries_router)
+    app.include_router(data_quality_router)
     app.include_router(timeseries_edit_router)
     app.include_router(timeseries_admin_router, dependencies=protected)
     app.include_router(transactions_router)
--- a/backend/routes/data_quality.py
+++ b/backend/routes/data_quality.py
@@ -1,0 +1,99 @@
+"""Read-only timeseries data quality endpoint.
+
+Reports, for each cached instrument's timeseries, gap/duplicate/outlier
+metrics so problems can be surfaced by a UI (see #4898) or a scheduled job,
+independently of any frontend. Never fetches new data or mutates the cache.
+"""
+
+import logging
+import re
+
+from fastapi import APIRouter, HTTPException, Query
+from fastapi.responses import JSONResponse
+
+from backend.logging_setup import sanitise_log_value
+from backend.timeseries.cache import (
+    has_cached_meta_timeseries,
+    list_cached_meta_tickers,
+    load_cached_meta_timeseries_full,
+)
+from backend.timeseries.quality import (
+    DEFAULT_GAP_THRESHOLD_DAYS,
+    DEFAULT_OUTLIER_SIGMA,
+    DEFAULT_ROLLING_WINDOW,
+    compute_quality,
+)
+
+router = APIRouter(prefix="/data-quality", tags=["data-quality"])
+logger = logging.getLogger("routes.data_quality")
+
+# Same allowlist shape as backend/routes/timeseries_meta.py — bounds
+# user-supplied ticker/exchange segments before they reach a cache file path.
+_TICKER_RE = re.compile(r"^[A-Z0-9_-]{1,50}$")
+_EXCHANGE_RE = re.compile(r"^[A-Z0-9._-]{1,50}$")
+
+
+def _validate_symbol(value: str, *, kind: str) -> str:
+    upper = value.upper()
+    pattern = _TICKER_RE if kind == "ticker" else _EXCHANGE_RE
+    if not pattern.match(upper):
+        raise HTTPException(status_code=400, detail=f"Invalid {kind} format")
+    return upper
+
+
+@router.get("/timeseries")
+async def get_timeseries_quality(
+    ticker: str | None = Query(None, description="Filter to a single ticker (must be provided with exchange)"),
+    exchange: str | None = Query(None, description="Filter to a single exchange (must be provided with ticker)"),
+    gap_threshold_days: int = Query(
+        DEFAULT_GAP_THRESHOLD_DAYS,
+        ge=0,
+        le=30,
+        description="A run of missing business days longer than this counts as a gap",
+    ),
+    outlier_sigma: float = Query(
+        DEFAULT_OUTLIER_SIGMA,
+        gt=0,
+        le=10,
+        description="Standard deviations from the rolling mean beyond which a point is an outlier",
+    ),
+    rolling_window: int = Query(
+        DEFAULT_ROLLING_WINDOW, ge=2, le=250, description="Rolling window size (trading days) for outlier detection"
+    ),
+):
+    if bool(ticker) != bool(exchange):
+        raise HTTPException(status_code=400, detail="ticker and exchange must be provided together")
+
+    if ticker and exchange:
+        tkr = _validate_symbol(ticker, kind="ticker")
+        exch = _validate_symbol(exchange, kind="exchange")
+        if not has_cached_meta_timeseries(tkr, exch):
+            raise HTTPException(status_code=404, detail="No cached timeseries found for that ticker/exchange")
+        pairs = [(tkr, exch)]
+    else:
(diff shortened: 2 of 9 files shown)


leonarduk:
Thanks for the review. Response:

**"Blocking" NameError claim — incorrect.** \`_s3_client\` (defined at cache.py:322, pre-existing) and \`BotoCoreError\`/\`ClientError\` (imported at cache.py:26, pre-existing) are both already in scope in this module — they're used by the existing \`has_cached_meta_timeseries\`/\`_s3_object_mtime\` functions. My new \`_s3_cached_meta_filenames\` just reuses them, same as the rest of the file. This isn't dead/untested code either: \`test_list_cached_meta_tickers_s3\` in \`tests/test_data_quality_cache_helpers.py\` calls \`list_cached_meta_tickers()\` with a mocked S3 client and asserts on the parsed output — it exercises exactly this path and passes. The review appears to have looked only at the diff hunk without the surrounding file, which doesn't show the pre-existing imports.

**Valid gaps, now fixed (pushed in e0773348):**
- Added \`test_find_gaps_ignores_uk_bank_holiday\` — a gap spanning Christmas (25th holiday, 26th Sat, 27th Sun, 28th Boxing Day substitute) is correctly not flagged.
- Added \`test_find_outliers_missing_close_column_returns_empty\`.
- Documented the \`min_periods=window//2\` early-series trade-off in \`find_outliers\`' docstring, per your note — the behavior is intentional (short-history tickers still get some outlier coverage) but the trade-off wasn't written down.

**Not changing, with reasoning:**
- \`find_gaps\`'s O(range) cost: agreed it scans the full calendar range rather than just the data points, but at this repo's actual scale (~76 tickers, ~10 years max) that's a few thousand date objects per ticker — not worth adding complexity for. Flagged as a non-blocking note in your review, and I agree with that framing.
- Caching \`uk_business_days_between\` per (start, end): same reasoning — premature at current data volume, and the issue's clarified scope (see the issue body) explicitly asks not to pre-optimize for scale that isn't there yet.

PR #4950, opened 2026-07-09
feat(backend): Moneyhub CSV importer with dedupe-on-import

### LINKED ISSUE #3426: Moneyhub CSV/OFX export parser to import transactions
# Moneyhub CSV/OFX export parser to import transactions

## What
Parse a Moneyhub manual export (CSV/OFX) and import the transactions into the existing transaction model.

## Why
This is a standalone, mechanical fallback that delivers value regardless of whether the live-API approach (#2749 / its implementation child #3425) proves viable. Split from #2749.

## Blocked on (added after review, 2026-07-08)
- **No sample Moneyhub export exists anywhere in the repo.** A coding agent cannot verify CSV vs. OFX format, column names, or a stable transaction id field without one. Attach a real (anonymized) sample export to this issue before implementation starts — otherwise the parser will be guessing at the format, which directly risks the "Failure looks like" scenario below.

## How
- Add a new provider entry to the dispatcher in `backend/importers/__init__.py` (`_IMPORTER_PATHS`, currently registers `degiro`, `hargreaves`, `test`), following the existing `backend/importers/degiro.py` as the template for CSV parsing with `csv.DictReader` into `Transaction` objects. (`hargreaves.py` is also CSV-based but parses holdings, not transactions, so it's not the right template here.)
- No OFX parsing exists anywhere in the repo today — write it from scratch if the Moneyhub export format requires it (check the actual sample export first; CSV may be sufficient).
- `backend/routes/transactions.py`'s `POST /transactions/import` endpoint currently only parses and returns transactions (via `backend.importers.parse(provider, data)`) — it does **not** persist or deduplicate against existing records. This issue's importer must write results through that persistence/dedupe path once it exists, or (if it still doesn't) add dedupe-on-import logic against `_load_all_transactions()` results, keyed on a stable Moneyhub transaction id.
- **Coordinate with #3425**: that issue independently requires the same net-new dedupe logic (also keyed on a stable Moneyhub transaction id, also confirmed as having no existing pattern to reuse). Whichever of #3426/#3425 lands first should build dedupe as a shared, provider-agnostic helper (not Moneyhub-CSV-specific) so the second issue doesn't reimplement it. Check for that helper before writing a new one. Note: the *key derivation* itself likely can't be fully shared — a CSV export has no source-provided transaction id, so this issue will need a synthetic composite key (e.g. date+ticker+amount+account), whereas #3425's API-based import gets a real external id passed through. Only the comparison/storage logic (check a key against existing records) is genuinely shareable.
- `scripts/import_transactions.py` is a thin CLI uploader (POSTs a file to `/transactions/import`); it has no mapping/dedupe logic of its own and does not need changes for this issue.
- Add pytest coverage with a small sample export fixture under `tests/`. **Note (correction, 2026-07-08 review):** the existing importer test (`tests/test_transactions_import.py`) builds CSV content as inline Python strings rather than a file fixture — there is no existing "sample export fixture" file/directory convention to copy. Introducing one here (e.g. under `tests/fixtures/`) is reasonable but new, not a reuse of an existing pattern.

## Constraints
- Manual-export ingestion only; no Moneyhub API/auth work (that's #3425 territory).
- Dedupe logic does not currently exist anywhere in the import path — do not assume it can be reused as-is.

## LLM tier
**Sonnet** (raised from Haiku after review, 2026-07-08) — the dispatcher/parser wiring itself is Haiku-mechanical, but this issue also requires designing net-new dedupe logic against `_load_all_transactions()` with no existing pattern to copy. #3425 scores that same kind of work as Sonnet-tier for the same reason; scoring it Haiku here was inconsistent. If dedupe is descoped from this issue (left entirely to #3425 or a follow-up), this can revert to Haiku — see Success criteria below for the descoped alternative.

## Success looks like
- [ ] New `moneyhub` entry in `backend/importers/_IMPORTER_PATHS` with a parser for the manual export format (CSV and/or OFX as needed), verified against a real attached sample export
- [ ] Imported transactions are deduplicated against existing records (not just parsed and discarded) — or, if descoped, the importer clearly returns unpersisted transactions and a follow-up issue is filed for dedupe/persistence
- [ ] Pytest coverage under `tests/` using a small sample export fixture

## Failure looks like
- Parser silently drops or duplicates rows on re-import of the same export file
- Parser only handles CSV when the real Moneyhub export is OFX (or vice versa) without being verified against an actual sample export
- Dedupe logic is built twice (once here, once in #3425) because neither issue checked whether the other had already added a shared helper

leonarduk:
## Summary
- Adds a `moneyhub` provider to `backend/importers/_IMPORTER_PATHS`, following `degiro.py`'s CSV template (see [backend/importers/moneyhub.py](backend/importers/moneyhub.py)). CSV only — no OFX parser, since neither format could be verified against a real export.
- **No real Moneyhub sample export exists anywhere in the repo or issue.** The column layout (`Owner`, `Account`, `Date`, `Amount`, `Description`, `Category`, optional `Id`) is a documented best-effort guess and should be verified/corrected against a real anonymized export before this is trusted in production.
- Adds a shared, provider-agnostic `importers.dedupe_against_existing()` helper and wires it into `POST /transactions/import`, so re-importing the same export file no longer surfaces already-persisted rows to the caller. The endpoint skips the store lookup entirely when no parsed row carries a stable id, so degiro/hargreaves imports pay no added cost.
- Dedupe key derivation: uses a real `Id` CSV column when present, otherwise falls back to a synthetic `date+account+amount+description` composite key — per updated guidance on issue #3426 noting a manual CSV export isn't guaranteed to carry a source id. The composite key can under-dedupe two genuinely distinct rows that share all four fields; that's a documented, accepted trade-off in the absence of a real id.
- Adds `Transaction.external_id` (backend model + SPA response contract, both Python and TypeScript sides). This was necessary because the existing `Transaction.id` is a synthetic `owner:account:index` value regenerated positionally on every load (`_build_transaction_id`) — it can never match across a re-import of the same source row, so a real stable key had to be introduced. Bumped `SPA_RESPONSE_CONTRACT_VERSION` accordingly and kept the Python/TypeScript copies in sync per `scripts/check_contract_version_sync.py`.
- **Persistence is deliberately out of scope**, matching the parent issue's own descoped alternative: `POST /transactions/import` parses and dedupes but does not persist, the same as it already behaves for degiro/hargreaves today (this is pre-existing behavior, not a regression). Filed #4965 to track wiring actual persistence for all three providers.

## Test plan
- [x] `python -m pytest tests/ --no-cov` — full suite passing (excluding pre-existing, unrelated WSL/bash-dependent failures in `test_review_comment.py`)
- [x] `npx vitest --run` (frontend) — passing (1 pre-existing flaky jsdom-navigation test, passes in isolation)
- [x] `ruff` / `black` / `eslint` / `tsc --noEmit` / `mypy` clean on all touched files
- [ ] Verify parser column assumptions against a real Moneyhub export once one is available (tracked as an open risk, not blocking per explicit product decision to proceed with a documented best-effort format)

Closes #3426
Follow-up filed: #4965

### DIFF
--- a/backend/contracts_spa.py
+++ b/backend/contracts_spa.py
@@ -4,7 +4,7 @@ from typing import List
 
 from pydantic import BaseModel, ConfigDict, Field
 
-SPA_RESPONSE_CONTRACT_VERSION = "2026-03-22"
+SPA_RESPONSE_CONTRACT_VERSION = "2026-07-08"
 
 
 class SpaContractBase(BaseModel):
@@ -122,6 +122,7 @@ class TransactionContract(SpaContractBase):
     owner: str
     account: str
     id: str | None = None
+    external_id: str | None = None
     date: str | None = None
     ticker: str | None = None
     type: str | None = None
--- a/backend/importers/__init__.py
+++ b/backend/importers/__init__.py
@@ -50,8 +50,11 @@ def dedupe_against_existing(candidates: List[Transaction], existing: List[Transa
     ``_build_transaction_id`` in ``backend.routes.transactions``), so it never
     matches across a re-import of the same source row. Candidates without an
     ``external_id`` have no stable key to compare against and are always
-    treated as new. Shared across import providers so this scheme isn't
-    reimplemented per-provider.
+    treated as new (this includes providers like ``degiro``/``hargreaves``
+    that don't set ``external_id`` at all). Shared across import providers
+    so this scheme isn't reimplemented per-provider -- see issue #3425,
+    which independently needs the same dedupe behaviour for the live
+    Moneyhub API importer.
     """
     existing_ids = {t.external_id for t in existing if t.external_id}
     return [t for t in candidates if not (t.external_id and t.external_id in existing_ids)]
--- a/backend/importers/__init__.py
+++ b/backend/importers/__init__.py
@@ -18,6 +18,7 @@ class UnknownProvider(Exception):
 _IMPORTER_PATHS: Dict[str, str] = {
     "degiro": "backend.importers.degiro",
     "hargreaves": "backend.importers.hargreaves",
+    "moneyhub": "backend.importers.moneyhub",
     "test": "backend.importers.test",
 }
 
@@ -37,3 +38,20 @@ def parse(provider: str, data: bytes) -> List[Transaction]:
         raise UnknownProvider(provider)
     module: Callable[[bytes], List[Transaction]] = import_module(module_path)
     return module.parse(data)  # type: ignore[attr-defined]
+
+
+def dedupe_against_existing(candidates: List[Transaction], existing: List[Transaction]) -> List[Transaction]:
+    """Filter ``candidates`` down to those not already present in ``existing``.
+
+    Comparison is keyed on ``Transaction.external_id`` -- the source
+    provider's own stable transaction id (e.g. Moneyhub's per-row ``Id``).
+    ``Transaction.id`` cannot be used for this: it is a synthetic
+    ``owner:account:index`` value regenerated positionally on every load (see
+    ``_build_transaction_id`` in ``backend.routes.transactions``), so it never
+    matches across a re-import of the same source row. Candidates without an
+    ``external_id`` have no stable key to compare against and are always
+    treated as new. Shared across import providers so this scheme isn't
+    reimplemented per-provider.
+    """
+    existing_ids = {t.external_id for t in existing if t.external_id}
+    return [t for t in candidates if not (t.external_id and t.external_id in existing_ids)]
--- a/backend/importers/moneyhub.py
+++ b/backend/importers/moneyhub.py
@@ -18,7 +18,10 @@ from backend.routes.transactions import Transaction
 
 
 def _to_float(value: str | None) -> float | None:
-    if value in (None, ""):
+    if value is None:
+        return None
+    value = value.strip()
+    if not value:
         return None
     try:
         return float(value)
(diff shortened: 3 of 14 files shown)


leonarduk:
Note on the DeepSeek review history for anyone following along: across the three review rounds so far, several claims were checked against the actual repo state and didn't hold up:

- **"`tests/data/moneyhub_sample.csv` is missing from the diff"** (raised in rounds 2 and 3) — verified false each time via `gh pr diff --name-only` and `gh api .../contents/tests/data/moneyhub_sample.csv?ref=...`. The file is present and tracked.
- **"frontend `api.ts` needs a snake_case→camelCase mapping for `external_id`"** (round 2) — verified false; no such transform layer exists in `frontend/src/api.ts` for this field.
- **"`_composite_key`'s `amount` parameter is never `None` when called"** (round 3) — verified false; `moneyhub.py`'s `parse()` calls `_composite_key(date, account, amount, description)` with `amount = _to_float(...)`, which returns `None` for missing/malformed `Amount` columns. `test_moneyhub_parse_unrecognised_columns_yields_empty_fields` exercises exactly this path.

Two claims across the rounds *did* hold up and were acted on:
- Round 3 caught that the dedupe key assumed a real `Id` CSV column when the (subsequently updated) issue text says a manual export may not have one — fixed in `2c6c032a` with an `Id`-when-present, composite-key-fallback approach.
- Round 3's point that persistence being out of scope wasn't documented in the PR body — fixed by updating the PR description and filing #4965.

Flagging this pattern in case the required check keeps failing on repeated false claims — happy to keep addressing real findings each round, but wanted the hallucination rate visible rather than buried across separate review comments.

PR #4964, opened 2026-07-09
test: add direct unit test for sanitise_log_value

### LINKED ISSUE #4868: Add a test for `sanitise_log_value` directly (unit test, not integration) to document its exact behavior with edge cases like null bytes, Unicode control chars
# Add a direct unit test for `sanitise_log_value`

## What

Add a new unit test (not integration test) for the `sanitise_log_value` function in the existing test file. The function is located in the codebase (likely in a logging utility module). The test should cover edge cases including:
- Null bytes (`\x00`)
- Unicode control characters (e.g., `\u0000`–`\u001f`, `\u007f`)
- CRLF sequences (`\r\n`)
- Mixed normal and control characters
- Empty string input
- Already-safe strings (no-op case)

## Why

The current integration tests (`test_load_account_falls_back_sanitises_log_fields`, `test_check_transactions_logs_sanitised_invalid_share_count`) verify sanitization indirectly through log record objects. As noted in the AI review, these tests don't prove `sanitise_log_value()` is the function being called — they only check the final sanitized value. A direct unit test:
- Documents exact behavior for edge cases (null bytes, Unicode control chars)
- Prevents regressions if the function is refactored or replaced
- Reduces agent confusion about what the function actually does
- Improves maintainability by making the contract explicit

## How

1. Locate the test file containing `test_load_account_falls_back_sanitises_log_fields` (likely `tests/test_logging.py` or similar)
2. Import `sanitise_log_value` from its source module
3. Add a new test function (e.g., `test_sanitise_log_value_direct`) that:
   - Tests null byte removal: `assert sanitise_log_value("bad\x00value") == "badvalue"`
   - Tests CRLF stripping: `assert sanitise_log_value("line\r\nbreak") == "linebreak"`
   - Tests Unicode control chars: `assert sanitise_log_value("ctrl\u0001char") == "ctrlchar"`
   - Tests empty string: `assert sanitise_log_value("") == ""`
   - Tests safe string unchanged: `assert sanitise_log_value("hello world") == "hello world"`
   - Tests mixed edge cases in one string
4. Use `pytest` parameterization if multiple cases are clean (optional but preferred)

## Constraints

- **Must not break** existing integration tests or modify `sanitise_log_value` itself
- **Out of scope**: Changing the function's implementation, adding integration tests, or modifying other sanitization functions
- The test should be self-contained and not require database or network access
- Follow existing test patterns in the file (same assertions style, same imports)

## LLM tier

**Sonnet** — This task requires moderate design judgment: deciding which edge cases to include, understanding the function's contract from the codebase, and writing parameterized tests that document behavior without over-specifying implementation details. The test must be correct for the actual function signature and behavior, which may require reading the source.

## Success looks like

- New test function `test_sanitise_log_value_direct` (or similar) exists in the test file
- All edge cases listed above are covered with explicit assertions
- `pytest` passes with the new test included
- The test is clearly documented with comments explaining each edge case

## Failure looks like

- Test passes but doesn't actually test the function (e.g., tests a different function or uses wrong import)
- Test is too vague (e.g., only tests one case) or too brittle (e.g., depends on implementation details like regex patterns)
- Test breaks existing tests due to import conflicts or side effects
- Test fails on valid inputs that should be preserved

_Follow-up from AI review of PR #4863._

leonarduk:
## Summary
- Adds `test_sanitise_log_value_direct`, a parametrized unit test that calls `sanitise_log_value` directly instead of only inferring its behaviour from log records.
- Documents the function's actual contract: it strips CRLF (`\r`, `\n`) to prevent CWE-117 log injection, coerces non-strings via `str()`, and leaves null bytes / other Unicode control characters untouched (it does not attempt general control-character sanitisation).

Note: the issue's "How" section assumed the function also strips null bytes and Unicode control characters. I verified against the actual implementation (`backend/logging_setup.py::sanitise_log_value`, which only does `.replace("\r", "").replace("\n", "")`) and wrote the test to match real behaviour, per the issue's own failure criteria ("Test fails on valid inputs that should be preserved").

Closes #4868

## Test plan
- [x] `python -m pytest tests/test_data_loader_aws.py -q --no-cov` — 32 passed
- [x] `python -m pytest tests/test_data_loader_aws.py -k sanitise --no-cov -v` — 11 passed
- [x] `ruff check tests/test_data_loader_aws.py` — no new findings (pre-existing I001/B010 unrelated to this change)

### DIFF
--- a/tests/test_data_loader_aws.py
+++ b/tests/test_data_loader_aws.py
@@ -6,6 +6,7 @@ import sys
 from types import SimpleNamespace
 
 import backend.common.data_loader as dl
+from backend.logging_setup import sanitise_log_value
 from botocore.exceptions import ClientError
 import pytest
 
@@ -337,6 +338,39 @@ def test_load_account_falls_back_sanitises_log_fields(
     assert "\n" not in record.getMessage()
     assert record.owner == "ownerFAKE LOG LINE"
     assert record.account == "acctinjected"
+
+
+@pytest.mark.parametrize(
+    "value, expected",
+    [
+        # CRLF sequences are stripped (the CWE-117 log-injection vector this
+        # function guards against).
+        ("line\r\nbreak", "linebreak"),
+        ("bare\rcr", "barecr"),
+        ("bare\nlf", "barelf"),
+        # Mixed normal text with a CRLF injection attempt.
+        ("owner\r\nFAKE LOG LINE", "ownerFAKE LOG LINE"),
+        # Empty string is a no-op.
+        ("", ""),
+        # Already-safe strings pass through unchanged.
+        ("hello world", "hello world"),
+        # Non-string input is coerced to str first.
+        (123, "123"),
+        # Null bytes and other Unicode control characters are NOT stripped;
+        # sanitise_log_value only targets CRLF (CWE-117), not general control
+        # characters, so they must be preserved as-is.
+        ("bad\x00value", "bad\x00value"),
+        ("ctrl\x01char", "ctrl\x01char"),
+        ("del\x7fchar", "del\x7fchar"),
+    ],
+)
+def test_sanitise_log_value_direct(value, expected):
+    """Direct unit test documenting sanitise_log_value's exact contract.
+
+    Complements the integration tests above (which only assert on the final
+    log record) by exercising the function itself for each edge case.
+    """
+    assert sanitise_log_value(value) == expected
 
 
 def test_load_account_missing_bucket(monkeypatch):


leonarduk:
Thanks for the review. Verified against the actual implementation rather than the issue text (which was itself inaccurate) — the test as written is correct and passing.

**Regarding null bytes / control chars:** `backend/logging_setup.py:16-22` (unchanged by this PR):

```python
def sanitise_log_value(value: object) -> str:
    """Return a single-line string safe for plain-text logs.

    Strips ``\r`` and ``\n`` so an attacker cannot inject fake log lines
    via CWE-117 (log injection).
    """
    return str(value).replace("\r", "").replace("\n", "")
```

The function's own docstring says it strips only `\r`/`\n`. It does not strip `\x00`, `\x01`, or `\x7f`. I confirmed this empirically:

```
>>> sanitise_log_value('bad\x00value')
'bad\x00value'
>>> sanitise_log_value('ctrl\x01char')
'ctrl\x01char'
>>> sanitise_log_value('del\x7fchar')
'del\x7fchar'
```

So option 2 from the review is what's actually true: the issue's acceptance criteria assumed broader sanitisation than the function performs. Per the issue's own failure criterion — "Test fails on valid inputs that should be preserved" and "Out of scope: Changing the function's implementation" — the correct move was to write the test to match real behavior, not to assert something false (which would make the test fail) or silently change the sanitiser's scope (out of scope for this issue). I called this out explicitly in the PR description and in a comment above the assertions in the test itself.

**Regarding `(123, "123")`:** this passes — `str(value)` is called unconditionally before the replacements, so non-string input is coerced. Confirmed above (`11 passed`, including `test_sanitise_log_value_direct[123-123]`).

Full test run: `python -m pytest tests/test_data_loader_aws.py -k sanitise --no-cov -v` → 11 passed, 0 failed.

I've filed a follow-up suggestion as an issue comment: the null-byte/control-char stripping described in #4868's "How" section doesn't exist yet in `sanitise_log_value` — if that's desired behavior, it needs its own implementation issue, since this PR is scoped to testing existing behavior only.

Issue #4810, opened 2026-07-07
CI: frontend vitest 'test' job OOM-crashes workers under concurrent runner load


leonarduk:
## What
The frontend \`test\` CI job (\`.github/workflows/ci.yml\`, \`npm test -- --run\` in \`frontend/\`) intermittently fails with \`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory\` in one or more vitest worker forks. Observed 3/3 times in a row on PR #4801 (branch \`fix/issue-4751-logout-button-visibility\`), each time with **zero real test failures** — 549/562 tests pass, no \`FAIL\` lines, only \"Worker exited unexpectedly\" from 3 crashed forks, leaving 3 of 87 test files never executed.

## Why
Vitest's default pool (\`forks\`) spawns roughly one fork per available CPU with no explicit memory cap in \`frontend/vitest.config.ts\`. \`ubuntu-latest\` GitHub-hosted runners are shared and their available memory varies with concurrent load from other jobs/PRs on the same runner pool — other, unrelated PRs (issue-4762, issue-4749, issue-4750) hit failures on different jobs in the same time window and passed on retry, consistent with shared-runner resource contention rather than a code defect. A local repro of the specific test files touched by PR #4801 showed no memory leak, ruling out that PR's diff as the cause.

## How
- In \`frontend/vitest.config.ts\`, cap worker concurrency (e.g. \`test.poolOptions.forks.maxForks\`) to a conservative value so the full suite's peak memory footprint fits comfortably within the runner's available RAM, trading some wall-clock time for reliability.
- Consider also setting an explicit \`NODE_OPTIONS\`/\`execArgv\` heap size per fork if a concurrency cap alone doesn't fully resolve it.
- Verify with several repeated CI runs (or local repro with \`--logHeapUsage\`) that the full suite completes without OOM.

## Constraints
- Don't change test behavior/coverage thresholds — this is a CI reliability fix only.
- Keep the change minimal and scoped to vitest/CI concurrency config.

## LLM tier
**Sonnet** — requires reasoning about CI runner resource limits and vitest pool configuration; low risk of behavior change but needs verification across repeated runs.

## Success looks like
- [ ] The frontend \`test\` CI job completes without worker OOM crashes across several consecutive runs
- [ ] Test coverage and pass/fail results are unchanged

## Failure looks like
- The job still OOMs intermittently after the concurrency change
[...]

PR #4974, opened 2026-07-09
test(frontend): add integration test for full logout flow (#4802)

### LINKED ISSUE #4802: Consider adding an integration test that verifies the full logout flow (click → session cleared → redirect) for both auth modes
# Integration test for full logout flow (click → session cleared → redirect)

## What

Add an integration test (likely in `cypress/integration/` or `src/__tests__/integration/`) that verifies the complete logout lifecycle for both auth modes:

- **Cognito mode**: Click logout → `cognitoLogout()` called → session cleared → redirect to `/`
- **Local dev mode**: Click logout → `navigate('/')` called → redirect occurs

## Why

The current unit tests only verify that the correct logout function is *available* and *rendered*, but don't confirm the full end-to-end behavior:

- A regression could silently break the actual logout execution (e.g., `cognitoLogout()` throws, redirect fails) while unit tests still pass
- The two auth modes have different logout implementations; an integration test ensures both paths work correctly
- Reduces risk of "agent confusion" where future changes assume logout works without verifying the full flow

## How

1. **Create test file**: `cypress/e2e/logout-flow.cy.ts` (or equivalent in existing test structure)

2. **Test structure**:
   ```typescript
   describe('Logout flow', () => {
     it('Cognito mode: logout clears session and redirects', () => {
       // Mock Cognito session
       // Click logout button
       // Assert cognitoLogout() was called
       // Assert session state cleared
       // Assert redirect to '/'
     })

     it('Local dev mode: logout navigates to /', () => {
       // Set local auth mode
       // Click logout button
       // Assert navigate('/') was called
       // Assert redirect occurred
     })
   })
   ```

3. **Key assertions**:
   - Logout button click triggers the correct function
   - `AuthContext` session state is cleared (e.g., `profile` becomes `null`)
   - URL changes to `/` (or appropriate post-logout route)
   - No console errors

## Constraints

- **Must not break**: Existing unit tests (`renders logout via AuthContext...`, `hides logout when neither...`)
- **Out of scope**: Backend API changes, CI/CD modifications, UI styling
- **Test isolation**: Each test should clean up mocks/state to avoid interference
- **Auth mode switching**: Tests should explicitly set auth mode, not rely on environment defaults

## LLM tier

**Sonnet** — This requires:
- Understanding the existing test infrastructure (Cypress vs Jest, mock patterns)
- Designing assertions for both auth modes with different logout implementations
- Setting up proper mocks for `cognitoLogout()` and `navigate()`
- Following existing integration test patterns in the repo

A smaller model (local-14b) could attempt this if given explicit file paths and mock patterns, but the risk of missing edge cases (e.g., async logout, race conditions) makes Sonnet the safer choice.

## Success looks like

- Both tests pass in CI for all auth mode configurations
- A deliberate break in `cognitoLogout()` or `navigate('/')` causes the corresponding test to fail
- Tests run in < 5 seconds each (no unnecessary waits)

## Failure looks like

- Tests pass but don't actually verify the full flow (e.g., only check button click, not session clearing)
- Tests are flaky due to async timing issues
- Tests require manual environment setup (should be fully automated)
- Tests break when run alongside other integration tests (state leakage)

_Follow-up from AI review of PR #4801._

leonarduk:
## Summary
- The unit tests in `Menu.test.tsx` only assert that the logout callback is *invoked* (via a mocked `contextLogout`), not that the real logout lifecycle in `src/main.tsx`'s `Root` component behaves correctly.
- Adds two Vitest integration tests that mount the real `Root` component (`AuthProvider` + `UserProvider` + `BrowserRouter` + `Root`, following the existing `rootBootstrap.test.tsx` convention) and click through the actual logout callback end to end:
  - `frontend/tests/unit/logoutFlowCognito.test.tsx`: Cognito mode — click → `cognitoLogout()` redirects to the hosted-UI logout endpoint via `window.location.assign()`, and `authToken`/`auth.user`/`user.profile`/`awsUiAuthSession` are all cleared before the redirect.
  - `frontend/tests/unit/logoutFlowLocalDev.test.tsx`: local dev mode (`disable_auth: true`, no `awsUiAuth`) — click → `navigate('/')` (opting back into the real `useNavigate` per the `App.test.tsx` convention, since `src/setupTests.ts` globally stubs it), session cleared, no Cognito round trip.
- Note: the repo doesn't use Cypress (issue #4802 assumed it did) — Playwright is the e2e tool and Vitest+RTL is the existing "integration test" convention (see `rootBootstrap.test.tsx`, `main.cognito*.test.tsx`), so these tests follow that established pattern instead.

Closes #4802

## Test plan
- [x] `npx vitest run tests/unit/logoutFlowCognito.test.tsx tests/unit/logoutFlowLocalDev.test.tsx` — both pass
- [x] `npx vitest run tests/unit/components/Menu.test.tsx tests/unit/rootBootstrap.test.tsx tests/unit/App.test.tsx` — all 48 existing tests still pass (no regressions)
- [x] `npx eslint` and `npx prettier --check` clean on the new files

## Summary by CodeRabbit

* **Tests**
  * Added coverage for logout behaviour in both Cognito and local development modes.
  * Verified that signing out clears saved session data and returns users to the login screen.
  * Confirmed redirect behaviour differs correctly between hosted sign-out and local sign-out flows.

### DIFF
--- a/frontend/tests/unit/logoutFlowCognito.test.tsx
+++ b/frontend/tests/unit/logoutFlowCognito.test.tsx
@@ -1,6 +1,7 @@
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { BrowserRouter } from 'react-router-dom';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import i18n from '@/i18n';
 
 // Full logout lifecycle in Cognito mode (issue #4802): click -> cognitoLogout()
 // called -> session cleared -> redirect to the Cognito hosted-UI logout
@@ -80,13 +81,17 @@ describe('Logout flow: Cognito mode (#4802)', () => {
       default: () => <div data-testid="login-page">sign in</div>,
     }));
 
-    vi.doMock('@/App.tsx', () => ({
-      default: ({ onLogout }: { onLogout?: () => void }) => (
-        <button type="button" data-testid="logout-button" onClick={onLogout}>
-          logout
-        </button>
-      ),
-    }));
+    // Render the real Menu component (the actual logout button end users
+    // click), not a stand-in, so a regression in Menu's onClick wiring would
+    // be caught here too.
+    vi.doMock('@/App.tsx', async () => {
+      const { default: Menu } = await import('@/components/Menu');
+      return {
+        default: ({ onLogout }: { onLogout?: () => void }) => (
+          <Menu onLogout={onLogout} />
+        ),
+      };
+    });
 
     window.localStorage.setItem('authToken', 'cognito-id-token');
     window.localStorage.setItem(
@@ -107,7 +112,13 @@ describe('Logout flow: Cognito mode (#4802)', () => {
 
     await mountRoot();
 
-    const logoutButton = await screen.findByTestId('logout-button');
+    const preferencesToggle = await screen.findByRole('button', {
+      name: i18n.t('app.menuCategories.preferences'),
+    });
+    fireEvent.click(preferencesToggle);
+    const logoutButton = await screen.findByRole('menuitem', {
+      name: i18n.t('app.logout'),
+    });
     fireEvent.click(logoutButton);
 
     // cognitoLogout() was called: the hosted-UI logout endpoint is reached
@@ -125,6 +136,8 @@ describe('Logout flow: Cognito mode (#4802)', () => {
 
     // The app falls back to the login screen once authed flips to false.
     expect(await screen.findByTestId('login-page')).toBeInTheDocument();
-    expect(screen.queryByTestId('logout-button')).not.toBeInTheDocument();
+    expect(
+      screen.queryByRole('menuitem', { name: i18n.t('app.logout') })
+    ).not.toBeInTheDocument();
   });
 });
--- a/frontend/tests/unit/logoutFlowCognito.test.tsx
+++ b/frontend/tests/unit/logoutFlowCognito.test.tsx
@@ -1,0 +1,130 @@
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { BrowserRouter } from 'react-router-dom';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Full logout lifecycle in Cognito mode (issue #4802): click -> cognitoLogout()
+// called -> session cleared -> redirect to the Cognito hosted-UI logout
+// endpoint. The unit tests in Menu.test.tsx only assert that the logout
+// callback is invoked; this exercises the real callback registered by
+// src/main.tsx's Root component end to end.
+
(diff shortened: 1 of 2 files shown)


leonarduk:
Responding to the DeepSeek review:

**Fixed:** the tests originally mocked `@/App.tsx` with a bare stand-in `<button onClick={onLogout}>`, which meant a regression in the real `Menu` component's logout button (rendering, `role="menuitem"`, the `effectiveLogout` fallback wiring) wouldn't have been caught. Both tests now mock `@/App.tsx` to render the real `Menu` component and click its actual rendered "Logout" menu item (verified by running the tests — see latest commit).

**Not changed, with justification** — three of the "blocking" claims don't hold up against the actual source:

1. *"Cognito test's redirect assertion should check navigation to `/`, not the hosted-UI logout endpoint"* — the issue's AC text was itself imprecise. Reading `src/awsUiAuth.ts:163-175`, `cognitoLogout()` clears the local session then does `window.location.assign(`${domain}/logout?...`)` — it redirects to the Cognito hosted-UI logout endpoint (which itself round-trips back to `/` afterward, per its own `logout_uri` param). Asserting the hosted-UI redirect is the correct behavior to test.

2. *"cognitoLogout might use `window.location.href` instead of `.assign()`"* — checked directly against `src/awsUiAuth.ts:169`: it's `window.location.assign(...)`. Confirmed, not a bug.

3. *"`window.location.pathname` won't change from client-side `navigate('/')`"* — this repo's `BrowserRouter` genuinely does update `window.location` via `history.pushState` under the hood, and I verified this empirically (the test failed for an unrelated reason on the first pass — `src/setupTests.ts` globally stubs `useNavigate` as a no-op per #4810 — and once the test opts back into the real hook, per the existing convention in `App.test.tsx`, `window.location.pathname` does update and the assertion passes).

4. *"double-render from mocking `react-dom/client`'s `createRoot`"* — this is the same pattern already used in the merged `rootBootstrap.test.tsx` (mocking `createRoot` to a no-op so `main.tsx`'s module-level production bootstrap doesn't fire, while RTL's own `render()` mounts `<Root/>` for the test). Not a new or risky pattern.

**Cypress vs. this repo's actual conventions** — confirmed by grep: there is no `cypress/` directory anywhere in the repo. Playwright is the e2e tool (`playwright.config.ts`, `tests/*.spec.ts`), and the established "integration test" convention for auth/session flows is Vitest + RTL mounting the real `Root` component (`rootBootstrap.test.tsx`, `main.cognito*.test.tsx`). These tests follow that existing pattern rather than introducing a new Cypress dependency the codebase doesn't otherwise use.

**Valid follow-ups I did not fold into this PR** (agreed, worth a separate ticket): error-path coverage for `cognitoLogout()` throwing, and a "hides logout when neither" browser regression test at the integration level (already covered at the unit level in `Menu.test.tsx`).

PR #5002, opened 2026-07-10
fix: fall back to a default SIGNUP_LOGIN_URL instead of empty string

### LINKED ISSUE #4385: Consider making `SIGNUP_LOGIN_URL` required or providing a sensible default
Here is a complete, actionable GitHub issue body in Markdown:

```markdown
## Consider making `SIGNUP_LOGIN_URL` required or providing a sensible default

### What
In `backend/routes/signup/signup.py`, the `_login_url()` function (line 125) returns an empty string when the `SIGNUP_LOGIN_URL` environment variable is unset. This empty string is passed to `send_signup_approved_email`, which then generates an approval email with no clickable login link — just the text "You can now log in to AllotMint".

### Why
- **Poor UX**: Approved users receive an email saying they can log in, but with no link. In production, admins must manually communicate the URL.
- **Correctness risk**: An empty URL is silently accepted, masking a configuration error. If the env var is accidentally omitted during deployment, users cannot access the app.
- **Maintainability**: The current behavior is fragile — any future email template that assumes a valid URL will break silently.

### How
1. **Option A (Recommended)**: Make `SIGNUP_LOGIN_URL` required at startup. Add validation in `signup.py` (or a config module) that raises a clear `ValueError` or `RuntimeError` if the env var is missing or empty when the module loads.
2. **Option B**: Provide a sensible default, e.g., `http://localhost:8000` for development, and log a warning in production if the env var is unset.
3. **Update `_login_url()`**: If Option A is chosen, remove the `os.getenv(..., "")` fallback and use a required config value. If Option B, keep the fallback but ensure the default is a valid URL.

### Constraints
- **Do not break existing deployments** that rely on the current behavior (e.g., if someone intentionally sets the env var to empty). If Option A is chosen, this is a breaking change — document it clearly.
- **Out of scope**: Changing the email template or adding frontend components. This issue is strictly about the URL source.
- **Must not introduce circular imports** — the config change should be self-contained in `signup.py` or a new config module.

### LLM tier
**Sonnet** — The change involves a design decision (required vs. default) that affects deployment configuration and may require updating documentation or startup scripts. It is not purely mechanical (local-7b/14b) and benefits from cloud-level reasoning about trade-offs.

### Success looks like
- `_login_url()` never returns an empty string.
- If the env var is missing at startup, the application fails fast with a clear error message (Option A) OR logs a prominent warning and uses a sensible default (Option B).
- All existing tests pass, and a new test verifies the behavior when the env var is unset.

### Failure looks like
- The application starts successfully but still sends emails with no login link.
- The change introduces a regression where a previously valid empty-string configuration now crashes unexpectedly (if Option A without proper migration).
- The fix is incomplete — e.g., only patching `_login_url()` but not validating at startup.

---

_Follow-up from AI review of PR #4382._
```

leonarduk:
## Summary
- `_login_url()` in `backend/routes/signup.py` returned an empty string when `SIGNUP_LOGIN_URL` was unset, so approval emails were sent with no clickable login link and the misconfiguration was silent.
- Chose Option B from the issue (default + warning) since it matches the existing pattern already used for `SIGNUP_APPROVAL_BASE_URL` in the same file, and doesn't risk breaking deployments that never set the var.
- `_login_url()` now logs a warning and falls back to `http://localhost:5173` (the frontend dev URL) when the env var is unset/empty, so it never returns an empty string.
- Documented `SIGNUP_LOGIN_URL` in `.env.example` for discoverability.

## Test plan
- [x] `python -m pytest tests/backend/routes/test_signup.py -q` — 24 passed, including new `test_missing_login_url_warns_and_falls_back`
- [x] `python -m black --config backend/pyproject.toml --check backend/routes/signup.py tests/backend/routes/test_signup.py` — no new formatting issues introduced
- [x] `python -m ruff check --config backend/pyproject.toml backend/routes/signup.py tests/backend/routes/test_signup.py` — all checks passed

Closes #4385

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,5 +24,6 @@ METADATA_PREFIX=instruments
 VITE_APP_BASE_URL=https://app.allotmint.io
 
 # Login page URL emailed to a newly approved signup request.
-# Falls back to http://localhost:5173 (with a warning logged) if unset.
+# Required for the signup-approval flow: unset/empty makes POST /signup/approve
+# return 503 rather than send an email with no login link.
 SIGNUP_LOGIN_URL=https://app.allotmint.io/login
--- a/.env.example
+++ b/.env.example
@@ -22,3 +22,7 @@ METADATA_PREFIX=instruments
 
 # Base URL for the deployed frontend
 VITE_APP_BASE_URL=https://app.allotmint.io
+
+# Login page URL emailed to a newly approved signup request.
+# Falls back to http://localhost:5173 (with a warning logged) if unset.
+SIGNUP_LOGIN_URL=https://app.allotmint.io/login
--- a/backend/routes/signup.py
+++ b/backend/routes/signup.py
@@ -134,25 +134,16 @@ async def post_signup_request(request: Request) -> dict[str, str]:
     return await _post_signup_request_impl(request)
 
 
-_DEFAULT_LOGIN_URL = "http://localhost:5173"
-
-
 def _login_url() -> str:
     """Return the login page URL emailed to a newly approved user.
 
-    Falls back to the local frontend dev URL (and logs a warning) rather than
-    returning an empty string, so the approval email always contains a
-    clickable link even if ``SIGNUP_LOGIN_URL`` is misconfigured (#4385).
+    Callers must check this is non-empty before using it (see
+    ``approve_signup_request``) rather than silently emailing a linkless or
+    fabricated URL — a misconfigured deployment should fail loudly, not send
+    users to a broken link (#4385).
     """
 
-    url = os.getenv("SIGNUP_LOGIN_URL", "").strip()
-    if not url:
-        logger.warning(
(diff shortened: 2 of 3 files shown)


leonarduk:
Addressed the DeepSeek review feedback in 1c00d896:

- Replaced the `http://localhost:5173` default with a fail-fast check: `POST /signup/approve` now returns `503` (before provisioning) if `SIGNUP_LOGIN_URL` is unset/empty, instead of ever sending a login email with a dev-only or missing link. This matches Option A from the issue and the existing `SIGNUP_ADMIN_EMAIL` pattern already in this file.
- Added a parametrized test covering both the unset and explicit-empty-string (`SIGNUP_LOGIN_URL=""`) cases, asserting no provisioning/email happens.
- Updated `.env.example` to reflect that the var is now required for the approval flow to succeed.

All 25 signup tests pass; black/ruff clean against `backend/pyproject.toml`.

PR #5003, opened 2026-07-10
fix(frontend): show loading/empty states for authorized owners in UserConfigPage

### LINKED ISSUE #4725: Consider adding a frontend follow-up to actually scope the dropdown in `UserConfigPage`
### What

Add frontend logic to `UserConfigPage` to filter the dropdown options based on the owner's authorized viewers/email list, using the backend authorization layer added in PR #4717.

**Location:** `frontend/src/pages/UserConfigPage.tsx` (or equivalent frontend component)

### Why

The backend now enforces `ensure_owner_access` on `/user-config/<owner>` and `/accounts/<owner>/approvals`, but the frontend dropdown still shows all owners regardless of authorization. This creates a poor UX where users can select an owner they don't have access to, only to receive a 403 error on save. It also risks agent confusion when automated tests or scripts interact with the dropdown.

### How

1. **Fetch authorized owners list** from a new or existing endpoint (e.g., `/user-config/authorized-owners` or parse from the current user's session data)
2. **Filter dropdown options** in `UserConfigPage` to only include owners the current user is authorized to access
3. **Handle edge cases:** Show a meaningful empty state if no owners are authorized, and preserve the current "loading" state while fetching

### Constraints

- **Must not break** existing owner selection behavior for authorized users
- **Out of scope:** Changing the backend authorization model, adding new endpoints, or modifying the dropdown component library
- **Must preserve** the existing error handling for unauthorized access (backend still returns 403 as fallback)

### LLM Tier

**Sonnet** — requires moderate design judgment to determine how to fetch the authorized owners list (existing endpoint vs. new one) and how to integrate with the current frontend state management pattern.

### Success looks like

- Dropdown only shows owners the current user is authorized to access
- Unauthorized owners are not visible in the dropdown
- Empty state is displayed when user has no authorized owners
- All existing tests pass

### Failure looks like

- Dropdown shows all owners (no filtering applied)
- Dropdown incorrectly filters out authorized owners
- Frontend crashes or shows loading state indefinitely
- Existing owner selection/editing workflow breaks

---

_Follow-up from AI review of PR #4717._

leonarduk:
## Summary

- Closes #4725.
- The `/owners` endpoint (`backend/routes/portfolio.py`) already scopes the returned list to owners the current authenticated identity may access, via `data_loader.list_plots` → `_is_authorized` → `identity_can_access_owner` (`backend/common/authz.py`, added in #4717 and the earlier `_list_local_plots` scoping). So `UserConfigPage`'s dropdown was already only ever populated from an authorized owner list — there was no unscoped/"show everything" bug to fix in the fetch itself.
- What was actually missing, per the issue's "Constraints" (no backend/endpoint changes) and "Handle edge cases" step: the page gave no feedback while `getOwners()` was in flight, and rendered a bare `<select>` with only the placeholder option when the authorized list came back empty — exactly the kind of silent, confusing state the issue calls out.
- Added an explicit `ownersLoading` state: shows a "Loading owners..." message while the owners fetch is in flight (dropdown is not rendered until it resolves), and shows a "No accounts are available..." message when the (already-authorized) list is empty, instead of an empty dropdown.
- Existing owner-selection behavior (auto-selecting the logged-in user's owner, fetching config/approvals on selection) is unchanged.

## Test plan

- [x] `npx vitest run tests/unit/pages/UserConfig.test.tsx` — 5/5 passing (added loading-state and empty-state cases)
- [x] `npx vitest run` (full frontend suite) — 591/591 passing, no regressions
- [x] `npx eslint src/pages/UserConfig.tsx tests/unit/pages/UserConfig.test.tsx` — clean

### DIFF
--- a/frontend/src/pages/UserConfig.tsx
+++ b/frontend/src/pages/UserConfig.tsx
@@ -18,6 +18,7 @@ export default function UserConfigPage() {
   const { user } = useAuth();
   const { theme } = useConfig();
   const [owners, setOwners] = useState<OwnerSummary[]>([]);
+  const [ownersLoading, setOwnersLoading] = useState(true);
   const [owner, setOwner] = useState('');
   const [cfg, setCfg] = useState<UserConfig>({});
   const [status, setStatus] = useState<string | null>(null);
@@ -33,8 +34,9 @@ export default function UserConfigPage() {
     getOwners()
       .then((os) => setOwners(sanitizeOwners(os)))
       .catch(() => {
-        /* ignore */
-      });
+        setOwners([]);
+      })
+      .finally(() => setOwnersLoading(false));
   }, []);
 
   useEffect(() => {
@@ -148,18 +150,31 @@ export default function UserConfigPage() {
           </p>
         </section>
       )}
-      <select
-        className="w-full border p-2"
-        value={owner}
-        onChange={(e) => setOwner(e.target.value)}
-      >
-        <option value="">{t('userConfig.selectOwner', 'Select owner')}</option>
-        {owners.map((o) => (
-          <option key={o.owner} value={o.owner}>
-            {o.full_name?.trim() ? o.full_name : o.owner}
-          </option>
-        ))}
-      </select>
+      {ownersLoading ? (
+        <p className="text-gray-800 dark:text-gray-200">
+          {t('userConfig.loadingOwners', 'Loading owners...')}
+        </p>
+      ) : owners.length === 0 ? (
+        <p className="text-gray-800 dark:text-gray-200">
+          {t(
+            'userConfig.noOwners',
+            'No accounts are available for your account. Contact an administrator if you believe this is an error.'
+          )}
+        </p>
+      ) : (
+        <select
+          className="w-full border p-2"
+          value={owner}
+          onChange={(e) => setOwner(e.target.value)}
+        >
+          <option value="">{t('userConfig.selectOwner', 'Select owner')}</option>
+          {owners.map((o) => (
+            <option key={o.owner} value={o.owner}>
+              {o.full_name?.trim() ? o.full_name : o.owner}
+            </option>
+          ))}
+        </select>
+      )}
       {owner && (
         <>
           <form onSubmit={save} className="space-y-2">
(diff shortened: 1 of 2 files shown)


leonarduk:
## Response to the DeepSeek automated review

The review's core claim — that this PR leaves the dropdown showing "the same unfiltered list as before" — doesn't hold up against the actual code, so I want to lay out the evidence for the record.

**`/owners` is already authorization-scoped, and has been for ~10 months:**

- [`backend/routes/portfolio.py:497-514`](https://github.com/leonarduk/allotmint/blob/main/backend/routes/portfolio.py#L497-L514) — when `config.disable_auth` is `False`, the `/owners` route resolves `current_user` via `Depends(get_current_user)` and passes it into `data_loader.list_plots(accounts_root, current_user)`.
- [`backend/common/data_loader.py` (`_is_authorized` inside `_list_local_plots`, ~L514-537, L561-562)](https://github.com/leonarduk/allotmint/blob/main/backend/common/data_loader.py) — every owner is checked against `identity_can_access_owner(user, owner, meta)` before being included; unauthorized owners are dropped.
- [`backend/common/authz.py:47-52`](https://github.com/leonarduk/allotmint/blob/main/backend/common/authz.py#L47-L52) — `identity_can_access_owner` is the *same* function `ensure_owner_access` (added in #4717) uses to gate `/user-config/<owner>` and `/accounts/<owner>/approvals`. Same authorization rule, same code path — not a separate/looser check.
- This scoping was introduced in commit `52ecb80a` ("Filter unauthorized owners from local plot listing"), dated **2025-09-27** — nearly ten months before this issue (#4725) was filed on 2026-07-06. So `getOwners()` was never unfiltered at any point relevant to this PR; it was already filtered before and remains filtered after.

**Why no additional client-side filter was added:**

The `OwnerSummary` shape returned to the frontend ([`frontend/src/types.ts`](https://github.com/leonarduk/allotmint/blob/main/frontend/src/types.ts)) exposes only `owner`, `email`, `full_name`, `accounts` — it does not expose the owner's `viewers` list. Re-implementing `identity_can_access_owner` client-side using only `owner`/`email` would incorrectly exclude legitimate viewer-granted access (a user authorized via another owner's `viewers` list, not their own `email`), which would violate the issue's own constraint: "Must not break existing owner selection behavior for authorized users." Building a client-side check that's actually correct would require a new endpoint exposing viewer grants — which the issue explicitly puts out of scope ("Out of scope: ... adding new endpoints").

Given that, the dropdown already contains only authorized owners by construction (backend-enforced). The remaining, in-scope gap was the missing loading/empty-state feedback described in the issue's "Handle edge cases" step, which is what this PR implements.

Happy to be corrected if there's a specific scenario where `/owners` returns an unauthorized owner today — but I haven't found one in the current code path.

PR #5021, opened 2026-07-10
fix(instruments): handle read-only filesystem gracefully in save_instrument_meta

### LINKED ISSUE #5019: `fix(instruments): handle read-only filesystem gracefully in save_instrument_meta`
Here is the complete GitHub issue body:

---

### What

In `instruments.py`, the function `save_instrument_meta` currently assumes the filesystem is writable. When the Lambda runs in a read-only filesystem environment (e.g., after the IAM gap fix in PR #5016 unblocks the normal read path), this function will raise an unhandled exception instead of degrading gracefully.

The change should be made in `instruments.py`, specifically in the `save_instrument_meta` function (or the surrounding write logic), to catch filesystem write failures and either log a warning or skip the write without crashing.

### Why

- **Correctness risk:** A read-only filesystem is a valid runtime configuration for the Lambda. The current code will crash with an `OSError` or `PermissionError` when it tries to write, breaking the entire instrument processing pipeline.
- **Maintainability:** The IAM gap fix in PR #5016 now allows the Lambda to reach this code path more reliably, making the crash scenario more likely.
- **Agent confusion:** A hard crash in a non-critical write path (caching metadata locally) is misleading — the system should continue operating with degraded local caching, not fail entirely.

### How

1. Wrap the filesystem write operation(s) in `save_instrument_meta` with a `try/except` block.
2. Catch `OSError` (or more specifically `PermissionError` / `IOError`).
3. Log a warning (e.g., `logger.warning("Cannot write instrument meta to read-only filesystem: %s", e)`) and return gracefully — either `None` or a fallback value.
4. Ensure the caller of `save_instrument_meta` handles the graceful return (e.g., skips caching but continues processing).

**Example sketch:**

```python
def save_instrument_meta(meta: dict, path: str) -> Optional[str]:
    try:
        with open(path, "w") as f:
            json.dump(meta, f)
        return path
    except OSError as e:
        logger.warning("Cannot write instrument meta to %s: %s", path, e)
        return None
```

### Constraints

- **Must not break** the existing write path when the filesystem *is* writable (local dev, tests).
- **Must not change** the Yahoo Finance read path or any other data-fetching logic — this is purely a write-sideguard.
- **Out of scope:** Removing the write entirely, refactoring the caching strategy, or changing the IAM policy.

### LLM Tier

**Sonnet** — This is a moderate design judgment task. It requires understanding the function’s contract, the caller’s expectations, and the correct exception hierarchy. A smaller model might miss edge cases (e.g., partial writes, atomicity concerns) or introduce overly broad exception handling.

### Success looks like

- `save_instrument_meta` no longer raises an unhandled exception on a read-only filesystem.
- A warning is logged when the write fails.
- The caller continues processing without crashing.
- All existing tests pass (including any tests that verify the write path works normally).

### Failure looks like

- The function silently swallows all exceptions (including unrelated ones like `TypeError` from bad data).
- The function returns a value that the caller interprets as a successful write when it actually failed.
- The change introduces a regression in the normal writable-filesystem path (e.g., a typo in the file path or missing import).

---

_Follow-up from AI review of PR #5016._

leonarduk:
## Summary
- `save_instrument_meta` in `backend/common/instruments.py` now catches `OSError` from the local metadata write and logs a warning instead of raising, so a read-only Lambda filesystem no longer crashes the instrument processing pipeline.
- The S3 upload path is unaffected — it still runs even when the local write fails.
- Return type changed to `Optional[Path]`; verified all callers (`instrument_admin.py`, internal auto-create path) discard or don't branch on the return value, so this is safe.

## Test plan
- [x] Added `test_save_instrument_meta_handles_read_only_filesystem` covering the `PermissionError` path, asserting `None` is returned and a warning is logged.
- [x] `python -m pytest tests/backend/common/test_instruments.py tests/backend/test_instrument_admin.py tests/test_instruments.py -q --no-cov` — 58 passed, 1 xfailed.
- [x] `python -m black --config backend/pyproject.toml --diff` / `python -m ruff check --config backend/pyproject.toml` on changed files — no new issues introduced.

Closes #5019

### DIFF
--- a/backend/common/instruments.py
+++ b/backend/common/instruments.py
@@ -195,11 +195,15 @@ def save_instrument_meta(
     ticker: str,
     exchange: str | Dict[str, Any],
     data: Optional[Dict[str, Any]] = None,
-) -> Path:
+) -> Optional[Path]:
     """Persist metadata for an instrument and optionally upload to S3.
 
     Supports calling as ``save_instrument_meta("ABC", "L", {...})`` or
     ``save_instrument_meta("ABC.L", {...})``.
+
+    Returns the path written on success, or ``None`` if the local filesystem
+    write failed (e.g. a read-only filesystem in the Lambda runtime). A
+    failed local write does not prevent the S3 upload below from proceeding.
     """
 
     if data is None:
@@ -220,9 +224,13 @@ def save_instrument_meta(
         with path.open("w", encoding="utf-8") as fh:
             json.dump(data, fh, indent=2, sort_keys=True)
             fh.write("\n")
-    except OSError as exc:  # pragma: no cover - filesystem errors are rare
-        logger.exception("Failed to write instrument metadata %s", path)
-        raise
+    except OSError as exc:
+        logger.warning(
+            "Cannot write instrument metadata to %s (read-only filesystem?): %s",
+            sanitise_log_value(path),
+            sanitise_log_value(exc),
+        )
+        path = None
 
     s3_loc = _s3_location()
     if s3_loc:
--- a/tests/backend/common/test_instruments.py
+++ b/tests/backend/common/test_instruments.py
@@ -325,6 +325,28 @@ def test_save_instrument_meta_variants(monkeypatch, tmp_path) -> None:
         instruments.save_instrument_meta("ABC", {"name": "ABC"})
 
 
+def test_save_instrument_meta_handles_read_only_filesystem(monkeypatch, tmp_path, caplog) -> None:
+    monkeypatch.setattr(instruments, "_INSTRUMENTS_DIR", tmp_path)
+    monkeypatch.setattr(instruments.config, "data_root", tmp_path)
+    monkeypatch.delenv(instruments.METADATA_BUCKET_ENV, raising=False)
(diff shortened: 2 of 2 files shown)


leonarduk:
Thanks for the review — the \`NameError\` concern is not applicable here. \`sanitise_log_value\` is already imported at the top of the file (\`from backend.logging_setup import sanitise_log_value\`, line 14) and is used elsewhere in \`instruments.py\` (e.g. the existing S3 upload warning at lines 164-166, and the JSON-decode warning at line 176). My new call in \`save_instrument_meta\` follows the same existing pattern.

This is also covered by the test suite: \`test_save_instrument_meta_handles_read_only_filesystem\` exercises the exact failure path and asserts the warning is logged successfully (no exception), and the full related test suite (58 passed, 1 xfailed) ran clean.

On the test-isolation note: agreed it's a minor concern, but \`monkeypatch.setattr(Path, "open", ...)\` is automatically reverted by pytest's \`monkeypatch\` fixture teardown after the test completes, so it can't leak into other tests.

PR #5113, opened 2026-07-17
fix(frontend): downgrade transient config-fetch failures to warn level

### LINKED ISSUE #5109: App boot fails intermittently with "Failed to load configuration: TypeError: Failed to fetch" console error
What

On initial page load (root "/" route, which redirects to /input), the browser console intermittently logs: "Failed to load configuration TypeError: Failed to fetch". This was observed right after the latest redeploy of the frontend (asset bundle index-u9ojasFL.js).

Why

This is a different failure mode than the HTTP 404/503 status errors tracked in other backend-reliability issues (issue 5072, issue 5082, issue 5088). Those are backend Lambda responses; this is a raw network-level fetch failure at app boot, before any specific data page is even reached. It could indicate a config file (e.g. config.json or similar) failing to load from the CDN/API, which may affect app initialization broadly.

How

Steps to reproduce: Navigate to the site root (https://d32n5ua14kb82t.cloudfront.net/), then open the browser console and observe it during the redirect to /input. The "Failed to load configuration TypeError: Failed to fetch" error appears in the console during this load.

Constraints

Only observed once so far during this retest pass; needs confirmation on whether it's consistently reproducible or intermittent/transient (e.g. a race condition on first load, or a flaky CDN/API response).

Success looks like

No "Failed to load configuration" fetch errors appear in the console on initial app load, across repeated tests.

Failure looks like

The console continues to show "Failed to load configuration TypeError: Failed to fetch" on some or all initial page loads, potentially indicating incomplete app initialization even when the page appears to render.

leonarduk:
## Summary
- On initial app load, `GET /config` failures were logged with `console.error` on every attempt, even though this fetch is already auto-retried up to 5 times with exponential backoff (`frontend/src/main.tsx`). A single transient `TypeError: Failed to fetch` that self-heals on the next attempt read as an unhandled console error even though the app recovers within seconds.
- Only the final, non-retryable failure now logs at `console.error`; intermediate retryable attempts log at `console.warn` instead. No behavior change to the retry/backoff/timeout logic itself.

## Root cause note
This addresses the console-noise symptom described in the issue. The issue itself flags the underlying network fetch failure as unconfirmed/intermittent (observed once) — this change doesn't claim to fix an unconfirmed backend/network root cause, only to stop a self-healing, retried failure from presenting as an app-breaking error in the console.

## Test plan
- [x] `npx eslint src/main.tsx` — clean
- [x] `npx vitest run` — 607/607 tests pass
- [ ] Manual verification on the deployed CloudFront app (can't reproduce the transient network condition locally)

Closes #5109

### DIFF
--- a/frontend/src/configFetchLogging.ts
+++ b/frontend/src/configFetchLogging.ts
@@ -1,0 +1,12 @@
+// A single transient "Failed to fetch" that self-heals on the next attempt
+// isn't an app-breaking failure — log it at warn so it doesn't read as an
+// unhandled error in the console (#5109). Only the final, non-retryable
+// failure (retries exhausted) escalates to console.error, since that is a
+// genuine, user-visible failure ("Unable to load configuration" screen).
+export const logConfigFetchFailure = (err: unknown, willRetry: boolean) => {
+  if (willRetry) {
+    console.warn('Failed to load configuration, retrying', err);
+  } else {
+    console.error('Failed to load configuration', err);
+  }
+};
--- a/frontend/src/main.tsx
+++ b/frontend/src/main.tsx
@@ -313,7 +313,6 @@ export function Root({ awsUiAuth = runtimeAwsUiAuth }: { awsUiAuth?: AwsUiAuthCo
           if (!isMounted.current || activeRequest.current !== controller)
             return;
 
-          console.error('Failed to load configuration', err);
           const error =
             err instanceof DOMException && err.name === 'AbortError'
               ? new Error('Request timed out while loading configuration.')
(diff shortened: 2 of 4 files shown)


leonarduk:
Thanks for the review — addressed the valid points, pushing back on one.

**Test coverage (fixed):** added `frontend/src/configFetchLogging.ts` extracting the warn/error decision into a small pure function, with direct unit tests for both branches (`frontend/tests/unit/logConfigFetchFailure.test.ts`), plus an assertion in the existing `configRetrySuccess.test.tsx` integration test that the transient-failure path logs `warn` and never calls `error`.

**On "no console errors ever" (pushing back):** the final, retries-exhausted attempt intentionally still logs `console.error`. That's not a bug — it's the case where `GET /config` has failed 5 times in a row and the app is about to show the user-facing "Unable to load configuration" screen. That's a genuine, unrecoverable failure and should be loud in the console for debugging; suppressing it entirely would hide real backend/network outages.

The issue as filed was a *single, unconfirmed* observation of one transient `Failed to fetch` on an otherwise-successful page load — i.e. exactly the case this PR now demotes to `warn`. Reading the AC literally as "never log this string at any level, even on total failure" would mean silencing evidence of a genuinely broken config endpoint, which seems like the wrong bar. Happy to narrow the AC wording on the issue if that's the sticking point, but I don't think suppressing the exhausted-retry error is the right fix here.

PR #5131, opened 2026-07-17
feat(ci): add non-blocking shellcheck lint for .github/scripts/*.sh

### LINKED ISSUE #4833: Add shellcheck linting for `.github/scripts/*.sh` in CI
# Add shellcheck linting for `.github/scripts/*.sh` in CI

## What
Add a shellcheck step to the existing CI workflow (likely `.github/workflows/ci.yml`) that lints all `.sh` files under `.github/scripts/`. This should run on every push/PR.

## Why
- **Correctness risk**: Shell scripts in `.github/scripts/` (e.g., the reconciliation script from PR #4827) are exercised in production but have no static analysis. Common bugs (unquoted variables, missing error handling) can silently cause workflow failures.
- **Maintainability**: As more shell scripts are added, manual review is insufficient.
- **Agent confusion**: AI code review tools may miss shell-specific issues that shellcheck catches reliably.

## How
1. Add a job or step in the CI workflow (e.g., `lint-shell`):
   ```yaml
   - name: Run shellcheck
     run: shellcheck .github/scripts/*.sh
   ```
2. Ensure `shellcheck` is available (GitHub-hosted Ubuntu runners include it by default; verify with `which shellcheck` or install via `apt-get` if needed).
3. Pin shellcheck version if desired (e.g., `shellcheck --version` check) to avoid surprise rule changes.

## Constraints
- Must not break existing CI jobs or add new required checks that block PRs without prior agreement (consider `continue-on-error: true` initially, or make it a separate non-blocking job).
- Only lint `.github/scripts/*.sh` — not other shell scripts in the repo (out of scope).
- Do not modify any `.sh` files as part of this issue; only add the linting step.

## LLM tier
**Haiku** — Simple, mechanical task: add a single CI step with a well-known tool. Requires no design judgment or multi-file changes. Cloud execution is appropriate for reliable tool availability.

## Success looks like
- A new CI step runs `shellcheck .github/scripts/*.sh` on every push/PR.
- The step passes (or is allowed to fail without blocking) for the current scripts.
- The step is visible in the CI checks list.

## Failure looks like
- The step is not added, or is added to the wrong workflow file.
- The step fails due to missing `shellcheck` binary (runner incompatibility).
- The step blocks PRs unexpectedly (if `continue-on-error` is not set and a script has warnings).

_Follow-up from AI review of PR #4827._

leonarduk:
## Summary
- Adds a `lint-shell-scripts` job to `.github/workflows/ci.yml` that runs `shellcheck .github/scripts/*.sh` on every push/PR.
- Non-blocking (`continue-on-error: true`) so existing script warnings can't unexpectedly block PRs, per the issue's constraints.
- Scoped only to `.github/scripts/*.sh`; no `.sh` files modified.

Closes #4833

## Test plan
- [x] Validated `ci.yml` parses as valid YAML.
- [x] `shellcheck` is preinstalled on `ubuntu-latest` runners, so no install step is needed.
- [ ] Confirm the new job appears and runs in the PR's CI checks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,17 @@ jobs:
         # respective dependency sets.
         run: .venv-lambda/bin/pytest tests/
 
+  lint-shell-scripts:
+    name: Lint .github/scripts shell scripts (shellcheck)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      # ubuntu-latest ships shellcheck preinstalled; non-blocking (continue-on-error)
+      # until the existing scripts are cleaned up and this can be made required.
+      - name: Run shellcheck
+        continue-on-error: true
+        run: shellcheck .github/scripts/*.sh
+
   lint-workflows:
     name: Lint GitHub Actions workflows
     runs-on: ubuntu-latest


leonarduk:
Re the DeepSeek review's blocking claim on \`actions/checkout@v7\`: this isn't something introduced by this PR — \`ci.yml\` on \`main\` already uses \`actions/checkout@v7\` in all 9 other jobs in this file (\`test\`, \`shell-tests\`, \`powershell-tests\`, \`validate-backend-deps\`, \`lambda-compat\`, \`lint-workflows\`, \`python-compatibility\`, \`detect-frontend-changes\`, \`frontend-smoke\`), and that's the version currently passing CI on main. The new \`lint-shell-scripts\` job just matches the existing repo convention in this exact file, so there's no breaking change here.

Re the "verify shellcheck is available" suggestion: \`ubuntu-latest\` ships shellcheck preinstalled (confirmed via GitHub's runner image docs), and the step is \`continue-on-error: true\`, so this is a reasonable minor follow-up but not a blocker — filing as a possible future enhancement rather than blocking this issue's scope.

No changes made to the PR based on this review since the flagged concern doesn't apply.

PR #5132, opened 2026-07-17
fix: add workflow_dispatch pr_number input to target one PR

### LINKED ISSUE #4834: Add `workflow_dispatch` input to force-reconcile a specific PR number
Here is the complete GitHub issue body:

---

### What

Modify `.github/workflows/sync-changes-requested-label.yml` to accept an optional `workflow_dispatch` input named `pr_number`. When provided, the reconciliation script (`reconcile-changes-requested-label.sh`) should operate on that single PR instead of scanning all open PRs.

### Why

Currently, `workflow_dispatch` triggers the same reconciliation logic as the scheduled run — it iterates over all open PRs. If a single PR is stuck (e.g., label not removed after new commits), there is no way to target it specifically. This forces maintainers to either wait up to 30 minutes for the next scheduled run or trigger a full scan, which is wasteful and slow. A targeted input reduces API calls, speeds up recovery, and avoids unnecessary load on GitHub's API.

### How

1. Add a `workflow_dispatch` input block to the workflow YAML:
   ```yaml
   on:
     workflow_dispatch:
       inputs:
         pr_number:
           description: 'PR number to reconcile (leave empty to reconcile all open PRs)'
           required: false
           type: string
   ```

2. Pass the input as an environment variable or argument to the reconciliation script:
   ```yaml
   - name: Reconcile label
     env:
       PR_NUMBER: ${{ github.event.inputs.pr_number }}
     run: .github/scripts/reconcile-changes-requested-label.sh
   ```

3. In `reconcile-changes-requested-label.sh`, add a guard at the top:
   ```bash
   if [ -n "$PR_NUMBER" ]; then
     # Fetch only that PR's head SHA and labels
     # (use gh pr view "$PR_NUMBER" --json headRefOid,labels)
     # Skip the "list all open PRs" loop entirely
   else
     # Existing logic: iterate all open PRs
   fi
   ```

### Constraints

- Must not break the existing scheduled trigger (no `pr_number` → existing behavior unchanged)
- Must not break the existing `workflow_dispatch` without input (empty input → existing behavior unchanged)
- The script must still exit 0 on success and handle errors gracefully (e.g., invalid PR number, PR not found)
- Out of scope: adding shellcheck linting, adding `gh` response validation (those are separate follow-ups)

### LLM tier

**Sonnet** — This is a multi-file change (workflow YAML + shell script) requiring moderate design judgment. The shell script needs a clean conditional branch that doesn't duplicate the existing reconciliation logic. A smaller model might introduce subtle bugs in shell variable handling or fail to preserve the existing fallback behavior.

### Success looks like

- Running the workflow with `pr_number: 123` reconciles only PR #123
- Running the workflow without `pr_number` reconciles all open PRs (same as before)
- Running the scheduled trigger (no input) reconciles all open PRs (same as before)
- Invalid PR numbers produce a clear error message and exit non-zero

### Failure looks like

- The script errors when `PR_NUMBER` is set but empty (should be treated as "not provided")
- The script still iterates all PRs even when `PR_NUMBER` is provided
- The scheduled trigger breaks because the script now requires an input
- The script fails silently on invalid PR numbers

---

_Follow-up from AI review of PR #4827._

leonarduk:
## Summary
- Adds an optional `pr_number` input to `sync-changes-requested-label.yml`'s `workflow_dispatch` trigger
- When `pr_number` is provided, the fallback job reconciles only that PR via the existing `.github/scripts/reconcile_changes_requested_label.sh` instead of sweeping every open PR
- Leaving `pr_number` empty (or the scheduled/no-input `workflow_dispatch` trigger) preserves the existing full-sweep behavior
- Documents the new input in `docs/AI_REVIEW_WORKFLOWS.md`

Invalid PR numbers are handled by the existing script, which already runs under `set -euo pipefail` and calls `gh pr view` first — an invalid number causes `gh` to error out with a non-zero exit and a clear message, satisfying the issue's "clear error, exit non-zero" requirement without extra validation code.

## Test plan
- [x] `python -c "import yaml; yaml.safe_load(...)"` to validate the workflow YAML parses
- [ ] Manually trigger the workflow with a `pr_number` input on a PR that carries the label and confirm only it is reconciled
- [ ] Manually trigger the workflow with no input and confirm the full sweep still runs

Closes #4834

### DIFF
--- a/.github/workflows/sync-changes-requested-label.yml
+++ b/.github/workflows/sync-changes-requested-label.yml
@@ -111,20 +111,27 @@ jobs:
           REPO: ${{ github.repository }}
           PR_NUMBER_INPUT: ${{ github.event.inputs.pr_number }}
         run: |
-          if [ -n "${PR_NUMBER_INPUT// }" ]; then
+          # xargs collapses/trims surrounding whitespace so a blank or
+          # whitespace-only input is correctly treated as "not provided".
+          PR_NUMBER_INPUT="$(echo "$PR_NUMBER_INPUT" | xargs)"
+
+          if [ -n "$PR_NUMBER_INPUT" ]; then
+            if ! [[ "$PR_NUMBER_INPUT" =~ ^[0-9]+$ ]]; then
+              echo "::error::Invalid pr_number '${PR_NUMBER_INPUT}': must be a positive integer."
+              exit 1
+            fi
             echo "Reconciling single PR #${PR_NUMBER_INPUT} (workflow_dispatch input)."
             bash .github/scripts/reconcile_changes_requested_label.sh "$PR_NUMBER_INPUT"
-            exit 0
+          else
+            PR_NUMBERS=$(gh pr list --repo "$REPO" --state open --label "Changes Requested" \
+              --json number --jq '.[].number')
+
+            if [ -z "$PR_NUMBERS" ]; then
+              echo "No open PRs carry the 'Changes Requested' label."
+              exit 0
+            fi
+
+            for PR_NUMBER in $PR_NUMBERS; do
+              bash .github/scripts/reconcile_changes_requested_label.sh "$PR_NUMBER"
+            done
           fi
-
-          PR_NUMBERS=$(gh pr list --repo "$REPO" --state open --label "Changes Requested" \
-            --json number --jq '.[].number')
-
-          if [ -z "$PR_NUMBERS" ]; then
-            echo "No open PRs carry the 'Changes Requested' label."
-            exit 0
-          fi
-
-          for PR_NUMBER in $PR_NUMBERS; do
-            bash .github/scripts/reconcile_changes_requested_label.sh "$PR_NUMBER"
-          done
--- a/.github/workflows/sync-changes-requested-label.yml
+++ b/.github/workflows/sync-changes-requested-label.yml
@@ -33,7 +33,12 @@ on:
     types: [completed]
   schedule:
     - cron: "*/30 * * * *"
-  workflow_dispatch: {}
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "PR number to reconcile (leave empty to sweep all open PRs)"
+        required: false
+        type: string
 
 # Multiple review workflows (Claude/GPT/DeepSeek) can complete in rapid
 # succession for the same PR, each triggering a workflow_run event that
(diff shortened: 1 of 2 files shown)


leonarduk:
Thanks for the review. Verified each point against actual runtime behavior before responding:

**Fixed (pushed in 94edc0c5):**
- Added explicit `^[0-9]+$` validation on the trimmed input before it reaches `gh`/the script, with a clear `::error::` message and `exit 1` on failure — this directly satisfies the AC and closes the shell-metacharacter injection concern, rather than relying on `gh`'s own error surfacing.
- Restructured to `if/else` so the single-PR and sweep paths are structurally exclusive, removing any reliance on a trailing `exit 0`.
- Replaced the space-stripping check with `xargs`-based trim, so a value like `" 123 "` now resolves to `123` cleanly instead of relying on internal space removal.

**Not changed, with rationale:**
- *"Whitespace-only would be treated as provided"* — verified this is incorrect for the original code too: `[ -n "${PR_NUMBER_INPUT// }" ]` strips all spaces before the `-n` test, so `" "` evaluates to empty and correctly falls through to the sweep. Confirmed with a standalone bash test. The new `xargs` version behaves the same way, just more legibly.
- *"`exit 0` masks script failure"* — GitHub Actions' default `run:` shell for non-Windows runners is `bash --noprofile --norc -eo pipefail {0}` (documented default), so a failing `reconcile_changes_requested_label.sh` call would abort the step via `-e` before ever reaching the trailing `exit 0`. Moot now anyway since the code is restructured to `if/else`.
- *"Guard belongs inside `reconcile-changes-requested-label.sh` per the issue's How section"* — the issue's draft assumed a hyphenated script name and an internal branch that doesn't match the current code. The actual script, `reconcile_changes_requested_label.sh`, already only ever operates on one PR number passed as `$1` — both existing callers (the `workflow_run` job and the sweep loop) call it per-PR. Branching single-vs-sweep in the workflow YAML matches that existing architecture rather than deviating from it.

**Open/acknowledged:**
- No automated test coverage for the new workflow branch — the existing sweep loop in this same job also has none (YAML branching isn't covered by the `bats` suite, which only tests `reconcile_changes_requested_label.sh` directly). Filing this as a possible follow-up rather than blocking on it, consistent with the issue's own out-of-scope list (shellcheck linting, response validation).

PR #5142, opened 2026-07-17
docs(ci): explain test/lambda-compat split in branch protection check

### LINKED ISSUE #4465: Update `scripts/check_branch_protection_required_checks.py` to match new CI job structure

---

### What

Update the file `scripts/check_branch_protection_required_checks.py` to reflect the restructured CI jobs introduced in PR #4464. Specifically:

- The `test` job no longer runs `tests/` (only `cdk/tests/`).
- The `tests/` directory is now covered exclusively by the `lambda-compat` job.
- The `validate-backend-deps` job still exists but was not addressed in the restructuring.

The script currently validates that required CI checks match branch protection rules. It must be updated to reference the correct job names and their associated test paths.

### Why

If branch protection rules are configured to require specific CI checks (e.g., `test`, `lambda-compat`), the script must accurately reflect which jobs cover which paths. Without this update:

- **Correctness risk**: CI could pass but branch protection could block the PR because the script reports mismatched check names.
- **Maintainability**: Future developers relying on this script to validate branch protection will get incorrect results.
- **Agent confusion**: Automated tooling (e.g., bots, CI pipelines) that uses this script to verify configuration will produce false positives/negatives.

### How

1. **Locate the relevant section** in `scripts/check_branch_protection_required_checks.py` where job-to-path mappings are defined (likely a dictionary or list near the top of the file).
2. **Update the mapping**:
   - Remove `tests/` from the `test` job's path list.
   - Ensure `tests/` is listed under the `lambda-compat` job.
   - If `validate-backend-deps` is still present, either keep it (and document why) or remove it if it has been superseded.
3. **Add a comment** explaining the split rationale (similar to the comment added in the CI config).
4. **Update any hardcoded job name references** if job names changed (e.g., if `lambda-compat` replaced a previous job).

### Constraints

- **Do not change** the logic of how branch protection rules are validated — only update the job/path mappings.
- **Do not remove** `validate-backend-deps` unless you confirm it is fully superseded by `lambda-compat` (see PR #4464 review notes: they are not identical).
- **Out of scope**: Adding new CI jobs, modifying branch protection rules themselves, or restructuring the script's validation algorithm.
- **Must not break** existing tests for this script (if any).

### LLM tier

**Sonnet** — This task requires moderate design judgment: understanding the mapping between CI jobs and test paths, deciding whether to keep or remove `validate-backend-deps`, and ensuring the update is consistent with the broader CI restructuring. It is not a simple mechanical rename (local-7b) nor a complex architectural change (Opus).

### Success looks like

- The script correctly reports that `test` covers `cdk/tests/` only.
- The script correctly reports that `lambda-compat` covers `tests/`.
- Running the script against the current branch protection configuration produces no false mismatches.
- All existing tests for the script pass.

### Failure looks like

- The script still lists `tests/` under the `test` job.
- The script does not list `tests/` under `lambda-compat`.
- The script fails or produces incorrect output when run against the actual branch protection rules.
- `validate-backend-deps` is removed without justification, causing a regression in dependency resolution feedback.

---

_Follow-up from AI review of PR #4464._

leonarduk:
## Summary

- Issue #4465 asked to update the job-to-path mapping in `scripts/check_branch_protection_required_checks.py` for the CI restructuring done in #4464.
- Investigation: the script has no such mapping — it only validates required-check *names* against the ruleset and workflow/job names (`EXPECTED_REQUIRED_CHECKS`), and it already lists `CI / test` and `CI / Lambda-compat pytest (backend/requirements.txt)` correctly (confirmed by running the script against current `main`: it reports no mismatches). `validate-backend-deps` is unrelated to `lambda-compat` and stays as-is.
- No functional change was needed. Added the clarifying comment the issue requested, mirroring the rationale already in `ci.yml`'s `lambda-compat` job, so future readers understand why `test` and `lambda-compat` are both required checks and what each covers.

Closes #4465

## Test plan

- [x] `python scripts/check_branch_protection_required_checks.py` passes locally against current `main`

### DIFF
--- a/scripts/check_branch_protection_required_checks.py
+++ b/scripts/check_branch_protection_required_checks.py
@@ -18,6 +18,10 @@ RULESET_PATH = REPO_ROOT / ".github" / "rulesets" / "default-branch-protection.j
 WORKFLOWS_DIR = REPO_ROOT / ".github" / "workflows"
 
 EXPECTED_REQUIRED_CHECKS = {
+    # `test` runs cdk/tests/ (CDK infrastructure) against root deps; the
+    # backend `tests/` suite runs once, under Lambda-pinned deps, in
+    # `lambda-compat` below. Keeping each suite to a single dependency set
+    # avoids running `tests/` twice per PR (see PR #4464).
     "CI / test",
     "CI / Validate backend/requirements.txt (dry-run)",
     "CI / Lambda-compat pytest (backend/requirements.txt)",


leonarduk:
Thanks for the review — I want to push back on the core premise before making further changes, since I don't think there's a job-to-path mapping in this script to update.

`grep -n "tests/\|cdk/tests\|path" scripts/check_branch_protection_required_checks.py` on this branch shows the **only** matches are the comment lines I just added. The script has no dictionary or list mapping jobs to test paths — see the full file (`scripts/check_branch_protection_required_checks.py`):

- `load_ruleset_contexts()` reads required-check *names* (strings like `"CI / test"`) from `.github/rulesets/default-branch-protection.json`.
- `workflow_check_contexts()` dynamically parses `.github/workflows/*.yml` and derives the same kind of check-name strings from the actual `name:`/job `name:` fields — it never reads job `run:` steps or test paths.
- `EXPECTED_REQUIRED_CHECKS` is a flat set of those same name strings, compared against both of the above.

So there's nothing in this file that says "`test` covers `tests/`" or "`lambda-compat` covers `cdk/tests/`" — that mapping only exists as prose in the issue body and as inline comments in `ci.yml` (which already documents the split, added in #4464). Running the script against current `main` (`python scripts/check_branch_protection_required_checks.py`) prints "Branch protection required checks match deterministic workflow contexts." with no mismatches, confirming the check-name set is already correct and in sync with the actual `ci.yml` jobs.

Given that, the AC bullets about "remove `tests/` from the `test` job's path list" / "list `tests/` under `lambda-compat`" don't map onto any code that exists in this file — there's no path list to edit. I added the comment the issue asked for (mirroring `ci.yml`'s existing rationale) since that was the one actionable, safe item. If I'm missing a mapping somewhere else in the repo that this script is supposed to consume, please point me to it and I'll update it.

PR #5148, opened 2026-07-17
[Issue #4480] Make get_changed_files() use the dynamically determined default branch

### LINKED ISSUE #4480: Make `get_changed_files()` use the dynamically determined default branch
# Make `get_changed_files()` use the dynamically determined default branch

## What

In `scripts/pr-script.sh` (or equivalent file), modify the `get_changed_files()` function (lines 168-176) to use the dynamically determined default branch instead of hardcoded `origin/main`.

Currently:
```bash
get_changed_files() {
    git diff --name-only origin/main...HEAD
}
```

Should become:
```bash
get_changed_files() {
    local default_branch
    default_branch=$(get_default_branch)
    git diff --name-only "origin/${default_branch}...HEAD"
}
```

## Why

The script already has a `get_default_branch()` function that may return "master" or another branch name. Hardcoding `origin/main` creates a correctness bug: if the repository's default branch is not "main", the diff comparison will be against the wrong branch, potentially missing changed files or including incorrect ones. This could cause the script to commit/push unintended changes or miss intended ones.

## How

1. Locate the `get_changed_files()` function (around line 168)
2. Replace `origin/main` with a call to `get_default_branch()` prefixed with `origin/`
3. Ensure the function handles the case where `get_default_branch()` returns an empty value (fall back to `main`)

## Constraints

- Do not change the behavior of `get_default_branch()` itself
- Do not modify any other functions or logic
- The fix must be a single, localized change within `get_changed_files()`
- Must remain compatible with all existing callers of `get_changed_files()`

## LLM Tier

**Sonnet** — While the change is small, it requires understanding the relationship between two functions and ensuring the dynamic value is correctly interpolated into the git command. The risk of breaking the git diff syntax (e.g., missing quotes, incorrect variable expansion) warrants a model that can reason about shell variable handling.

## Success Looks Like

- `get_changed_files()` uses the output of `get_default_branch()` instead of hardcoded `main`
- Running the script on a repo with default branch "master" correctly detects changed files
- Running the script on a repo with default branch "main" continues to work identically

## Failure Looks Like

- The git diff command fails due to incorrect branch name interpolation
- `get_changed_files()` returns empty results when `get_default_branch()` returns a non-"main" value
- The script breaks on repos where `get_default_branch()` returns unexpected output (e.g., with newlines)

_Follow-up from AI review of PR #4476._

leonarduk:
## What
`scripts/dev_tools/publish_pr.py::get_changed_files()` (and its caller `stage_and_commit()`) now builds the `git merge-base`/`git diff` comparison ref as `origin/<default_branch>` instead of using the bare branch name directly.

## Why
`main()` already calls `get_default_branch(owner, repo)` to resolve the repo's actual default branch (e.g. `"master"`), but that bare name was being passed straight into `get_changed_files()`, which ran `git merge-base branch <default_branch>` against a **local** ref. The function's own default parameter value (`"origin/main"`) implied a remote-tracking ref, so on any repo whose default branch isn't `main`, or where the local branch is stale/absent, the diff comparison would silently target the wrong ref — missing changed files or including unintended ones.

## How
- `get_changed_files(branch, default_branch="main")` now builds `remote_default_branch = f"origin/{default_branch or 'main'}"` and uses that for `git merge-base` / `git diff`.
- `stage_and_commit()`'s default parameter updated to match (`"main"` instead of `"origin/main"`) since it just forwards to `get_changed_files()`.
- Added `tests/scripts/test_publish_pr.py` covering both the explicit `default_branch="master"` case (asserts `origin/master` is used) and the no-argument case (asserts fallback to `origin/main`).

## Testing
```
python -m pytest tests/scripts/test_publish_pr.py -q --no-cov
```
2 passed.

Closes #4480

### DIFF
--- a/scripts/dev_tools/publish_pr.py
+++ b/scripts/dev_tools/publish_pr.py
@@ -93,9 +93,10 @@ def check_working_tree_clean() -> bool:
         return False
 
 
-def get_changed_files(branch: str, default_branch: str = "origin/main") -> list[str]:
+def get_changed_files(branch: str, default_branch: str = "main") -> list[str]:
     """Get list of changed files: either uncommitted changes or commits on the branch."""
     changed_files = []
+    remote_default_branch = f"origin/{default_branch or 'main'}"
     try:
         # Check for both staged and unstaged changes
         result = subprocess.run(
@@ -110,7 +111,7 @@ def get_changed_files(branch: str, default_branch: str = "origin/main") -> list[
 
         # Check for commits on the branch only if we have a merge base
         result = subprocess.run(
-            ["git", "merge-base", branch, default_branch],
+            ["git", "merge-base", branch, remote_default_branch],
             capture_output=True,
             text=True,
             check=False,
@@ -134,7 +135,9 @@ def get_changed_files(branch: str, default_branch: str = "origin/main") -> list[
     return list(set(changed_files))
 
 
-def stage_and_commit(files: Optional[list[str]], message: str, branch: str, default_branch: str = "origin/main") -> bool:
+def stage_and_commit(
+    files: Optional[list[str]], message: str, branch: str, default_branch: str = "main"
+) -> bool:
     """Stage and commit the specified files (or changed files in branch if none specified)."""
     try:
         if not files:
--- a/tests/scripts/test_publish_pr.py
+++ b/tests/scripts/test_publish_pr.py
@@ -1,0 +1,45 @@
+import scripts.dev_tools.publish_pr as publish_pr
+
+
+class _FakeResult:
+    def __init__(self, returncode=0, stdout=""):
+        self.returncode = returncode
+        self.stdout = stdout
+
+
+def _merge_base_calls(calls):
+    return [c for c in calls if c[:2] == ["git", "merge-base"]]
+
+
+def test_get_changed_files_uses_origin_prefixed_default_branch(monkeypatch):
+    """merge-base/diff must run against origin/<default_branch>, not a bare local ref."""
+    calls = []
+
+    def fake_run(cmd, **kwargs):
+        calls.append(cmd)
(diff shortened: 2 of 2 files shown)


leonarduk:
Thanks for the review. The two blocking findings don't hold up against the full file — they're both based on not having visibility into `main()`, which the diff didn't include context for:

**"`get_default_branch()` is never called"** — it is, in `main()`:
```python
# scripts/dev_tools/publish_pr.py:459
default_branch = get_default_branch(owner, repo)
...
# :473
if stage_and_commit(args.files, commit_msg, branch, default_branch):
```
This was already the case before this PR — `get_default_branch()` being wired into `main()` predates this change. What this PR fixes is that the value flowing from `get_default_branch()` (a bare name like `"master"`) was being passed straight to `git merge-base`/`git diff` as a local ref instead of a remote-tracking ref.

**"`stage_and_commit()` has a broken default parameter / missing `origin/` prefix"** — `stage_and_commit()` never runs a git command against `default_branch` itself. Its only use of the parameter is forwarding it to `get_changed_files()`:
```python
# :145
files = get_changed_files(branch, default_branch)
```
and `get_changed_files()` is exactly where the `origin/` prefixing now happens (line 99: `remote_default_branch = f"origin/{default_branch or 'main'}"`, used at line 114 in `git merge-base`). So the prefix logic applies correctly regardless of whether `stage_and_commit()`'s default (`"main"`) or an explicit value is used.

Given that, I believe this satisfies the issue's actual ask (`get_changed_files()` no longer hardcodes `origin/main` and instead builds the ref from the dynamically-resolved default branch), and I don't think a second `origin/` prefix construction inside `stage_and_commit()` is needed since it has no direct git call to prefix.

Happy to add a test asserting the `main()` → `stage_and_commit()` → `get_changed_files()` wiring if that would help close this out — let me know.

Issue #5035, opened 2026-07-10
Re-scope the PR/branch-automation follow-up cluster: issues point at nonexistent .sh files (real code is Python in scripts/dev_tools/)


leonarduk:
## What

A cluster of AI-suggested follow-ups (#4479, #4480, #4481, #4486, #4488, #4495, #4496) all describe changes to shell scripts that **do not exist** in the repo — `scripts/create-pr.sh`, `scripts/pr-script.sh`, `scripts/create-issue-branch.sh`. They carry the `shell` label and cite bash functions with concrete line numbers that correspond to nothing.

The real implementation is **Python**, in `scripts/dev_tools/`:
- PR publishing → [`scripts/dev_tools/publish_pr.py`](../blob/main/scripts/dev_tools/publish_pr.py)
- Branch creation → [`scripts/dev_tools/work_on_issue.py`](../blob/main/scripts/dev_tools/work_on_issue.py)

As written, an implementer following any of these specs would likely invent a brand-new shell script instead of editing the existing Python. This issue tracks re-scoping (or closing) each child against the real code.

## Why

- **Correctness:** every child issue names a wrong file, wrong language, and fabricated line numbers.
- **Wasted work:** two of the seven are already implemented — an agent would re-do or conflict with existing code.
- **Root cause is now fixed:** `create_followup_issues.py` was emitting unverified paths because it worked from a PR diff, not the repo. The generator hardening in #5030 nudges future issues away from guessing concrete paths, but this batch predates it and needs manual correction.

## Corrected mapping (verified against `main`)

| Issue | Original (phantom) target | Real location | Status |
|---|---|---|---|
| #4479 gh CLI availability check | `create-pr.sh` `main()` L310 | `publish_pr.py` `check_gh_installed()` L360, called in `main()` L416 | **Already done** — only gap is it checks `gh --version`, not `gh auth status`. Close or narrow to the auth check. |
| #4480 dynamic default branch | `pr-script.sh` `get_changed_files()` L168 | `publish_pr.py` `get_changed_files(branch, default_branch="origin/main")` L93; `get_default_branch()` already exists L63 | **Partially done** — the param exists; real fix is to make the **caller** pass `origin/{get_default_branch(...)}` instead of the `origin/main` default. Verify caller, then narrow. |
| #4481 Ollama progress message | `create-pr.sh` L224-256 | `publish_pr.py` `generate_pr_body_with_ollama()` L244 | Valid — re-point. |
[...]

PR #5159, opened 2026-07-17
feat(scripts): --type flag and fetch delay for work_on_issue.py

### LINKED ISSUE #4495: Add small delay before fetching newly created branch to avoid race condition
Here is the complete, actionable GitHub issue body:

---

### What

In `scripts/create-issue-branch.sh`, around line 155, add a short sleep (0.5–1 second) before the `git fetch origin {branch_name}` command that runs immediately after creating a branch via the GitHub API.

Specifically, insert the sleep between the branch creation logic (lines 85–100) and the fetch on line 155.

### Why

When a branch is created via the GitHub API, the API may return success before the new ref is fully propagated and visible to `git fetch`. This creates a race condition where the subsequent `git fetch origin {branch_name}` fails because the remote doesn't yet recognize the branch. While the fallback `git fetch origin` on line 158 catches this, it's wasteful and can cause confusing transient failures in scripts or CI. A small delay makes the behavior deterministic and reliable.

### How

1. Open `scripts/create-issue-branch.sh`
2. Locate the section after branch creation (after line 100, before line 155)
3. Add a single line: `sleep 0.5` (or `sleep 1` for extra safety)
4. Ensure the sleep is placed **before** the `git fetch origin {branch_name}` on line 155

No other changes needed.

### Constraints

- Must not affect the fallback `git fetch origin` on line 158 — that should remain as-is
- Must not introduce any new dependencies (sleep is POSIX, available everywhere)
- Must not change the script's behavior for branches that already exist (the sleep only runs after API creation)
- Out of scope: changing the branch naming scheme, adding `--type` flag, or any other feature work

### LLM tier

**Haiku** — This is a simple, mechanical one-line addition with no ambiguity. It requires no design judgment, no multi-file changes, and no broad repo context. A small local model could also handle it, but Haiku is the safer default for cloud execution.

### Success looks like

- A sleep of 0.5–1 second is added before the targeted `git fetch`
- The script no longer fails on the first `git fetch origin {branch_name}` after creating a branch via API
- All existing tests and manual workflows continue to pass

### Failure looks like

- The sleep is added in the wrong location (e.g., after the fallback fetch)
- The sleep value is excessive (e.g., 10 seconds) or missing entirely
- The change introduces a syntax error or breaks the script on systems without `sleep` (unlikely but worth noting)
- The race condition still occurs because the delay is too short

---

_Follow-up from AI review of PR #4472._

### LINKED ISSUE #4496: Consider adding `--type` flag to allow `feat/` prefix instead of always using `fix/`

---

### What

Currently, the script in `allotmint` (likely `scripts/create-issue-branch.sh` or similar) always creates branches with the `fix/` prefix (e.g., `fix/123-some-title`). The change is to add a `--type` command-line flag that accepts values like `fix` or `feat`, and uses that value as the branch prefix instead of hardcoding `fix/`.

**Location:** The branch name construction logic (around line 107-120 in the reviewed script) and the argument parsing section.

### Why

- **Correctness risk:** Always using `fix/` mislabels feature work, which can confuse CI/CD pipelines, changelog generators, and human reviewers.
- **Maintainability:** Hardcoding the prefix makes the script less reusable across different workflows (e.g., some teams use `feat/`, `chore/`, `docs/`).
- **Agent confusion:** Automated tooling (e.g., release notes bots) often relies on branch prefixes to categorize changes. Incorrect prefixes lead to misclassified releases.

### How

1. **Add argument parsing:** In the argument parsing section, add a `--type` flag with a default value of `fix`. Accept common values: `fix`, `feat`, `chore`, `docs`, `refactor`, `test`.
2. **Validate input:** Reject unknown values with a clear error message (e.g., `Error: --type must be one of: fix, feat, chore, docs, refactor, test`).
3. **Update branch name construction:** Replace the hardcoded `fix/` prefix with `{type}/` where `type` is the value from the `--type` flag.
4. **Update help/usage text:** Add `--type` to the script's help output and README.

### Constraints

- **Must not break existing usage:** The default must remain `fix` so existing invocations (without `--type`) behave identically.
- **Out of scope:** Do not change the issue ID parsing, branch creation API call, or checkout logic. Only the prefix string should change.
- **No new dependencies:** Use simple string interpolation or `case` statements; no external argument parsing libraries.

### LLM Tier

**Sonnet** — This is a moderate design task. It requires:
- Understanding the existing argument parsing pattern (which may use `getopts` or manual parsing)
- Adding validation logic with a clear error message
- Updating both the script and the README
- Ensuring backward compatibility

A smaller model (local-14b) could handle the mechanical change if the argument parsing is already simple and well-documented, but Sonnet is safer to avoid subtle edge cases (e.g., handling `--type=feat` vs `--type feat`).

### Success looks like

- Running `./script.sh --type feat 123` creates branch `feat/123-some-title`
- Running `./script.sh 123` (no `--type`) creates branch `fix/123-some-title` (unchanged behavior)
- Running `./script.sh --type invalid` prints a clear error and exits non-zero
- README documents the new flag with examples

### Failure looks like

- Branch is created with wrong prefix (e.g., `fix/` when `feat/` was requested)
- Script crashes or silently ignores the `--type` flag
- Existing invocations without `--type` break or produce different branch names
- Error message is cryptic or missing

---

_Follow-up from AI review of PR #4472._

leonarduk:
## What
Consolidates two related follow-ups (re-scoped by #5035) that both target `scripts/dev_tools/work_on_issue.py`:

- **#4496** — add a `--type {fix,feat}` CLI flag (default `fix`, preserving current behavior) so branches can be created with a `feat/` prefix instead of always `fix/`.
- **#4495** — sleep 2s after `create_branch()` and before fetching the new branch, to avoid a race where the just-created ref isn't yet visible to `git fetch origin <branch_name>`.

## Why
Per #5035's triage, these issues originally pointed at a phantom `scripts/create-issue-branch.sh`; the real target is `scripts/dev_tools/work_on_issue.py`. Both changes are small and touch the same function (`main()`), so one PR avoids overlapping diffs.

## Testing
- Added `tests/test_work_on_issue.py` (no test file previously existed for this script) covering: default `fix/` prefix, `--type feat` prefix, invalid `--type` rejection, and that the delay is invoked before the branch fetch.
- `python -m pytest tests/test_work_on_issue.py -q` — 4 passed
- `python -m black`, `python -m ruff check` — clean

Closes #4495
Closes #4496

### DIFF
--- a/scripts/dev_tools/work_on_issue.py
+++ b/scripts/dev_tools/work_on_issue.py
@@ -7,6 +7,7 @@ import os
 import re
 import subprocess
 import sys
+import time
 from pathlib import Path
 
 import requests
@@ -82,9 +83,7 @@ def get_main_branch_sha(owner: str, repo: str) -> str:
         sys.exit(1)
 
 
-def create_branch(
-    owner: str, repo: str, branch_name: str, sha: str, token: str | None = None
-) -> None:
+def create_branch(owner: str, repo: str, branch_name: str, sha: str, token: str | None = None) -> None:
     """Create a branch in the remote repo."""
     url = f"https://api.github.com/repos/{owner}/{repo}/git/refs"
     headers = {"Accept": "application/vnd.github.v3+json"}
@@ -112,6 +111,12 @@ def main() -> None:
         "--token",
         help="GitHub personal access token (for creating branches)",
         default=None,
+    )
+    parser.add_argument(
+        "--type",
+        choices=["fix", "feat"],
+        default="fix",
+        help="Branch prefix to use (default: fix)",
     )
     args = parser.parse_args()
 
@@ -143,7 +148,7 @@ def main() -> None:
 
     # Create branch name
     slug = slugify(title)
-    branch_name = f"fix/issue-{args.issue_id}-{slug}"
+    branch_name = f"{args.type}/issue-{args.issue_id}-{slug}"
     print(f"Branch name: {branch_name}")
 
     # Get the current main/master branch SHA
@@ -153,6 +158,9 @@ def main() -> None:
     # Create branch in remote
     print("Creating branch in remote...")
     create_branch(owner, repo, branch_name, sha, args.token or os.getenv("GITHUB_TOKEN"))
+
+    # Small delay to avoid a race where the branch ref isn't visible yet
+    time.sleep(2)
 
     # Fetch the newly created branch so the local checkout can reference it
     try:
--- a/scripts/dev_tools/work_on_issue.py
+++ b/scripts/dev_tools/work_on_issue.py
@@ -160,7 +160,7 @@ def main() -> None:
     create_branch(owner, repo, branch_name, sha, args.token or os.getenv("GITHUB_TOKEN"))
 
     # Small delay to avoid a race where the branch ref isn't visible yet
-    time.sleep(2)
+    time.sleep(1)
 
     # Fetch the newly created branch so the local checkout can reference it
     try:
--- a/tests/test_work_on_issue.py
+++ b/tests/test_work_on_issue.py
@@ -1,0 +1,68 @@
+"""Unit tests for work_on_issue script."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from unittest import mock
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "dev_tools"))
+from work_on_issue import main
+
+
+def _run_main(monkeypatch, tmp_path, cli_args, sleep_mock):
+    """Run main() with every external side effect mocked, return the resolved branch name."""
+    monkeypatch.chdir(tmp_path)
+    monkeypatch.setattr(sys, "argv", ["work_on_issue.py", *cli_args])
+    monkeypatch.setattr("work_on_issue.time.sleep", sleep_mock)
+    monkeypatch.setattr("work_on_issue.get_repo_info", lambda: ("leonarduk", "allotmint"))
+    monkeypatch.setattr(
+        "work_on_issue.fetch_issue",
+        lambda owner, repo, issue_id: {"title": "Some Issue Title", "body": "body text"},
+    )
+    monkeypatch.setattr("work_on_issue.get_main_branch_sha", lambda owner, repo: "deadbeef")
+    monkeypatch.setattr("work_on_issue.create_branch", lambda owner, repo, branch_name, sha, token: None)
+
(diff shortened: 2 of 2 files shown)


leonarduk:
## PR Merge Checker — Summary

### Merge readiness
| Gate | Status | Notes |
|------|--------|-------|
| Issue linkage | ✅ | `Closes #4495` / `Closes #4496` in PR body |
| Issue coverage | ✅ | Both issues' AC met (see below) |
| Review comments | ✅ | No inline review threads open |
| CI | ⚠️ | See below — one check is stale/invalid, one legitimate point fixed |
| Logic correctness | ✅ | Diff is correct and minimal |

### On the "DeepSeek PR Review" failing check

That review claims the PR modifies the "wrong file" (`work_on_issue.py` instead of `scripts/create-issue-branch.sh`) and that `--type` is out of scope. **This is based on the original, unedited issue bodies of #4495/#4496, which #5035 already identified as pointing at a phantom file that doesn't exist in this repo** (`scripts/create-issue-branch.sh` returns 404 on `main`). #5035's corrected mapping explicitly re-points both issues at `work_on_issue.py`:
- #4495 → `work_on_issue.py` branch fetch (was L161, now the delay line)
- #4496 → `work_on_issue.py`'s hardcoded `fix/issue-{id}-{slug}` (was L150)

So the PR is targeting the correct file per the accepted re-scoping; the review's core objection doesn't hold. (#5035 also calls for updating #4495/#4496's bodies to reflect the corrected mapping — that's still outstanding as separate triage work, not a blocker for this PR.)

### Actions taken
- Fixed the one legitimate point from the review: `time.sleep(2)` exceeded #4495's spec'd 0.5–1s delay. Reduced to `time.sleep(1)` (the issue's own "extra safety" ceiling) in commit [`9e2ba56`](https://github.com/leonarduk/allotmint/commit/9e2ba562a7c9499907e120f59ea0e4d8f6b8b3b7).
- No changes needed for the `--type` flag — it's in scope per #5035's mapping, not out of scope as the stale review claims.

### Remaining blockers
- "Backend Integration Tests" was still in progress at review time — confirm it passes on the latest commit before merging.
- Consider re-running/dismissing the DeepSeek check now that the file-target objection is addressed, or manually overriding it given the explanation above.
- Optional housekeeping (not a blocker): update #4495/#4496 issue bodies per #5035's item 3, since this PR relies on the re-scoped mapping rather than the issues' literal text.

### Verdict
**READY TO MERGE** once Backend Integration Tests complete successfully on the latest commit — no code changes remain outstanding.

PR #5166, opened 2026-07-17
docs(ci): explain why mypy needs --explicit-package-bases

### LINKED ISSUE #4842: Revert CI workflow change or document its purpose in a separate PR
Here is the complete, actionable GitHub issue body in Markdown format.

---

**Issue Title:** Revert CI workflow change or document its purpose in a separate PR

**What**
- **File:** `.github/workflows/backend-integration.yml`
- **Change:** Remove the addition of `--explicit-package-bases` from the mypy invocation (line added in PR #4840).
- **Alternative:** If the flag is intentionally needed, revert it in this PR and open a separate PR that includes both the flag change and its justification.

**Why**
- The change is unrelated to the stated goal of PR #4840 (adding a type hint to `serialise_config()`).
- Adding `--explicit-package-bases` alters mypy’s import resolution behavior, which can mask real type errors that were previously caught.
- This creates a correctness risk: future type errors may go undetected in CI, leading to subtle bugs in production.
- It violates the principle of minimal, focused PRs, making the diff harder to review and maintain.

**How**
1. Open `.github/workflows/backend-integration.yml`.
2. Locate the mypy invocation line that was modified in PR #4840.
3. Remove the `--explicit-package-bases` flag, restoring the original mypy command.
4. If the flag is genuinely needed, revert it here and create a new PR that adds it with a clear rationale (e.g., "Fix mypy import resolution for new package structure").

**Constraints**
- Do **not** change any other part of the CI workflow or the Python source code.
- The core type hint change from PR #4840 must remain untouched.
- The unrelated `__init__.py` file at the repo root (also flagged in the review) should be handled in a separate issue.

**LLM tier**
- **Haiku** — The task is simple and mechanical (reverting a single flag in a YAML file), but it requires awareness of the broader CI context and the ability to verify the change doesn’t break other workflows. A local model may lack the necessary repo context or tool access to confirm correctness.

**Success looks like**
- The mypy invocation in `backend-integration.yml` no longer contains `--explicit-package-bases`.
- CI passes with the same type-checking behavior as before PR #4840.
- No other files are modified.

**Failure looks like**
- The flag is only partially removed (e.g., typo or leftover comment).
- The CI workflow breaks due to an unrelated change made in the same edit.
- The core type hint change from PR #4840 is accidentally reverted or altered.

_Follow-up from AI review of PR #4840._

leonarduk:
## What
Add a comment above the `Type-check backend` step in `.github/workflows/backend-integration.yml` explaining why `--explicit-package-bases` is needed on the mypy invocation.

## Why
#4842 asked to either revert the flag (added in #4840 alongside a new root-level `__init__.py`, with no explanation) or document it. I reproduced the failure mode locally: mypy walks up from a root `__init__.py` to infer a module name from the containing directory, and fails with `"<dir> contains __init__.py but is not a valid Python package name"` whenever that directory name isn't a valid identifier — e.g. any worktree path containing a hyphen, which this repo uses constantly (`allotmint-issue-NNNN`). `--explicit-package-bases` avoids that walk-up entirely.

Reverting would make CI fragile to checkout path naming; the flag is load-bearing, so documenting in place is the right call per #4842's own "alternative" option.

## Testing
- Confirmed the failure by running `python -m mypy backend --config-file mypy.ini` (no flag) from inside a hyphenated worktree path — reproduces the exact error the comment describes.
- `python -m mypy backend --config-file mypy.ini --show-error-codes --pretty --explicit-package-bases` — passes as before (no functional change, comment only)
- `actionlint` — clean

Closes #4842

### DIFF
--- a/.github/workflows/backend-integration.yml
+++ b/.github/workflows/backend-integration.yml
@@ -56,6 +56,13 @@ jobs:
           aws s3 sync "s3://$DATA_BUCKET/" data/
 
       - name: Type-check backend
+        # --explicit-package-bases is required because of the root-level __init__.py
+        # (added in #4840 so `backend.contracts_spa` imports resolve from the repo
+        # root). Without it, mypy walks up from that __init__.py to infer a module
+        # name from the checkout directory itself, which fails whenever that
+        # directory name isn't a valid Python identifier (e.g. a worktree path
+        # like allotmint-issue-4842 or any path containing a hyphen) with
+        # "contains __init__.py but is not a valid Python package name".
         run: |
           test -f mypy.ini
           python -m mypy --version


leonarduk:
Re: the DeepSeek REQUEST CHANGES review — responding on the substance rather than just the letter of #4842.

#4842 asked for one of two things: remove `--explicit-package-bases` outright, or revert it here and re-add it in a separate PR with justification. This PR takes a third option — keep the flag, document it in place — because I reproduced the actual failure mode first:

```
python -m mypy backend --config-file mypy.ini   # no flag, run from a hyphenated worktree path
```
fails with `"<dir> contains __init__.py but is not a valid Python package name"`. The repo's own worktree convention (`allotmint-issue-NNNN`) hits this every time, since directory names with hyphens aren't valid Python identifiers. Removing the flag would make CI fail depending on the checkout path name — not a safe revert.

#4842 was filed as an automated follow-up to #4840's review without that repro, so its premise ("remove the flag, restore original behavior") doesn't hold once you account for the worktree naming issue. Documenting the flag in place, closing #4842, is the more correct fix than blindly following its literal steps. Happy to discuss further if there's a scenario I'm missing.

PR #5207, opened 2026-07-18
fix: allow Value to take precedence when Units+Price are also filled

### LINKED ISSUE #5194: Input form rejects submission when Value, Units, and Price are all filled
Summary: On the Account + Holdings Input page (/input), saving a holding fails validation if the Value (GBP), Units, and Price (GBP) fields are all filled in at the same time.

Steps to reproduce: Go to /input. Fill Owner, Account, and Ticker as normal. Fill in Value (GBP), Units, and Price (GBP) all at once (e.g. Value=500, Units=5, Price=100). Click "Save holding".

Expected: The holding saves successfully, using Value as an override or simply ignoring Units/Price if Value is also present.

Actual: The form shows the message "Provide either Value (GBP) or both Units + Price (GBP)" and blocks the save. Clearing the Value field (leaving only Units and Price) allows the save to succeed with "Holding saved."

This may be intentional exclusive-or validation, but the error message and UX are confusing since all three fields accept input simultaneously without any inline warning until submission is attempted. Consider either disabling/graying out Value when Units+Price are populated (and vice versa), or allowing all three fields to be filled with Value taking precedence.

Tested on release v0.28.1 at the deployed site.

leonarduk:
## Summary
- Fixes the manual holding form/API rejecting saves when Value, Units, and Price are all filled in at once.
- Value now takes precedence over Units+Price when both are provided; the error only fires when neither a Value nor a complete Units+Price pair is present.
- Applied the same fix to both the frontend validation (`TransactionsPage.tsx`) and the backend validation (`backend/routes/transactions.py`) — the backend handler already prioritized `value_gbp` when saving, so only the guard needed to change.

## Test plan
- [x] `python -m pytest tests/test_transactions_route.py -q` (39 passed)
- [x] `npm --prefix frontend run test -- --run TransactionsPage` (12 passed)
- [x] `npm --prefix frontend run lint` (clean)

Closes #5194

### DIFF
--- a/backend/routes/transactions.py
+++ b/backend/routes/transactions.py
@@ -275,7 +275,7 @@ def _validate_manual_holding_payload(payload: ManualHoldingCreate) -> None:
     has_price = payload.price_gbp is not None
     has_units_price = has_units and has_price
     has_partial_units_price = has_units != has_price
-    if has_partial_units_price or has_value == has_units_price:
+    if has_partial_units_price or not (has_value or has_units_price):
         raise HTTPException(
             status_code=400,
             detail="Provide either value_gbp or both units and price_gbp",
--- a/frontend/src/components/TransactionsPage.tsx
+++ b/frontend/src/components/TransactionsPage.tsx
@@ -245,7 +245,7 @@ export function TransactionsPage({ owners, inputOnly = false }: Props) {
 
     const hasValue = hasValueInput;
     const hasUnitsPrice = hasUnitsInput && hasPriceInput;
-    if (hasValue === hasUnitsPrice) {
+    if (!hasValue && !hasUnitsPrice) {
       setManualError('Provide either Value (GBP) or both Units + Price (GBP).');
       return;
     }
--- a/frontend/tests/unit/components/TransactionsPage.test.tsx
+++ b/frontend/tests/unit/components/TransactionsPage.test.tsx
@@ -301,6 +301,39 @@ describe('TransactionsPage', () => {
     expect(createManualHoldingMock).not.toHaveBeenCalled();
   });
 
+  it('saves using Value when Value, Units, and Price are all provided', async () => {
+    const user = userEvent.setup();
+    render(
+      <TransactionsPage
+        owners={[
+          { owner: 'alex', full_name: 'Alex Example', accounts: ['isa'] },
+        ]}
+        inputOnly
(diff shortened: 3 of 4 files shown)


leonarduk:
Thanks for the review. Responding to each point:

**Section 2 (claimed regression):** This is a misread of the boolean logic, not a bug. The condition is:

```python
if has_partial_units_price or not (has_value or has_units_price):
    raise HTTPException(...)
```

`or` here is a disjunction on the *raise* condition, not a fallthrough — if `has_partial_units_price` is `True`, the whole `if` is `True` and the exception is raised regardless of the second operand. Short-circuit evaluation just means the second operand isn't *evaluated*; it does not mean the branch is skipped. For `{units: 5, price_gbp: null}`: `has_partial_units_price = True` → `True or X` → `True` → still rejected. I added two explicit test cases (`units` only, `price_gbp` only, both with no `value_gbp`) to `test_create_manual_holding_rejects_invalid_metric_combo` in 841242a4 to make this regression-proof and directly verifiable.

**Section 3 (payload not stripped):** Already covered — the backend never persists both. See `backend/routes/transactions.py:730-734`: `if payload.value_gbp is not None: holding["value_gbp"] = ...` else units/price, so `units`/`price_gbp` are dropped whenever `value_gbp` is present, independent of what the frontend sends. The existing test `test_create_manual_holding_with_value_takes_precedence_over_units_and_price` asserts this directly: `resp.json()["holding"] == {"ticker": "MSFT", "value_gbp": 500.0}` for a payload that includes all three fields — no `units`/`price_gbp` in the stored holding. The frontend test's payload-shape assertion is testing the frontend's own request body (it already omits the fields via the pre-existing `hasValue ? {...} : {...}` ternary, unchanged by this PR), not a backend guarantee — but the backend test above confirms the backend enforces the same outcome independently.

**Section 4 (coverage gaps):** Added the units-only/price-only-without-value cases above. Value-only and units+price-only paths were already covered by pre-existing tests (`test_create_manual_holding_with_value_persists_account_file`, `test_create_manual_holding_with_units_and_price`). Not adding a `value=0` case since `value_gbp: float | None = Field(default=None, gt=0)` already rejects `0` at the pydantic layer (422, not the 400 path this PR touches) — out of scope for this fix.

Pushed as 841242a4.

PR #5237, opened 2026-07-18
fix: close signup owner-slug race window between mkdir and person.json write

### LINKED ISSUE #4402: Consider adding directory-level lock for `_resolve_owner_slug` to prevent race conditions on concurrent approvals for the same email
# Consider adding directory-level lock for `_resolve_owner_slug` to prevent race conditions on concurrent approvals for the same email

## What

Add a directory-level locking mechanism in `_resolve_owner_slug` (file: `signup_provision.py`, lines 60-78) to prevent two concurrent approval requests for the same email from creating duplicate owner directories.

## Why

The current implementation uses `mkdir(exist_ok=False)` to atomically claim a slug, which is correct for preventing slug collisions. However, there's a window between `mkdir` succeeding and `_write_person_identity` writing the `person.json` file where a second concurrent approval for the same email could also succeed on `mkdir` (since the first hasn't written `person.json` yet). The `_read_person_email` check on lines 72-73 only mitigates this if the first call has already written `person.json`. This is a theoretical race condition that could create orphan directories and inconsistent state.

## How

1. After `mkdir` succeeds, create a lock file (e.g., `person.json.lock`) inside the new directory using `os.open` with `O_CREAT | O_EXCL` for atomic creation
2. Before writing `person.json`, acquire the lock (or verify the lock file is ours)
3. Release the lock after `_write_person_identity` completes
4. Alternatively, use `tempfile.mkstemp` to write `person.json` atomically and then rename it into place

**Simpler approach:** Write `person.json` immediately after `mkdir` succeeds (before releasing the directory), making the email check on line 72-73 effective for concurrent callers.

## Constraints

- Must not change the public API of `_resolve_owner_slug`
- Must not introduce deadlocks or performance regressions
- Must remain compatible with existing error handling (OSError, JSONDecodeError)
- Out of scope: changing the overall provisioning workflow, adding distributed locks, or modifying `ensure_owner_scaffold`

## LLM tier

**Sonnet** — This requires moderate design judgment to choose the right locking strategy, understand the existing concurrency model, and implement without introducing new bugs. The change spans multiple functions and needs careful error handling.

## Success looks like

- Two concurrent approvals for the same email never produce duplicate directories
- Existing tests continue to pass without modification
- No new deadlock or timeout scenarios introduced
- The lock is reliably cleaned up on success or failure

## Failure looks like

- Deadlock under concurrent access
- Lock files left behind after crashes (orphan locks)
- Performance regression from lock contention
- Tests fail intermittently due to timing issues

_Follow-up from AI review of PR #4382._

leonarduk:
## Summary
- `_resolve_owner_slug` claims a slug atomically via `mkdir`, but previously only wrote `email` into `person.json` much later, inside `provision_owner`.
- A second concurrent approval for the *same* email that lost the `mkdir` race would hit `FileExistsError`, find no `person.json` yet, fail the `_read_person_email` match, and allocate a second slug for the same person — an orphan account.
- Fix: scaffold and stamp `person.json` with the email immediately after `mkdir` succeeds, closing the race window at the source instead of leaving it to a later non-atomic write.

## Test plan
- [x] Added `test_resolve_owner_slug_stamps_email_immediately_after_mkdir`, which simulates the losing side of the race by calling `_resolve_owner_slug` twice directly.
- [x] Verified the new test fails on the pre-fix code (`jane-2` instead of reusing `jane`) and passes after the fix.
- [x] `pytest tests/backend/test_signup_provision.py tests/backend/routes/test_signup.py` — 39 passed.
- [x] `black --check` / `ruff check` on the changed file — clean.

Closes #4402

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate owner directories from being created when concurrent sign-up requests use the same email address.
  * Ensured owner details are recorded immediately during provisioning, allowing subsequent requests to reuse the existing owner identity.

* **Tests**
  * Added regression coverage for concurrent owner provisioning scenarios.

### DIFF
--- a/backend/common/signup_provision.py
+++ b/backend/common/signup_provision.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 import json
 import logging
 import re
+import shutil
 from pathlib import Path
 
 from backend.common.compliance import ensure_owner_scaffold
@@ -78,7 +79,7 @@ def _read_person_email(owner_dir: Path) -> str:
     return email.strip().lower() if isinstance(email, str) else ""
 
 
-def _resolve_owner_slug(email: str, accounts_root: Path) -> str:
+def _resolve_owner_slug(email: str, accounts_root: Path, full_name: str = "") -> str:
     """Atomically claim an owner slug for ``email`` under ``accounts_root``.
 
     The candidate directory is created with ``mkdir`` (no ``exist_ok``) so two
@@ -88,13 +89,17 @@ def _resolve_owner_slug(email: str, accounts_root: Path) -> str:
     another user's data. ``email`` is lowercased to match the normalised form
     stored in ``person.json``.
 
-    ``person.json`` is scaffolded and stamped with ``email`` immediately after
-    ``mkdir`` wins, before returning. Without this, a second concurrent
-    approval for the *same* email that loses the ``mkdir`` race would hit
-    ``FileExistsError`` and find no ``person.json`` yet (the winner hadn't
-    written it), read no matching email via ``_read_person_email``, and go on
-    to claim a second slug for the same person. Writing it here closes that
-    window instead of leaving it to the caller's later, non-atomic write.
+    ``person.json`` is scaffolded and stamped with the full identity
+    (``email`` + ``full_name``) immediately after ``mkdir`` wins, before
+    returning. Without this, a second concurrent approval for the *same* email
+    that loses the ``mkdir`` race would hit ``FileExistsError`` and find no
+    ``person.json`` yet (the winner hadn't written it), read no matching email
+    via ``_read_person_email``, and go on to claim a second slug for the same
+    person. Writing it here closes that window instead of leaving it to the
+    caller's later, non-atomic write.
+
+    If the scaffold or identity write fails, the created directory is removed
+    so we never leave an orphan directory behind.
     """
 
     email = email.strip().lower()
@@ -109,8 +114,12 @@ def _resolve_owner_slug(email: str, accounts_root: Path) -> str:
             if _read_person_email(owner_dir) == email:
                 return candidate
             continue
-        ensure_owner_scaffold(candidate, accounts_root)
-        _write_person_identity(owner_dir, email, "")
+        try:
+            ensure_owner_scaffold(candidate, accounts_root)
+            _write_person_identity(owner_dir, email, full_name)
+        except Exception:
+            shutil.rmtree(owner_dir, ignore_errors=True)
+            raise
         return candidate
     raise RuntimeError(f"could not allocate an owner slug for base {base!r}")
 
@@ -154,7 +163,7 @@ def provision_owner(
 
     accounts_root = Path(accounts_root)
     email = record.email.strip().lower()
-    owner = _resolve_owner_slug(email, accounts_root)
+    owner = _resolve_owner_slug(email, accounts_root, record.name)
     owner_dir = ensure_owner_scaffold(owner, accounts_root)
     _write_person_identity(owner_dir, email, record.name)
 
--- a/backend/common/signup_provision.py
+++ b/backend/common/signup_provision.py
@@ -87,6 +87,14 @@ def _resolve_owner_slug(email: str, accounts_root: Path) -> str:
     re-provision); one owned by a different email is skipped so we never clobber
     another user's data. ``email`` is lowercased to match the normalised form
     stored in ``person.json``.
+
+    ``person.json`` is scaffolded and stamped with ``email`` immediately after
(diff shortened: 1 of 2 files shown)


leonarduk:
Addressed the DeepSeek review in 1d53c27a.

**Fixed — orphan directory on stamp failure.** If `_write_person_identity` raised after `mkdir` had already claimed the slug in `_resolve_owner_slug`, the directory was left behind with no `person.json`. Since `_read_person_email` would never match an empty/missing `person.json` to the email, every future call would skip that slug and burn through candidates forever. The scaffold+stamp step is now wrapped in try/except: on failure the half-scaffolded directory is removed (`shutil.rmtree`) and the exception re-raised, so a retry reclaims the same slug. Added a regression test (`test_resolve_owner_slug_cleans_up_directory_when_stamping_fails`) that forces the write to fail and asserts both the cleanup and a successful retry. This also satisfies #4402's stated success criterion — "the lock is reliably cleaned up on success or failure."

**Not changing — the "blank display_name window" point.** I checked this against `compliance.py`'s `_ensure_owner_scaffold`, which is `ensure_owner_scaffold`'s underlying default: a freshly scaffolded `person.json` already has `full_name: ""`. Passing `""` from `_resolve_owner_slug` into `_write_person_identity` doesn't blank anything that wasn't already blank — `_write_person_identity` only overwrites `full_name` when the incoming value is truthy (`if full_name and not str(loaded.get("full_name") or "").strip()`), so it's a no-op here. And the idempotent-reuse branch (existing directory, matching email) returns immediately without touching `person.json` at all, so an already-populated `full_name` is never at risk. There's no window where a previously-set display name gets blanked.

**Not changing — the "missing lock file" AC critique.** #4402 explicitly names the write-immediately-after-`mkdir` approach taken here as the endorsed "simpler approach," and its constraints state the fix "must not change the public API of `_resolve_owner_slug`" and lists "adding distributed locks" as out of scope. A lock-file mechanism isn't warranted here.

Also flagging: the review bot is listed as "Advisory only" in its own comment footer, but the PR was auto-labeled "Changes Requested" off it — worth checking that label logic doesn't block merges on advisory-only findings that don't apply.

leonarduk:
Re: the latest DeepSeek review's "critical bug" (double `ensure_owner_scaffold`/`_write_person_identity` calls) — I don't think this is a real bug, and its suggested fix (tracked in #5258) would actually cause a regression:

- `ensure_owner_scaffold` is idempotent by design: `_ensure_owner_scaffold` in `compliance.py` does `owner_dir.mkdir(parents=True, exist_ok=True)` and skips writing each default file `if path.exists()`. Calling it a second time cannot overwrite anything `_resolve_owner_slug` already wrote.
- The second `_write_person_identity` call in `provision_owner` is **not redundant** — it's the only place `record.name` (the real display name) gets written. `_resolve_owner_slug`'s call passes `full_name=""`, and `_write_person_identity` only fills `full_name` when the incoming value is truthy, so that first call is a no-op for that field. Removing `provision_owner`'s call, as #5258 proposes, would mean `full_name` is never populated at all — that would break the existing test `test_provision_owner_scaffolds_and_writes_email`, which asserts `person["full_name"] == "Jane Doe"`.
- This double-call shape isn't new from my latest push either — it was already present in the original commit (35e2fd7d).

Closing #5258 with this explanation. #5259 (a targeted test for `ensure_owner_scaffold` failure cleanup) looks like a genuinely small, valid gap — my try/except already wraps both calls together so that failure path is covered, but there's no test isolating it specifically. I'll pick that one up.

leonarduk:
Correction to my earlier comments on this thread: I was working against a stale copy of this branch and didn't notice the PR had already been merged (by @leonarduk, commit `d7a1f8b3` → merge commit `b5cd3b89`, merged 2026-07-18 09:45 UTC) before I pushed my follow-up commits (`1d53c27a`, `52c6b984`). Those two commits landed on the now-orphaned branch, not on `main`, and were never part of this PR's merged history.

The version that actually merged took a different approach than the one I reasoned about: it passes `full_name` through to `_resolve_owner_slug` (`full_name: str = ""` param) instead of leaving it blank, so `person.json` gets the real display name immediately on the winning `mkdir`. That means my "the double `ensure_owner_scaffold`/`_write_person_identity` call isn't a real bug" analysis was wrong for the code that actually shipped — the second call in `provision_owner` is genuinely redundant against `main` as it stands now. I've reopened #5258 with the corrected analysis, and I'll follow up with a fresh PR against `main` for the `ensure_owner_scaffold`-failure test coverage from #5259, since that gap is still real and this PR is closed.

Sorry for the noise here.

Issue #4930, opened 2026-07-08
Backend Lambda times out on every request: unbounded live Yahoo Finance fetches during startup warmup


leonarduk:
## What

The deployed `BackendLambdaStack` backend Lambda times out (30s) on **every** invocation, including trivial routes like `GET /health` and `GET /config`. API Gateway returns `503 Service Unavailable` for both the real request and the CORS preflight `OPTIONS` — which the browser reports as a CORS policy block, but the actual cause is unrelated to CORS.

**Root cause chain (corrected — see note below):**

1. On lifespan startup, `AppLifecycleService._warm_snapshot()` (`backend/bootstrap/startup.py:79`) calls `await asyncio.to_thread(instrument_api.prime_latest_prices)` (`backend/bootstrap/startup.py:80`). Offloading to a thread does **not** stop the awaiting lifespan coroutine from blocking on it — startup still can't finish until this call returns.
2. `prime_latest_prices` (`backend/common/instrument_api.py:186`) calls `load_latest_prices(_ALL_TICKERS)` — every known instrument ticker, unconditionally, on every cold start — which calls `load_meta_timeseries_range` → `_memoized_range` (`backend/timeseries/cache.py:454`) → `load_meta_timeseries` (`backend/timeseries/cache.py:386`). For any ticker without a cached parquet, this is a **live synchronous full-year price-history fetch from Yahoo Finance**, one ticker at a time.
3. Separately, inside that same range-load path, `_convert_to_base_currency` (`backend/timeseries/cache.py:474`) calls `get_instrument_meta(f"{ticker}.{exchange}")` to resolve the instrument's currency. For any ticker with no local JSON metadata file, this falls through to `_auto_create_instrument_meta` (`backend/common/instruments.py:415`), which performs a **second, independent live HTTP fetch to Yahoo Finance** (this time for metadata) and then calls `save_instrument_meta`.
4. `save_instrument_meta` (`backend/common/instruments.py:194`) always attempts a local filesystem write first (`path.open("w", ...)` at line 220). In Lambda, `/var/task` is the deployment package and is **read-only**, so every write fails with `OSError: [Errno 30] Read-only file system`. This is caught in `_auto_create_instrument_meta` (logged as "Failed to persist auto-created metadata...") and does not itself crash the request — but it's evidence of, not the primary cause of, the timeout.
[...]

PR #5318, opened 2026-07-19
test: add METADATA_PREFIX cross-check test, mirroring the WRITABLE_ACCOUNTS_PREFIX guard

### LINKED ISSUE #4933: Consider adding a similar cross-check test for any other duplicated configuration constants in the codebase

---

### What
Identify all configuration constants (string literals, default values, fallback values) that are duplicated across the codebase (e.g., in `backend` modules and `cdk`/infrastructure code). For each such constant, add a cross-check test (similar to `test_writable_accounts_prefix_matches_backend_accounts_store`) that uses `ast` parsing to verify the two copies remain in sync.

### Why
- **Correctness risk:** Duplicated constants can silently diverge, leading to hard-to-debug runtime mismatches between backend logic and infrastructure configuration.
- **Maintainability:** Without cross-checks, developers must manually remember to update both copies — a common source of bugs.
- **Agent confusion:** AI code generation tools may inadvertently update only one copy, introducing subtle breakage.

### How
1. **Audit the codebase** for duplicated configuration constants. Likely candidates include:
   - Environment variable names and their fallback values
   - Prefix strings (e.g., `"writable-accounts"`)
   - Default values for feature flags or resource names
   - Any string literal that appears in both `backend/` and `cdk/` (or `infrastructure/`)
2. **For each candidate**, write a test that:
   - Uses `ast.parse()` to extract the fallback literal from the backend source file (as done in the existing test)
   - Compares it to the literal used in the CDK/infrastructure code
   - Fails if they differ
3. **Place tests** in the same test file as the existing cross-check test, or in a dedicated `test_config_consistency.py` file.

### Constraints
- **No new runtime dependencies** on backend modules — use `ast` parsing only
- **No runtime behavior changes** — tests only, no production code modifications
- **No circular imports** introduced
- **Out of scope:** Refactoring the constants into a shared module (that would be a separate, larger effort)

### LLM tier
**Sonnet** — This task requires moderate design judgment: identifying which constants are truly duplicated (not just coincidentally equal), deciding which ones warrant a cross-check, and following the existing `ast`-based pattern correctly. It is not purely mechanical (local-7b/14b) and does not require complex architecture (Opus).

### Success looks like
- A new test (or set of tests) that verifies at least 2–3 additional duplicated constants are in sync
- All existing tests continue to pass
- The new tests fail if a constant is intentionally changed in only one location

### Failure looks like
- The tests pass even when constants are out of sync (false negative)
- The tests introduce a runtime dependency on backend modules
- The tests are brittle (e.g., break on minor formatting changes unrelated to the constant value)

---

_Follow-up from AI review of PR #4927._

leonarduk:
## Summary
Follows the existing `test_writable_accounts_prefix_matches_backend_accounts_store` pattern (issue #4323): a new `test_metadata_prefix_matches_backend_instruments_s3_location` statically cross-checks `cdk/stacks/backend_lambda_stack.py`'s `METADATA_PREFIX = "instruments"` against the fallback literal in `backend/common/instruments.py`'s `_s3_location()` (`os.getenv(METADATA_PREFIX_ENV, "instruments")`), reusing the existing `_fallback_string_literal()` ast helper.

If these two literals drift, CDK would set `METADATA_PREFIX` to one value in the Lambda's environment while the backend's own fallback (used only if the env var is ever unset) would silently point somewhere else — see issue #4930 for why this matters for auto-created instrument metadata storage.

### Audit scope
I searched `cdk/stacks/*.py` for every module-level string constant and checked each against the backend for a matching fallback literal. Found only two: `WRITABLE_ACCOUNTS_PREFIX` (already covered) and `METADATA_PREFIX` (added here). No other duplicated configuration constants exist between `cdk/` and `backend/` at this time — the issue's suggested candidates (env var names, other prefix strings) either aren't duplicated as literals anywhere else, or are already single-sourced via import (e.g. `BACKEND_API_URL_EXPORT`).

## Test plan
- [x] Manually verified the new test fails when `METADATA_PREFIX` is changed to a different string in the CDK stack (confirms it actually catches drift), then reverted
- [x] `pytest cdk/tests/test_backend_lambda_stack_permissions.py cdk/tests/test_backend_lambda_stack.py` — 70 passed
- [x] `pytest tests/backend` — 782 passed, 1 skipped, 12 xfailed — no regressions
- [x] `ruff check --config backend/pyproject.toml` — clean

Closes #4933

### DIFF
--- a/cdk/tests/test_backend_lambda_stack_permissions.py
+++ b/cdk/tests/test_backend_lambda_stack_permissions.py
@@ -878,6 +878,46 @@ def test_metadata_prefix_matches_backend_instruments_s3_location() -> None:
     )
 
 
+# Every module-level *_PREFIX string constant in backend_lambda_stack.py that is
+# duplicated in a backend fallback literal (see issue #4933's audit) must have a
+# corresponding cross-check test above. If this set drifts from what's actually
+# declared in the stack, either a new duplicated constant needs a cross-check
+# test, or this set is stale and should be trimmed.
+_COVERED_STACK_PREFIX_CONSTANTS = frozenset({"WRITABLE_ACCOUNTS_PREFIX", "METADATA_PREFIX"})
+
+
+def test_no_uncovered_duplicated_prefix_constants_in_backend_lambda_stack() -> None:
+    """Guards the issue #4933 audit finding: WRITABLE_ACCOUNTS_PREFIX and
+    METADATA_PREFIX are the only module-level ``*_PREFIX`` string constants in
+    ``backend_lambda_stack.py`` that duplicate a backend fallback literal, and
+    both already have a cross-check test. If a future change adds another
+    ``*_PREFIX`` constant here, this test fails as a reminder to either add a
+    matching cross-check test (and extend the covered set above) or, if the
+    new constant isn't actually duplicated in the backend, confirm that and
+    extend the set to acknowledge it's out of scope."""
+    stack_path = CDK_DIR / "stacks" / "backend_lambda_stack.py"
+    tree = ast.parse(stack_path.read_text(encoding="utf-8"))
+
+    declared_prefix_constants = {
+        target.id
+        for node in tree.body
+        if isinstance(node, ast.Assign)
+        for target in node.targets
+        if isinstance(target, ast.Name)
+        and target.id.endswith("_PREFIX")
+        and isinstance(node.value, ast.Constant)
+        and isinstance(node.value.value, str)
+    }
+
+    assert declared_prefix_constants == _COVERED_STACK_PREFIX_CONSTANTS, (
+        "Module-level *_PREFIX string constants in backend_lambda_stack.py "
+        f"({sorted(declared_prefix_constants)}) no longer match the audited/covered "
+        f"set ({sorted(_COVERED_STACK_PREFIX_CONSTANTS)}). Add a cross-check test "
+        "for any new constant that duplicates a backend fallback literal, or "
+        "update _COVERED_STACK_PREFIX_CONSTANTS if it's confirmed out of scope."
+    )
+
+
 def test_grant_bucket_access_raises_on_no_permissions() -> None:
     class _MockFn:
         def add_to_role_policy(self, policy_statement: object) -> None:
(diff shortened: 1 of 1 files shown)


leonarduk:
## Response to DeepSeek review

**Re: "blocking" false-negative risk in `_fallback_string_literal`** — traced through manually: `backend/common/instruments.py` has exactly one `ast.Assign` whose target is a plain `Name` called `prefix` (`prefix = os.getenv(METADATA_PREFIX_ENV, "instruments").strip("/")` at line 119). The two other `prefix` references in that file (`bucket, prefix = s3_loc`) are tuple-unpacking assigns, so their target is an `ast.Tuple`, not `ast.Name` — the helper's `isinstance(t, ast.Name)` check already excludes them. Within that one matching assignment, the only string `Constant` nodes in the value subtree are `"instruments"` and `"/"` (`.strip("/")` arg); `"/"` is explicitly excluded, so `"instruments"` is the unambiguous result. Confirmed by running the full suite — 22/22 pass in `test_backend_lambda_stack_permissions.py`, 71/71 across the CDK stack tests.

**Re: "only 1 of 2-3 constants covered"** — re-audited `cdk/stacks/*.py` for every module-level string constant: only `WRITABLE_ACCOUNTS_PREFIX` and `METADATA_PREFIX` are duplicated as fallback literals in `backend/`. `BACKEND_API_URL_EXPORT` (the only other module-level constant) is single-sourced via import, not duplicated. There genuinely isn't a third duplicated constant to add a test for — the issue's "2-3" was a suggested ceiling based on likely candidates, not a confirmed count.

Rather than leave that audit finding only in the PR description, pushed b2134b70 which adds `test_no_uncovered_duplicated_prefix_constants_in_backend_lambda_stack` — it statically parses `backend_lambda_stack.py` for `*_PREFIX` constants and asserts the set matches exactly `{WRITABLE_ACCOUNTS_PREFIX, METADATA_PREFIX}`. If a future PR adds a new duplicated `*_PREFIX` constant without a cross-check test, this fails as a reminder, closing the "audit only in prose" gap the review flagged.

PR #5356, opened 2026-07-19
test: verify enrich_holding's Monday fallback reaches the previous Friday's close

### LINKED ISSUE #5208: `enrich_holding` fallback should use last trading day, not `prev_date = pricing_date - 1`
### What

In `backend/common/holding_utils.py`, the `enrich_holding` function's fallback logic at lines 572-579 uses `prev_date = pricing_date - timedelta(days=1)` to compute `prev_px`. When `pricing_date` falls on a Monday (or any day after a market holiday), `prev_date` lands on a non-trading day, causing `_get_price_for_date_scaled` to return `None` and defeating the fallback's purpose.

### Why

This is a correctness risk: the fallback is designed to provide a price when today's close is missing, but it systematically fails ~1/7 of the time (weekends) plus any market holidays. This leads to `px = None` propagating through the holding enrichment, potentially causing downstream failures or silently missing prices that *are* available from the last actual trading day.

### How

Replace the simple `prev_date = pricing_date - timedelta(days=1)` with a loop that walks backwards through recent dates (e.g., up to 10 days) until a trading day with a valid close price is found. Alternatively, use an existing utility like `get_last_trading_day()` if one exists in the codebase, or query the exchange calendar for the most recent trading day before `pricing_date`.

Concrete approach:
1. In `enrich_holding`, replace the `prev_date` computation with a loop (max 10 iterations) that decrements by 1 day and calls `_get_price_for_date_scaled` until a non-None price is returned.
2. If no price is found after the loop, set `px = None` and `is_stale = True` (consistent with the existing stale logic).

### Constraints

- Must not change the behavior for cases where `prev_date` already works correctly (e.g., Tuesday-Friday).
- Must not introduce performance regressions — the loop is bounded to 10 iterations and only runs in the fallback path.
- Must not break the existing stale-price logic at lines 559-571; the fallback should either set `is_stale = True` and skip the stale check, or the stale check should run after the fallback (as noted in the review's first blocking issue).
- Out of scope: unifying NaN guards into a shared helper (separate follow-up), adding the `_derived_cost_basis_close_px` regression test (separate follow-up).

### LLM tier

**Sonnet** — this requires moderate design judgment: understanding the trading calendar semantics, choosing between a loop vs. a utility function, and ensuring the fix integrates correctly with the existing stale-price logic without introducing regressions. The change spans a single function but has subtle correctness implications.

### Success looks like

- `enrich_holding` returns a valid `px` when `pricing_date` is a Monday and the previous Friday has a valid close price.
- All existing tests pass, including `test_enrich_holding_falls_back_to_previous_close_when_today_missing`.
- A new test (or updated existing test) verifies the Monday/weekend fallback scenario.

### Failure looks like

- The fallback still returns `None` for Monday pricing dates because the loop doesn't find Friday's price (e.g., due to incorrect date arithmetic or exchange calendar lookup).
- The stale flag logic becomes inconsistent (e.g., `is_stale` set to `True` but the holding is actually using a fresh price from 1 day ago).
- Performance regression from unbounded looping or expensive calendar queries.

_Follow-up from AI review of PR #5188._

leonarduk:
## Summary
Investigated before implementing: this issue describes `enrich_holding`'s fallback using a naive `prev_date = pricing_date - timedelta(days=1)`, which would land on a non-trading Sunday when `pricing_date` is a Monday. Checked `backend/common/holding_utils.py` and that code doesn't exist — the fallback already uses `calc.previous_pricing_date` (`PricingDateCalculator`, weekend-aware via `_nearest_weekday`), so the bug this issue describes doesn't reproduce on current `main`.

Added `test_enrich_holding_monday_fallback_reaches_previous_friday` to lock this in as a regression test, since it wasn't directly covered — the existing `test_enrich_holding_falls_back_to_previous_close_when_today_missing` (despite its docstring) doesn't actually exercise the fallback branch at all; tracing through it, its fake `today` resolves `reporting_date` directly to the one date with a price, so `px` is set on the first lookup and the `prev_px` fallback path never triggers.

**Verified the new test has teeth**: manually monkeypatched `PricingDateCalculator.previous_pricing_date` to the naive `reporting_date - timedelta(days=1)` formula the issue describes and confirmed the assertions would then fail (`price` resolves to `None` instead of `42.0`), before reverting and confirming they pass against the real implementation.

### Scope note
The issue's secondary concern — "or any market holiday" — is a real, narrower gap: `_nearest_weekday` only skips weekends, not UK bank holidays. Left out of scope here since it's not covered by the issue's own stated success criteria (which specifically target the Monday/weekend case) and would need its own design decision about which UK-holiday calendar utility to wire in.

## Test plan
- [x] `pytest tests/backend/test_holding_utils.py` — 5 passed
- [x] Manual verification that the test fails against the naive buggy `previous_pricing_date` formula (see above)
- [x] `pytest tests/backend` — 783 passed, 1 skipped, 12 xfailed — no regressions
- [x] `ruff check --config backend/pyproject.toml` — clean

Closes #5208

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved price-date handling for holdings when pricing falls on a Monday.
  * Holdings now correctly use the previous Friday’s closing price instead of a non-trading day.
  * Stale pricing is accurately indicated when fallback pricing is used.

* **Tests**
  * Added regression coverage for Monday pricing fallback behaviour.

### DIFF
--- a/tests/backend/test_holding_utils.py
+++ b/tests/backend/test_holding_utils.py
@@ -190,3 +190,37 @@ def test_enrich_holding_falls_back_to_previous_close_when_today_missing(monkeypa
     assert result["price"] == pytest.approx(50.0)
     assert result["current_price_gbp"] == pytest.approx(50.0)
     assert result["market_value_gbp"] == pytest.approx(500.0)
+
+
+def test_enrich_holding_monday_fallback_reaches_previous_friday(monkeypatch):
+    """Regression for issue #5208: when pricing_date lands on a Monday and
+    Monday's own close is missing, the fallback must reach back to the
+    previous Friday's close rather than Sunday (a non-trading day). Uses
+    PricingDateCalculator.previous_pricing_date (already weekend-aware via
+    _nearest_weekday), not a naive pricing_date - timedelta(days=1)."""
+    import backend.common.portfolio_utils as pu
+
+    monkeypatch.setattr(hu, "get_instrument_meta", lambda *_: {"currency": "GBP"})
+    monkeypatch.setattr(pu, "get_security_meta", lambda *_: {})
+    monkeypatch.setattr(pu, "_PRICE_SNAPSHOT", {})
+
+    previous_friday = dt.date(2026, 5, 15)
+
+    def fake_price_for_date(ticker, exchange, d, field="Close_gbp"):
+        # Monday's close is missing; only the previous Friday has a price.
+        if d == previous_friday:
+            return 42.0, "Yahoo"
+        return None, None
+
+    monkeypatch.setattr(hu, "_get_price_for_date_scaled", fake_price_for_date)
+    monkeypatch.setattr(hu, "get_effective_cost_basis_gbp", lambda h, cache, price_hint=None: 0.0)
+
+    holding = {TICKER: "FOO.L", UNITS: 10, COST_BASIS_GBP: 0.0, ACQUIRED_DATE: "2025-01-01"}
+    # today = Tuesday so calc.reporting_date resolves to the preceding Monday.
+    today = dt.date(2026, 5, 19)
+    result = hu.enrich_holding(holding, today, price_cache={})
+
+    assert result["price"] == pytest.approx(42.0)
+    assert result["current_price_gbp"] == pytest.approx(42.0)
+    assert result["market_value_gbp"] == pytest.approx(420.0)
+    assert result["is_stale"] is True


leonarduk:
## Re: DeepSeek AI Code Review (REQUEST CHANGES)

Checked this against the actual PR branch rather than the issue's description of the bug, and the review's premise doesn't hold up.

**Claim:** `enrich_holding` still uses `prev_date = pricing_date - timedelta(days=1)`, so for a Monday `pricing_date` it computes Sunday, the mock returns `None`, and the new test's assertions (`price == 42.0`, `is_stale is True`) should fail.

**Verification:**

1. Read `backend/common/holding_utils.py` on this branch (around line 571-582) — the fallback path is:
   ```python
   prev_date = calc.previous_pricing_date
   ...
   prev_px, _ = _get_price_for_date_scaled(ticker, exchange, prev_date, field="Close_gbp")
   ```
   `calc` is a `PricingDateCalculator`, and `previous_pricing_date` is weekend-aware (via `_nearest_weekday`). There is no `pricing_date - timedelta(days=1)` naive subtraction anywhere in the current `enrich_holding` — that formula only exists in the original issue's *description* of the bug, not in the code on `main`/this branch.

2. Actually ran the test suite on this branch:
   ```
   pytest tests/backend/test_holding_utils.py -v
   ============================== 5 passed in 3.08s ==============================
   ```
   All 5 tests pass, including `test_enrich_holding_monday_fallback_reaches_previous_friday`.

So the review appears to have evaluated the issue's hypothetical buggy code rather than the actual diff/branch contents. No production code change is needed here — the PR description's own investigation (that `previous_pricing_date` already resolves Monday → Friday correctly) is correct, and the new regression test passes as intended.

CodeRabbit's review on this PR separately reported no actionable comments.

PR #5411, opened 2026-07-19
Time-bound snapshot loading in cold-start warmup

### LINKED ISSUE #4940: "Make snapshot loading (`_load_snapshot`, `_update_latest_prices_from_snapshot`) also non-blocking or time-bounded"
Here is a complete, actionable GitHub issue body based on the provided review.

---

### What

The synchronous `await` calls to `_load_snapshot` and `_update_latest_prices_from_snapshot` in `startup.py` (lines 76-79) must be made non-blocking or time-bounded. Currently, these operations block the entire startup coroutine, which violates AC1 of the original issue (health/config respond quickly).

**Files & locations:**
- `startup.py`, lines 76-79: The `await` calls to snapshot loading
- `startup.py`, `_load_snapshot` method: Needs a timeout wrapper or conversion to a background task
- `startup.py`, `_update_latest_prices_from_snapshot` method: Same treatment

### Why

- **Correctness risk:** On cold start with a large snapshot file (e.g., on S3), these synchronous awaits can consume most of the 30s Lambda timeout, leaving no room for the actual request handler.
- **Agent confusion:** The current code partially addresses the startup blocking issue but leaves a critical gap that undermines the entire fix. Future maintainers may incorrectly assume startup is fully non-blocking.
- **Maintainability:** Without this change, the system remains vulnerable to snapshot size/network latency, and the AC1 requirement is only half-met.

### How

1. **Wrap snapshot loading in a timeout:** Use `asyncio.wait_for` with a configurable timeout (e.g., 5 seconds) around the snapshot loading calls. On timeout, log a warning and proceed without the snapshot data.
2. **Alternative: Convert to background task:** Move `_load_snapshot` and `_update_latest_prices_from_snapshot` into a separate background task (similar to `_prime_latest_prices_background`), with GC protection (module-level reference or awaited with timeout in the request handler).
3. **Add instrumentation:** Insert timing metrics (e.g., `time.perf_counter` or structured logging) around both snapshot operations to measure their duration, addressing the "reproduce/instrument the cold-start path" requirement from the original issue.

**Recommended approach (simplest):** Add `asyncio.wait_for(snapshot_coro, timeout=5.0)` and log a warning if it times out. This is minimal, reversible, and directly addresses the AC1 gap.

### Constraints

- Must not break `SKIP_SNAPSHOT_WARM` behavior (AC2)
- Must not change the API contract or frontend behavior
- Must not introduce new blocking points in the request path
- The existing `_warm_snapshot` pattern (line 85-87) should remain as a fallback
- Out of scope: Refactoring the `@staticmethod` decorator, adding CDK tests, or fixing the background task GC issue (those are separate follow-ups)

### LLM Tier

**Sonnet** — This requires moderate design judgment: choosing between timeout vs. background task, deciding where to place the timeout, and adding instrumentation without breaking existing behavior. The change spans multiple methods in one file but involves non-trivial async control flow decisions.

### Success looks like

- `startup()` returns within a bounded time (e.g., < 10s) even when snapshot loading is slow or hangs
- A warning log is emitted when snapshot loading times out
- Timing metrics for snapshot loading appear in logs (e.g., `"snapshot_load_duration_ms": 1234`)
- All existing tests pass, including `test_app_bootstrap.py`
- AC1 is fully satisfied: health/config endpoints respond quickly on cold start

### Failure looks like

- Snapshot data is silently missing when timeout occurs (no warning log)
- The timeout is too short, causing snapshot data to be unavailable in normal conditions
- The change introduces a new blocking point elsewhere in the startup path
- Tests fail due to changed timing behavior (e.g., test assumes snapshot is always loaded)
- The `SKIP_SNAPSHOT_WARM` env var is accidentally bypassed

---

_Follow-up from AI review of PR #4934._

leonarduk:
## What
Closes #4940

`_warm_snapshot()`'s synchronous calls to `_load_snapshot()` and `instrument_api.update_latest_prices_from_snapshot()` ran directly on the event loop inside `startup()`. On a cold start with a large snapshot file (e.g. on S3) or slow network, these could consume most of the Lambda timeout before any request handler — including `/health` — got a chance to run.

## Changes
- `backend/bootstrap/startup.py`: both calls now run via `asyncio.to_thread(...)` (off the event loop) wrapped in `asyncio.wait_for(..., timeout=_SNAPSHOT_LOAD_TIMEOUT_SECONDS)`.
  - New module constant `_SNAPSHOT_LOAD_TIMEOUT_SECONDS`, default `5.0`, overridable via the `SNAPSHOT_LOAD_TIMEOUT_SECONDS` env var.
  - On timeout: log a `logger.warning(...)` and fall back to the same "no data" path already used for other snapshot-load failures (empty snapshot / skip the metadata update) — not a new blocking point, not a new failure mode, just a bounded one.
  - `SKIP_SNAPSHOT_WARM` behavior, the `_warm_snapshot` fallback pattern, and the API contract are all unchanged; the background `_prime_latest_prices_background` task and its own timing/logging are untouched.
- Added two regression tests to `tests/backend/test_app_bootstrap.py`:
  - `test_lifecycle_service_snapshot_load_is_time_bounded` — a hanging `_load_snapshot()` doesn't block `startup()` past the (test-shortened) timeout; snapshot falls back to `{}`/`None` and a warning is logged.
  - `test_lifecycle_service_metadata_update_is_time_bounded` — same for a hanging `update_latest_prices_from_snapshot()`.

## Validation
- [x] `pytest tests/backend/test_app_bootstrap.py --no-cov` — 9/9 pass (7 existing + 2 new)
- [x] `ruff check` / `black --check` on changed files — no new violations (pre-existing lint debt on untouched lines only)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/backend/bootstrap/startup.py
+++ b/backend/bootstrap/startup.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import os
 import shutil
 import time
 from contextlib import contextmanager
@@ -16,6 +17,13 @@ from backend.logging_setup import sanitise_log_value
 from backend.utils import page_cache
 
 logger = logging.getLogger(__name__)
+
+# Time-bound the two synchronous, potentially-slow snapshot warmup calls in
+# _warm_snapshot() (S3 read + in-memory metadata scan) so a large snapshot or
+# slow network can never consume the whole Lambda cold-start budget (#4940).
+# Each call gets its own budget rather than sharing one, so a timeout on the
+# first doesn't eat into the second's allowance.
+_SNAPSHOT_LOAD_TIMEOUT_SECONDS = float(os.getenv("SNAPSHOT_LOAD_TIMEOUT_SECONDS", "5.0"))
 
 
 @contextmanager
@@ -84,10 +92,19 @@ class AppLifecycleService:
 
         try:
             with _timed_phase("snapshot_load"):
-                result = _load_snapshot()
+                result = await asyncio.wait_for(
+                    asyncio.to_thread(_load_snapshot),
+                    timeout=_SNAPSHOT_LOAD_TIMEOUT_SECONDS,
+                )
             if not isinstance(result, tuple) or len(result) != 2 or not isinstance(result[0], dict):
                 raise ValueError("Malformed snapshot")
             snapshot, ts = result
+        except TimeoutError:
+            logger.warning(
+                "Snapshot load timed out after %ss; proceeding without snapshot data",
+                _SNAPSHOT_LOAD_TIMEOUT_SECONDS,
+            )
+            snapshot, ts = {}, None
         except Exception as exc:
             logger.error("Failed to load price snapshot: %s", exc)
             snapshot, ts = {}, None
@@ -98,7 +115,15 @@ class AppLifecycleService:
 
         try:
             with _timed_phase("metadata_update_from_snapshot"):
-                instrument_api.update_latest_prices_from_snapshot(snapshot)
+                await asyncio.wait_for(
+                    asyncio.to_thread(instrument_api.update_latest_prices_from_snapshot, snapshot),
+                    timeout=_SNAPSHOT_LOAD_TIMEOUT_SECONDS,
+                )
+        except TimeoutError:
+            logger.warning(
+                "Updating latest prices from snapshot timed out after %ss",
+                _SNAPSHOT_LOAD_TIMEOUT_SECONDS,
+            )
         except Exception:
             logger.exception("Failed to update latest prices from snapshot")
 
--- a/backend/bootstrap/startup.py
+++ b/backend/bootstrap/startup.py
@@ -102,11 +102,11 @@ class AppLifecycleService:
         except TimeoutError:
             logger.warning(
                 "Snapshot load timed out after %ss; proceeding without snapshot data",
-                _SNAPSHOT_LOAD_TIMEOUT_SECONDS,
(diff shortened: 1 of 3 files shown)


leonarduk:
Re: [DeepSeek review](#) — pushing back on both blocking findings; neither reproduces against the current file.

**"`sanitise_log_value` is undefined"** — it isn't. It's imported at the top of `backend/bootstrap/startup.py` (`from backend.logging_setup import sanitise_log_value`) and was already in use by the existing `_timed_phase()` helper before this PR touched the file. The commit just added two more call sites for an already-imported symbol.

**"bare `TimeoutError` won't catch `asyncio.TimeoutError`"** — not for this codebase. Since Python 3.11, `asyncio.TimeoutError` is an alias for the builtin `TimeoutError` (see the [asyncio exceptions docs](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.TimeoutError)), and `pyproject.toml` pins `requires-python = ">=3.11"` — CI itself runs Python 3.12.13. So `except TimeoutError` correctly catches what `asyncio.wait_for` raises here. The review's own point 4 ("tests pass... because `TimeoutError` is an alias for `asyncio.TimeoutError`... on Python 3.11+") actually confirms this is fine, not a bug — 3.11+ is this project's floor, not an edge case to guard against.

Sections 1 (acceptance criteria) and 4 (test coverage intent) look accurate otherwise. No code changes made in response to this review — verified locally that `sanitise_log_value` is imported and that the log-sanitization audit test (the actual CI failure) passes with the current fix.

Issue #2749, opened 2026-04-12
Automate reading of transactions from Moneyhub


leonarduk:
Automate reading of transactions from Moneyhub (https://client.moneyhub.co.uk/#transactions) rather than manual entry. Investigate Moneyhub API or scraping approach.

leonarduk:
## Scoping detail (added to reduce complexity)

Moneyhub is an Open Banking aggregator with an OAuth2 / OpenID Connect API (no public scraping-friendly surface, and scraping the web client would break and likely violates TOS). The genuine complexity here is the auth/integration spike, which is why this stays a larger task. Propose splitting into two phases:

### Phase 1 — Spike (this issue)
- Register a Moneyhub API client and confirm available scopes for reading transactions (`transactions:read` or equivalent) against their sandbox.
- Prototype the OAuth2 client-credentials / user-consent flow needed to obtain a transactions token.
- Pull a sample of transactions for one account and document the JSON shape returned.
- Decide where tokens/secrets live (SSM Parameter Store / Secrets Manager — never hardcoded) and how refresh is handled.
- Deliverable: a short design note + working sandbox call, not production wiring.

### Phase 2 — Implementation (follow-up issue, to be split out)
- Scheduled Lambda that pulls new transactions and maps them into the existing transaction model.
- Idempotent ingestion keyed on Moneyhub transaction id.

### Acceptance criteria for Phase 1
- [ ] Documented Moneyhub auth flow and required scopes.
- [ ] Working sandbox call retrieving sample transactions, with the response schema captured.
- [ ] Decision recorded on secret storage and token refresh strategy.
- [ ] Mapping plan from Moneyhub transaction fields to the AllotMint transaction model.

### Constraints
- Do NOT scrape `client.moneyhub.co.uk`; use the official API only.
- All credentials/tokens via SSM/Secrets Manager, configured through CDK.

leonarduk:
## Rescoped & split (stays `opus`)
Kept at `opus` because the access-method decision (Moneyhub Open Banking API registration/OAuth2/cost vs scraping vs manual export) carries real unknowns and a security/auth design.

**This issue is now the feasibility spike**: evaluate the options and recommend an approach + auth model. Implementation is split into downstream issues that can be picked up by smaller models once the approach is fixed:
- #3425 — Implement transaction import via chosen approach (`sonnet`) — depends on this spike
- #3426 — Moneyhub CSV/OFX export parser fallback (`haiku`) — independent, delivers value regardless

[...]

PR #5477, opened 2026-07-20
feat(backend): Moneyhub live API transaction import (#3425)

### LINKED ISSUE #3425: Implement Moneyhub transaction import via chosen approach
# Implement Moneyhub transaction import via chosen approach

## What
Ingest transactions from Moneyhub using the access method selected by the feasibility spike (#2749), map them into the existing transaction model, and deduplicate against existing records.

## Why
Manual entry of Moneyhub transactions doesn't scale; once the auth/access approach is confirmed by #2749, this issue delivers the actual automated ingestion.

## How
- Implement the importer using the approach selected by the #2749 spike (expected to be the Moneyhub Open Banking OAuth2/OIDC API, per #2749's scoping comments — no scraping). **Note (2026-07-08 review):** no existing OAuth2-client-to-external-API pattern exists anywhere in this repo to use as a starting template — all existing OAuth-related code (`backend/auth.py`, `frontend/src/awsUiAuth.ts`) is *inbound* auth (verifying Cognito/Google id-tokens for users logging into AllotMint), not an outbound OAuth2 client calling a third-party API. Budget real time for this rather than assuming there's a pattern to copy.
- Add a new provider entry to `backend/importers/__init__.py` (`_IMPORTER_PATHS`), mapping Moneyhub fields to the existing `Transaction` model (defined at `backend/routes/transactions.py:33`, a Pydantic model with `owner`/`account`/`id`/`date`/`ticker`/`type`/`amount_minor`/`price`/`units`/`fees`/etc., `extra="ignore"`). `backend/importers/degiro.py:21-41` shows the `parse(data: bytes) -> List[Transaction]` shape a Moneyhub importer should follow.
- Deduplicate against existing transactions — this logic does not exist anywhere in the codebase today (confirmed: `backend/routes/transactions.py`'s `import_transactions` handler, lines 670-680, parses and returns transactions directly with no persistence or dedupe step), so it must be built as part of this issue, keyed on a stable Moneyhub transaction id.
- **Cross-reference #3426**: it needs similar dedupe logic, but note the *key derivation* differs even if the comparison/storage logic can be shared — Moneyhub's API returns a real, stable external transaction id directly, while #3426's CSV import has no source-provided id and must derive a synthetic composite key (e.g. date+ticker+amount+account). Whichever of #3425/#3426 lands first should build the comparison/storage step as a shared, provider-agnostic helper that accepts a pre-derived key — not assume the key-derivation logic itself is shareable. Check for that helper before writing a new one.
- Mock external Moneyhub calls in tests; add pytest coverage under `tests/`.
- Store any credentials/tokens via SSM/Secrets Manager per #2749's constraints — never hardcoded. **Note:** no `secretsmanager.Secret`/SSM `StringParameter` CDK construct exists anywhere in `cdk/` to copy (same gap noted in #2749) — the existing pattern to follow instead is `backend/common/storage.py`'s `ssm://parameter-name` URI scheme (`boto3.client("ssm").get_parameter`, roughly lines 93-132), which #2749 should also settle on.

## Constraints
- **Depends on #2749** (feasibility spike) — do not start until the access method and auth model are decided there. (The milestone already correctly marks this issue as blocked, not ready-to-start.)
- Access-method decision and credentials/OAuth design are out of scope here (covered by #2749).
- Dedupe-against-existing-records logic has no precedent in this codebase to copy from; treat it as new design work, not a reuse of an existing pattern. The key-derivation step specifically (stable external id vs. synthetic composite key) is provider-specific — don't assume it's identical to #3426's.

## LLM tier
**Sonnet** — moderate complexity: OAuth-authenticated API client, field mapping, and new dedupe logic, but the approach is pre-decided by the #2749 spike.

## Success looks like
- [ ] Moneyhub transactions are pulled via the approach chosen in #2749 and mapped into the `Transaction` model
- [ ] Re-running the import does not create duplicate transactions
- [ ] External Moneyhub calls are mocked in tests; pytest coverage added under `tests/`

## Failure looks like
- Importer duplicates transactions on repeated runs because no dedupe key is enforced
- Tests make live calls to the real Moneyhub API instead of mocking
- Credentials end up hardcoded or committed instead of stored via SSM/Secrets Manager (via the `backend/common/storage.py` `ssm://` pattern)
- Dedupe logic is built twice (once here, once in #3426) because neither issue checked whether the other had already added a shared helper

leonarduk:
## Summary
- Adds `backend/integrations/moneyhub_api.py`: a stateless OAuth2 HTTP client (`refresh_access_token`, `fetch_transactions`) against the Moneyhub Open Banking API, following the access method decided in #2749 (never scraping `client.moneyhub.co.uk`).
- Adds `backend/common/moneyhub_tokens.py`: per-owner OAuth token storage via the existing `ssm://` scheme (`ParameterStoreJSONStorage`, `SecureString`), with lazy refresh-on-read — the storage decision from `docs/moneyhub-integration.md`.
- Adds `backend/importers/moneyhub_api.py`: maps raw Moneyhub transaction JSON onto the `Transaction` model (`kind="account"`, `external_id` prefixed `moneyhub:`, `pending` transactions excluded), reusing the shared `dedupe_against_existing` helper #3426's CSV importer already added — no dedupe logic duplicated.
- Adds `POST /transactions/import/moneyhub` to `backend/routes/transactions.py`, which pulls an owner's transactions live from the API and persists new ones through the same `_persist_transaction` path as the file-upload importer, returning `424` if the owner has no stored Moneyhub consent and `502` on upstream API failures.
- Updates `docs/moneyhub-integration.md` with an implementation-status section.

## Context
This closes out the implementation half of #2749/#3425. Real Moneyhub sandbox credentials are still a human prerequisite per #2749 (no self-service developer registration), so this is built against the documented-but-unconfirmed API schema (`id`, `accountId`, `date`, `amount.amount`/`amount.currency`, `description`, `category`, `status`) with all external HTTP calls mocked in tests. Once real sandbox access exists, the field mapping in `backend/importers/moneyhub_api.py` should be verified/adjusted against an actual response.

Out of scope for this PR (flagged as follow-up in the doc): the CDK `StringParameter`/IAM construct for the `/allotmint/moneyhub/*` namespace, and the Moneyhub `accountId` → AllotMint account-slug linking table.

Closes #3425

## Test plan
- [x] `python -m pytest tests/test_moneyhub_api_import.py -q` — 20 tests, 97-100% coverage on new modules
- [x] `python -m pytest tests/` — full suite, 3103 passed, 92.32% coverage (gate: 90%)
- [x] `ruff check` / `black --check` on all changed/new files


leonarduk:
@codex fix tests

leonarduk:
Closing as superseded. Every file this PR adds — `backend/integrations/moneyhub_api.py`, `backend/common/moneyhub_tokens.py`, `backend/importers/moneyhub_api.py`, the `/transactions/import/moneyhub` route in `backend/routes/transactions.py`, and `tests/test_moneyhub_api_import.py` — already exists on `main`, merged separately via #5624 and #5628 ("Reuse Moneyhub client and validate credentials"). Main's version is a strict superset: it includes everything here plus pagination support and stricter client-credential validation (raises clearly if `MONEYHUB_CLIENT_ID`/`MONEYHUB_CLIENT_SECRET` are unset, instead of silently defaulting to empty strings — the exact gap flagged in review here).

That's why this PR shows as conflicting rather than mergeable: it's not a real conflict to resolve, the work has already landed through a different path. No unique changes from this branch need to be carried forward.

PR #5624, opened 2026-07-29
Add Moneyhub transaction pagination

### LINKED ISSUE #5478: Add pagination support to `MoneyhubClient.fetch_transactions`
# Add pagination support to `MoneyhubClient.fetch_transactions`

## What

Modify `MoneyhubClient.fetch_transactions` in `backend/integrations/moneyhub_api.py` to handle paginated API responses from the Moneyhub transactions endpoint.

## Why

The current implementation makes a single GET request with no pagination handling. If Moneyhub returns paginated results (common for transaction APIs), only the first page of transactions is imported. This is a data-loss risk — subsequent imports will miss transactions, and the deduplication logic cannot compensate for data that was never fetched.

## How

1. **Detect pagination in the response** — Check if the Moneyhub API response includes pagination metadata (e.g., `next` URL, `totalPages`, or `hasMore`). Based on typical REST pagination patterns, look for a `next` link in the response body or headers.

2. **Loop until all pages are consumed** — After processing each page, if a next-page indicator exists, make the next request (using the same authentication token) and append the results. Continue until no more pages remain.

3. **Return the combined list** — The function signature should remain unchanged; it returns a single list of all transactions across all pages.

4. **Add a configurable page limit** — Introduce an optional `max_pages` parameter (default: 10) to prevent runaway pagination in case of unexpected API behavior.

5. **Update the route handler** — No changes needed in `backend/routes/transactions.py`; the route already calls `fetch_transactions` and processes the returned list.

## Constraints

- **Must not break existing behavior** — Single-page responses must continue to work identically.
- **Must not introduce new dependencies** — Use only `requests` (already used) or standard library modules.
- **Must not change the function signature** — Keep `fetch_transactions()` callable without arguments from the route handler.
- **Out of scope** — Rate limiting, retry logic, or caching of paginated results. These can be follow-up issues.
- **Out of scope** — Changing the `MoneyhubClient` constructor or authentication flow.

## LLM tier

**Sonnet** — This is a moderate design task requiring understanding of the existing API response structure, pagination patterns, and the need to maintain backward compatibility. The implementation involves non-trivial heuristics (detecting pagination format) and a small design decision (max_pages default). A smaller model may not have sufficient context about the Moneyhub API response format.

## Success looks like

- `fetch_transactions` returns all transactions from a paginated Moneyhub API response (verified by a test that mocks a multi-page response).
- A new test in `tests/test_moneyhub_api.py` covers the pagination scenario with at least 2 pages.
- The existing single-page test continues to pass.
- The `max_pages` parameter is documented in the function docstring and defaults to 10.

## Failure looks like

- The function still only returns the first page (pagination not implemented).
- The function enters an infinite loop if the API returns a `next` link that points back to the same page.
- The function breaks when the API returns a non-paginated response (e.g., missing pagination fields).
- The function signature changes, requiring updates to the route handler.
- No new test covers the pagination scenario.

_Follow-up from AI review of PR #5477._

leonarduk:
### Motivation

Closes #5478 

- Moneyhub transaction responses can be paginated and the previous single-request implementation risked omitting transactions beyond the first page. 
- Ensure imports return a combined list of transactions across pages while preserving existing single-page and bare-list behaviour. 
- Prevent runaway loops from malformed upstream pagination by bounding page traversal and detecting repeated URLs. 
- Validate each followed page URL to avoid SSRF-style risks when the client follows next-page links.

### Description

- Add a stateless OAuth2/HTTP adapter `MoneyhubClient` in `backend/integrations/moneyhub_api.py` with `refresh_access_token` and a paginating `fetch_transactions` method. 
- Implement pagination support that follows next-page indicators from common formats (`next`, nested `links`/`pagination`, object `href`) and RFC-style `Link` headers, resolving relative URLs with `urljoin`. 
- Add protections: configurable `max_pages` parameter (default 10), repeated-URL cycle detection, per-page `validate_external_url` calls, and consistent `MoneyhubAPIError` wrapping for network failures. 
- Factor page retrieval into `_fetch_transaction_page` and next-link parsing into `_next_page_url`, and add unit tests in `tests/test_moneyhub_api.py` covering body links, `Link` headers, max-page enforcement, repeated URL cycles, account filtering, and single-page responses, and include `Closes #5478`.

### Testing

- Ran `PYENV_VERSION=3.11.15 python -m pytest tests/test_moneyhub_api.py -q --no-cov` and observed `5 passed` for the new focused test suite. 
- Ran `PYENV_VERSION=3.11.15 python -m ruff check --config backend/pyproject.toml backend/integrations/moneyhub_api.py tests/test_moneyhub_api.py` and `PYENV_VERSION=3.11.15 python -m black --check --config backend/pyproject.toml backend/integrations/moneyhub_api.py tests/test_moneyhub_api.py`, both of which succeeded after formatting. 
- Note: running the full repository test+coverage gate reports an unrelated global coverage failure when only the focused tests are executed against the large existing codebase, so focused verification was performed with `--no-cov` as shown above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a6a6733a7f883278d66761d85c3d306)


leonarduk:
@codex fix tests

Issue #5594, opened 2026-07-28
CI: intelligently skip integration/heavy test jobs for doc-only / comment-only changes


leonarduk:
## What

Skip heavyweight CI jobs (backend integration tests, Lambda-compat pytest, Python-compatibility smoke, CDK infra tests, frontend smoke) when a PR's changed files are all "safe" — comments/docstrings, `.md`/`docs/**`, or other non-executable content that cannot affect runtime behavior.

Concretely: extend `.github/workflows/ci.yml` and `.github/workflows/backend-integration.yml` with a `dorny/paths-filter`-based (or diff-based) gate similar to the existing `detect-frontend-changes` job in `ci.yml` (lines 220-234), but instead of gating *in* on relevant paths, gate on whether the change is *purely* a comment/docstring/doc edit and skip the expensive jobs when so.

## Why

[PR #5554](https://github.com/leonarduk/allotmint/pull/5554) is a good example: it added a 5-line code comment to `backend/bootstrap/startup.py` explaining a timeout constant (a documentation-only change with zero behavioral impact), but it still had to run the full `backend-integration.yml` suite (provider parity tests, mypy, full pytest with 80% coverage gate, coverage upload) and all of `ci.yml` (CDK tests, PowerShell/bats tests, frontend build+lint+typecheck+vitest, python-compatibility smoke).

This wastes CI minutes and reviewer wait time on every comment-only or doc-only fix — a category of PR that recurs regularly via the aider/Claude auto-fix workflow (see `scripts/developer_tools/implement_issue.ps1` and the `.aider.conf.yml`-driven issue pipeline), since many follow-up issues from AI PR review are exactly this kind of "add a clarifying comment" change.

## How

1. Add a `classify-change` (or similarly named) job to `ci.yml` that:
   - Uses `git diff` against the PR base to get changed hunks per file (not just changed files — a file can have both a code change and a comment change).
   - Classifies the PR as **doc-only** if every changed line, across every changed file, is one of: a full-line comment (`#...` in Python, `//...`/`/* */` in TS/JS), a blank line, or falls inside a file matched by doc-glob patterns (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`, etc.) — including added comment lines inside otherwise-unchanged code, which is the PR #5554 case.
[...]

PR #5723, opened 2026-07-31
ci: run only the test jobs a PR's changes can affect

### LINKED ISSUE #5722: ci: run only the test jobs affected by a PR instead of the full suite
## What

Make CI test execution **change-aware**: run only the test jobs relevant to the
areas a PR actually touches, instead of running the whole suite (backend pytest,
frontend vitest/lint/type-check, CDK tests, bats, Pester, Playwright smoke) on
every PR.

Concretely, introduce a single shared "affected areas" classifier that emits
per-area boolean outputs (`backend`, `frontend`, `cdk`, `shell`, `powershell`,
`workflows`, `docs`), and gate the existing test jobs on those outputs.

## Why

Today every PR runs effectively everything, and the change-detection logic that
does exist is fragmented across **four different mechanisms**:

| Mechanism | Where | Scope |
|---|---|---|
| `scripts/classify_change.py` (doc-only) | `ci.yml` + `backend-integration.yml` (job duplicated verbatim) | all-or-nothing |
| `dorny/paths-filter` | `ci.yml` → `detect-frontend-changes` | frontend only |
| custom `changed-files` step | `iac-validation.yml` → `detect-iac-changes` | IaC only |
| `on.pull_request.paths` | `cdk-dry-run.yml` | whole workflow |

Consequences:

- A backend-only PR still runs `npm ci`, ESLint, `tsc -b`, vitest **and** a
  second full frontend install in `Frontend Tests / frontend-tests`.
- A frontend-only PR still runs the full backend pytest suite twice — once in
  `Backend Integration Tests / integration-tests` (with coverage + mypy) and
  once in `CI / Lambda-compat pytest`.
- A `scripts/*.ps1`-only PR runs backend, frontend, CDK and bats jobs, none of
  which can be affected by it.
- `ci.yml`'s `test` job is a grab-bag: branch-protection validation, lockfile
  platform coverage, frontend lint/type-check/vitest **and** CDK pytest all in
  one job, so none of it can be gated independently.
- The `classify-change` job is copy-pasted between `ci.yml` and
  `backend-integration.yml` with a comment explaining that job outputs can't
  cross workflow files — the fix for that (a reusable workflow) is the same
  fix needed here.

The doc-only skip (#5614) proved the pattern works and is safe; this issue
generalises it from one binary flag to per-area flags.

## How

High level, in reviewable stages:

1. **Extend the classifier.** Generalise `scripts/classify_change.py` (or add a
   sibling `scripts/classify_affected_areas.py`) to map the changed-file list to
   area flags via an explicit, checked-in path→area map. Keep the existing
   conservative posture: unknown path ⇒ *all* areas affected; non-`pull_request`
   events ⇒ all areas affected; classifier failure ⇒ all areas affected. Add
   pytest coverage under `tests/` for the mapping, mirroring the existing
   `classify_change` tests.
2. **Make it a reusable workflow.** Move the classify job into
   `.github/workflows/_classify-change.yml` (`workflow_call`) exposing the area
   flags as outputs, and have `ci.yml`, `backend-integration.yml`,
   `frontend-tests.yml` and `iac-validation.yml` all `uses:` it. This removes
   the duplicated job and the three competing detection mechanisms.
3. **Split `ci.yml`'s `test` job.** Separate the always-run repo-hygiene checks
   (`check_branch_protection_required_checks.py`,
   `check_lockfile_platform_coverage.py`, `test_extract_verdict.py`) from the
   frontend checks and the CDK pytest run, so each can be gated independently.
4. **Gate the jobs** on the relevant area flag: `frontend` → frontend lint /
   type-check / vitest / smoke; `backend` → integration-tests, lambda-compat,
   python-compatibility; `cdk` → CDK tests + `cdk-dry-run`; `shell` → bats;
   `powershell` → Pester.
5. **Keep the merge gate intact.** Every context listed in
   `EXPECTED_REQUIRED_CHECKS` in
   `scripts/check_branch_protection_required_checks.py` must still resolve on
   every PR. Either keep the `if: "!cancelled() && <flag> != 'false'"` shape
   already used for doc-only, or introduce an aggregate `ci-gate` job that
   `needs:` the area jobs and is the single required context. Whichever is
   chosen, update the ruleset JSON and that script together.
6. **Document** the path→area map and the "how do I add a new area" procedure
   in `AGENTS.md` / `docs/`, so the map doesn't silently rot.

Files most likely touched: `.github/workflows/ci.yml`,
`.github/workflows/backend-integration.yml`,
`.github/workflows/frontend-tests.yml`, `.github/workflows/iac-validation.yml`,
`.github/workflows/cdk-dry-run.yml`,
`.github/rulesets/default-branch-protection.json`,
`scripts/classify_change.py`,
`scripts/check_branch_protection_required_checks.py`, `tests/`.

## Constraints

- **No test may be silently dropped.** The union of jobs run across all possible
  PRs must equal today's coverage. `push`-to-`main` events must keep running the
  full suite unconditionally.
- **Fail-safe, never fail-open.** Any ambiguity — unrecognised path, classifier
  crash, missing base SHA, fork PR — must widen to "run everything", exactly as
  `classify_change.py` does today.
- **Cross-area dependencies must be encoded explicitly.** `backend/**` changes
  affect CDK synth (already reflected in `cdk-dry-run.yml`'s `paths`), and
  `backend/contracts_spa` changes affect the frontend contract-version check.
  These edges belong in the path→area map, not in reviewers' heads.
- **Branch protection must not break.** `scripts/check_branch_protection_required_checks.py`
  is itself a required check; the ruleset JSON and the script must stay in sync
  within the same PR.
- Shared workflow changes must be validated by `actionlint` (`lint-workflows`
  job) — it stays ungated.
- Prefer generalising the existing classifier over adding a fifth detection
  mechanism. Pin any new third-party action to a commit SHA (see #4814).

## LLM tier

**opus** — cross-cutting change spanning five workflow files, the branch
protection ruleset, and the merge gate itself. A wrong `if:` expression here
doesn't fail loudly; it silently stops running a test suite while the PR still
shows green. Needs whole-system reasoning about skip-propagation and required
status checks, not mechanical edits.

## Success looks like

- [ ] One shared, reusable change-classification workflow; the duplicated
      `classify-change` job in `backend-integration.yml` is gone.
- [ ] A backend-only PR does not run frontend install/lint/type-check/vitest or
      the Playwright smoke job.
- [ ] A frontend-only PR does not run the backend pytest suite, mypy, or
      lambda-compat.
- [ ] A `scripts/**.ps1`-only PR runs Pester (plus hygiene/lint) and nothing else.
- [ ] A docs-only PR behaves exactly as it does today (#5614 unchanged).
- [ ] Every context in `EXPECTED_REQUIRED_CHECKS` still resolves on every PR;
      `check_branch_protection_required_checks.py` passes and the ruleset JSON
      matches.
- [ ] `push` to `main` still runs the full suite.
- [ ] Unit tests cover the path→area mapping, including the fail-safe
      "unknown path ⇒ everything" case.
- [ ] Median PR CI wall-clock measurably drops; the reduction is reported in
      the PR description with before/after run times.

## Failure looks like

- A PR merges green because the job that would have caught the regression was
  skipped by a too-narrow path filter.
- The path→area map drifts from reality as directories are added, and nobody
  notices because the symptom is *fewer* tests running, not a red build.
- A required status check never reports because a gated job skipped, leaving PRs
  permanently blocked on a pending check.
- Detection logic multiplies instead of consolidating — a fifth mechanism is
  added alongside the four that already exist.
- Wall-clock savings are traded for flakiness: jobs gated on a classifier that
  behaves differently for fork PRs, merge queues, or force-pushed branches.

leonarduk:
## What

Generalises the doc-only CI fast path (#5594) into **per-area gating**: a PR now
runs the test suites its changes can actually break, instead of all of them.

Closes #5722

## How

**One classifier, not four.** `scripts/classify_change.py` now emits a flag per
area — `backend`, `frontend`, `cdk`, `shell`, `powershell` — alongside the
existing `doc-only`, derived from an ordered path→area map (`AREA_PATH_RULES`).
A doc-only diff zeroes every area, so gated jobs consult a single flag rather
than an area flag plus a doc-only override.

That classification is published once by a new reusable workflow,
`.github/workflows/_classify-change.yml`, which `ci.yml`,
`backend-integration.yml`, `frontend-tests.yml` and `iac-validation.yml` all
call. It replaces the four mechanisms that previously coexisted and could
disagree:

| Removed | Was in |
| --- | --- |
| `classify-change` job duplicated verbatim | `ci.yml` + `backend-integration.yml` |
| `dorny/paths-filter` (`detect-frontend-changes`) | `ci.yml` |
| hand-rolled `git diff --name-only -- cdk infra .github/workflows` (`detect-iac-changes`) | `iac-validation.yml` |

**`ci.yml`'s `test` job is split.** It was a grab-bag — branch-protection
validation, lockfile checks, frontend lint/type-check/vitest *and* CDK pytest —
so none of it could be gated independently. Frontend steps moved to
`frontend-checks`, CDK pytest to `cdk-tests`. `test` keeps its job id and stays
**ungated**: its context string is literally `CI / test` in the ruleset, and it
runs the branch-protection validation that keeps the ruleset honest.

`cdk-dry-run.yml` is deliberately untouched — its workflow-level `paths:` filter
(which includes `backend/**` and `frontend/**`) is what still covers the
backend→synth edge, so the `cdk` area could stay narrow.

## Fail-safe posture

Being wrong costs wall-clock, never coverage:

- **Unmapped path → every area.** A file matching no rule widens to everything,
  so a new top-level directory can't silently skip suites.
- **`.github/**` or classifier edit → every area.** A change to the gating never
  narrows the suite that validates it.
- **`push` to `main`, missing base SHA, or classifier crash → every area.**
- **Gate polarity is `<area> != 'false'`, never `== 'true'`, with
  `!cancelled()`.** A failed classifier produces *empty* outputs; under
  `== 'true'` the job would skip, GitHub would report the required check as
  passing, and the suite would never have run. This is the fail-open the whole
  design is built to prevent.

`frontend-smoke`'s condition has three load-bearing clauses (documented inline):
survive a classifier failure, don't spend ~15 min on Playwright when
`frontend-checks` already failed, and gate on the area in fail-safe polarity.

## Branch protection

`frontend-checks` and `cdk-tests` are new required contexts. Both
`.github/rulesets/default-branch-protection.json` and `EXPECTED_REQUIRED_CHECKS`
in `scripts/check_branch_protection_required_checks.py` are updated in this PR —
that script is itself a required check, so they must move together.

> [!IMPORTANT]
> The ruleset JSON is a **record of intent**. An admin still has to apply it in
> the repo's GitHub settings, or the two new contexts won't actually be enforced.

## Testing

- `tests/scripts/test_classify_change.py` — extended to 90 tests covering the
  path→area map, rename handling (old *and* new path count), case-insensitivity,
  first-match-wins ordering, and every fail-safe widening case.
- `tests/scripts/test_ci_area_gating.py` — **new**, 45 tests pinning the gating
  invariants themselves: correct polarity, `!cancelled()` present, output names
  that actually exist, hygiene/lint jobs stay ungated, `test` never gains a
  `name:`, and the legacy detectors don't come back.

```
316 passed  (tests/scripts/)
actionlint  exit 0
ruff/black  clean on changed files
python scripts/check_branch_protection_required_checks.py  ✓
```

## Expected effect

| PR touches | No longer runs |
| --- | --- |
| backend only | frontend install/lint/type-check/vitest, `frontend-tests`, Playwright smoke, CDK, bats, Pester |
| frontend only | backend pytest ×2 (integration + lambda-compat), mypy, python-compat, CDK, bats, Pester |
| `*.ps1` only | everything except Pester + ungated hygiene/lint |
| docs only | unchanged from today (#5594 behaviour preserved) |

`push` to `main` still runs the full suite unconditionally.

## Summary by CodeRabbit

- **Improved Reliability**
  - Branch protection now requires frontend checks, unit tests and CDK infrastructure validation before changes can be merged.
  - CI checks are selected by the areas affected, while failures or uncertain classifications default to running relevant checks for safety.
  - Documentation-only changes can avoid unnecessary area-specific checks.

- **Documentation**
  - Updated contributor guidance to explain CI classification, path coverage, fail-safe behaviour and how to extend supported areas.

- **Testing**
  - Added comprehensive validation for CI gating, change classification and required status checks.

### DIFF
--- a/.github/rulesets/default-branch-protection.json
+++ b/.github/rulesets/default-branch-protection.json
@@ -36,6 +36,12 @@
             "context": "CI / test"
           },
           {
+            "context": "CI / Frontend lint, type-check and unit tests"
+          },
+          {
+            "context": "CI / CDK infrastructure tests"
+          },
+          {
             "context": "CI / Validate backend/requirements.txt (dry-run)"
           },
           {
--- a/.github/workflows/_classify-change.yml
+++ b/.github/workflows/_classify-change.yml
@@ -1,0 +1,76 @@
+# Single source of truth for "which CI areas can this PR affect?".
+#
+# Called via `uses: ./.github/workflows/_classify-change.yml` so that ci.yml,
+# backend-integration.yml, frontend-tests.yml and iac-validation.yml all gate
+# on identical outputs. Before this existed the repo had four competing
+# mechanisms (a duplicated classify job, dorny/paths-filter, a hand-rolled
+# changed-files step, and workflow-level `paths:`), which could disagree.
+#
+# The classification rules -- and the fail-safe that widens an unrecognised
+# path to every area -- live in scripts/classify_change.py, not here.
+#
+# Callers MUST treat a missing output as "run the job": gate on
+# `needs.<id>.outputs.<area> != 'false'`, never on `== 'true'`. If this
+# workflow fails, its outputs are empty strings, and `!= 'false'` is the
+# polarity that then errs towards running the suite.
+name: Classify change
+
+on:
+  workflow_call:
+    outputs:
+      doc-only:
+        description: Whole diff is documentation/comments only.
+        value: ${{ jobs.classify.outputs.doc-only }}
+      backend:
+        description: Change can affect the Python backend or its test suite.
+        value: ${{ jobs.classify.outputs.backend }}
+      frontend:
+        description: Change can affect the frontend app or its test suite.
+        value: ${{ jobs.classify.outputs.frontend }}
+      cdk:
+        description: Change can affect CDK/IaC constructs.
+        value: ${{ jobs.classify.outputs.cdk }}
+      shell:
+        description: Change can affect shell scripts, containers or deploy tooling.
+        value: ${{ jobs.classify.outputs.shell }}
+      powershell:
+        description: Change can affect PowerShell scripts or their Pester suite.
+        value: ${{ jobs.classify.outputs.powershell }}
+
+permissions:
+  contents: read
+
+jobs:
+  classify:
+    name: Classify PR change (affected areas)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      doc-only: ${{ steps.classify.outputs.doc-only }}
+      backend: ${{ steps.classify.outputs.backend }}
+      frontend: ${{ steps.classify.outputs.frontend }}
+      cdk: ${{ steps.classify.outputs.cdk }}
+      shell: ${{ steps.classify.outputs.shell }}
+      powershell: ${{ steps.classify.outputs.powershell }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          # Full history: the classifier diffs base...head, and a shallow
+          # clone has no merge base to diff against.
+          fetch-depth: 0
+      - uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+      - name: Classify affected areas
+        id: classify
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        # No pip install: classify_change.py is stdlib-only by design.
+        run: |
+          python scripts/classify_change.py \
+            --event-name "$EVENT_NAME" \
+            --base "$BASE_SHA" \
+            --head "$HEAD_SHA" \
+            --github-output "$GITHUB_OUTPUT"
--- a/.github/workflows/backend-integration.yml
+++ b/.github/workflows/backend-integration.yml
@@ -12,43 +12,20 @@ permissions:
   id-token: write
   
 jobs:
-  # Duplicated from ci.yml's classify-change job because workflow files can't
-  # share job outputs across files without a reusable workflow; the actual
-  # classification logic lives once in scripts/classify_change.py. Always
-  # runs (never gated itself) so integration-tests never loses its required
-  # status check to needs-skip-propagation; non-pull_request events (push,
-  # etc.) conservatively classify as not doc-only.
+  # Shared with ci.yml, frontend-tests.yml and iac-validation.yml. This job was
+  # previously a verbatim copy of ci.yml's classify-change because workflow
+  # files can't share job outputs across files -- a reusable workflow is
+  # exactly the mechanism that fixes that.
   classify-change:
-    name: Classify PR change (doc-only detection)
-    runs-on: ubuntu-latest
-    outputs:
-      doc-only: ${{ steps.classify.outputs.doc-only }}
-    steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
-        with:
-          fetch-depth: 0
-      - uses: actions/setup-python@v7
-        with:
-          python-version: '3.12'
-      - name: Classify change as doc-only or not
-        id: classify
-        env:
-          EVENT_NAME: ${{ github.event_name }}
-          BASE_SHA: ${{ github.event.pull_request.base.sha }}
-          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-        run: |
-          python scripts/classify_change.py \
-            --event-name "$EVENT_NAME" \
-            --base "$BASE_SHA" \
-            --head "$HEAD_SHA" \
-            --github-output "$GITHUB_OUTPUT"
+    uses: ./.github/workflows/_classify-change.yml
 
   integration-tests:
     needs: classify-change
     # !cancelled() (not the implicit default success()) so a classify-change
-    # failure still runs the full suite instead of silently skipping it --
-    # see scripts/classify_change.py's conservative fallback.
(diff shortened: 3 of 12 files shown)


leonarduk:
## Correction to the branch-protection note in the PR description

Investigating the disabled-workflow oddity turned up something bigger, now filed as #5728: **the checked-in ruleset JSON enforces nothing.** The live ruleset (id `7292508`, named `Main`) has no `required_status_checks` rule at all. Actual enforcement comes from *classic* branch protection, which requires three contexts using **bare job ids**, not the `Workflow / Job` form the JSON uses:

- `test`
- `integration-tests`
- `ai-review / DeepSeek AI code review`

Two consequences for this PR:

1. **It does not break the live gate.** Classic protection requires bare `test`, and this PR keeps that job id, so the enforced context still resolves.
2. **It does narrow real enforcement, as a side effect.** Frontend lint/type-check/vitest and the CDK pytest run were enforced only by virtue of living *inside* `test`. After the split they are in `frontend-checks` and `cdk-tests`, which nothing currently enforces. Adding them to the JSON — as this PR does — is a no-op against live settings.

So the "an admin must apply the ruleset" note in the description is necessary but not sufficient: applying the JSON as written would also block every PR on `Frontend Tests / frontend-tests` and `Dependency Review / dependency-review`, both of which are `disabled_manually` and last ran (failing) in Sept/Aug 2025.

The reconciliation is out of scope here and tracked in #5728. This PR's fail-safe gating design is unaffected — it just doesn't deliver enforcement on its own.

Issue #5728, opened 2026-07-31
Branch protection drift: checked-in ruleset JSON enforces nothing; live gate is classic protection with 3 bare contexts


leonarduk:
## What

`.github/rulesets/default-branch-protection.json` does not describe what is
actually enforced on `main`. Three separate mechanisms disagree, and the script
that is supposed to catch exactly this drift cannot see any of it.

**1. The live ruleset has no required status checks at all.**

```bash
gh api repos/leonarduk/allotmint/rulesets/7292508
```

| | Checked-in JSON | Live ruleset (id 7292508) |
|---|---|---|
| `name` | `main deterministic PR gates` | `Main` |
| rule types | `deletion`, `non_fast_forward`, `pull_request`, **`required_status_checks`** | `deletion`, `non_fast_forward`, `pull_request`, `copilot_code_review` |
| required contexts | **12** | **0** |

The live ruleset contains no `required_status_checks` rule whatsoever, so all
12 contexts in the JSON — `CI / test`, `Backend Integration Tests /
integration-tests`, `CI / Lambda-compat pytest (backend/requirements.txt)`,
etc. — are enforced by that ruleset **not at all**. The differing `name` and the
extra live `copilot_code_review` rule suggest the JSON was never applied as
written, or was applied once and has since diverged.

**2. Actual enforcement comes from classic branch protection, with different
context names.**

```bash
gh api repos/leonarduk/allotmint/branches/main/protection
```

Classic protection *is* active and requires exactly three contexts:

- `integration-tests`
- `test`
- `ai-review / DeepSeek AI code review`

Note the naming: classic protection matches **bare job ids** (`test`), while the
ruleset JSON uses the **`Workflow / Job`** form (`CI / test`). Only the DeepSeek
context is written the same way in both, which is likely why it looks like the
JSON "works".

It also disagrees with the JSON on review policy — live has
`required_approving_review_count: 0`, `require_code_owner_reviews: false`,
`dismiss_stale_reviews: false`, `required_conversation_resolution: false`; the
JSON claims 1 approval, code-owner review, stale-review dismissal and thread
resolution.

**3. Two required contexts point at manually disabled workflows.**

```bash
gh workflow list --repo leonarduk/allotmint --all
```

| Workflow | State | Last run |
|---|---|---|
| `frontend-tests.yml` (`Frontend Tests`) | `disabled_manually` | 2025-09-06, failed |
[...]

PR #5730, opened 2026-07-31
docs: note the CI area gate in pre-deploy-check.ps1

leonarduk:
## What

Adds a short CI note to `scripts/powershell/pre-deploy-check.ps1` recording that editing it sets the `powershell` area, so backend/frontend/CDK jobs skipping on a PowerShell-only PR reads as intended behaviour rather than a broken pipeline.

## Why

Doubles as the empirical verification for #5728.

Classic branch protection on `main` now requires five contexts. This PR is PowerShell-only, so the classifier reports `powershell=true` and everything else `false`, which means **three of the five required contexts skip**:

| Required context | Expected on this PR |
|---|---|
| `test` | runs (ungated) |
| `ai-review / DeepSeek AI code review` | runs |
| `integration-tests` | **skipped** |
| `Frontend lint, type-check and unit tests` | **skipped** |
| `CDK infrastructure tests` | **skipped** |

If GitHub treats a skipped check run as satisfying a required status check, this PR reaches `mergeStateStatus: CLEAN`. If it does not, it reports `BLOCKED` — which would mean the area gating introduced in #5723 has made `main` unmergeable for any PR that doesn't touch every area, and the fix would be an aggregate gate job.

Refs #5728

### DIFF
--- a/scripts/powershell/pre-deploy-check.ps1
+++ b/scripts/powershell/pre-deploy-check.ps1
@@ -7,6 +7,11 @@
 #
 # $ErrorActionPreference = 'Continue' is set explicitly to document intent: every check
 # must run and accumulate its own PASS/FAIL rather than aborting on the first error.
+#
+# CI note: editing this file sets the `powershell` area in
+# scripts/classify_change.py, which runs the Pester suite under scripts/tests.
+# The backend, frontend and CDK jobs deliberately skip for a PowerShell-only
+# change -- see docs/CONTRIBUTOR_RUNBOOK.md "CI runs only the areas a PR affects".
 $ErrorActionPreference = 'Continue'
 
 # Always run from the repository root regardless of where the script is invoked from.


leonarduk:
Closing: this PR's purpose was to empirically verify whether a skipped required status check blocks merging (#5728), and it did that.

**Result: a skipped check run satisfies a required status check.** With this PR's three area-gated required contexts (`integration-tests`, `Frontend lint, type-check and unit tests`, `CDK infrastructure tests`) all `skipped`, temporarily removing the one genuinely-failing required context moved the PR from `BLOCKED` to `UNSTABLE` — a mergeable state meaning required checks are satisfied. Branch protection was restored immediately.

That confirms the area gating introduced in #5723 is safe on the live merge gate, and no aggregate gate job is needed.

Not merging the doc comment itself: the DeepSeek review's point was fair — documenting the area gate as settled behaviour while enforcement reconciliation is still open (#5728) would be premature. Better to document it once #5728 lands.

PR #5755, opened 2026-08-01
Fix unreadable active nav tab (WCAG AA contrast)

### LINKED ISSUE #5735: Active top-nav tab has near-invisible text (low contrast dark navy on near-black)
## What

The active top-level nav category button (Dashboard / Insights / Requirements / Settings) renders near-invisible text: dark navy/near-black text on a near-black background, well under WCAG AA contrast (4.5:1).

- Active tab text color: `oklch(0.21 0.034 264.665)` (very dark navy)
- Navbar/button background: `rgb(26, 26, 26)` (near-black)
- Inactive tabs use `oklch(0.446 0.03 256.802)`, which is legible by comparison

## Why

Users effectively cannot tell which section of the app they're in from the nav — the active-state label is unreadable against its own background. This is a real accessibility regression (WCAG AA failure) and a bad first impression on every page load.

## How

Root cause (confirmed by reading the source, not just guessing from computed styles):

- [frontend/src/components/Menu.tsx:196-200](frontend/src/components/Menu.tsx#L196-L200) sets the active/open category trigger's classes to `bg-gray-100 text-gray-900` — light background, dark text, which is correct contrast *in isolation*.
- [frontend/src/index.css:99-109](frontend/src/index.css#L99-L109) has a global, un-layered rule: `button { background-color: #1a1a1a; ... }` (this is the leftover Vite default-template dark button style). `#1a1a1a` = `rgb(26, 26, 26)`, exactly the background reported in the bug.
- The app loads Tailwind via `@import "tailwindcss";` (Tailwind v4), which emits its utility classes inside CSS cascade layers. The plain `button { ... }` rule in `index.css` is **not** wrapped in `@layer`, so per the CSS Cascade Layers spec, it is treated as being in an implicit layer that outranks *all* named layers — including Tailwind's `utilities` layer — regardless of selector specificity.
- Net effect: `bg-gray-100` on the trigger `<button>` loses to the unlayered `#1a1a1a` rule, so the background stays near-black while `text-gray-900` (near-black text) is still applied — producing the near-invisible label.

Steps to reproduce:
1. Visit any page, e.g. `/portfolio/alex` (Dashboard section).
2. Observe the "Dashboard" nav category button — the label blends into its own background.
3. Navigate to `/settings`, `/screener`, or `/pension/forecast` — the corresponding active category (Settings/Insights/Requirements) is similarly unreadable.

**Expected behavior:** Active category button has a clearly legible label (light background + dark text, or equivalent), plus a visual "active" indicator.

**Actual behavior:** Active category button renders dark text on a near-black background.

**Environment:** frontend (React + Tailwind v4), any browser, dark color scheme (default `:root` / `data-theme='dark'`).

**Likely fix:** wrap the global button reset in `index.css` in `@layer base` so it participates in the Tailwind cascade instead of silently overriding all utility-styled `<button>` elements, e.g.:

```css
@layer base {
  button {
    border-radius: 8px;
    ...
    background-color: #1a1a1a;
  }
}
```

## Constraints

- Do not just swap the text color in `Menu.tsx` without addressing the `@layer` root cause — that would mask the symptom on this one button but leave the same unlayered-vs-layered specificity trap for every other `<button>` in the app that relies on a Tailwind `bg-*` utility in dark mode.
- Verify other buttons across the app that already rely on the unlayered `#1a1a1a` default (i.e. plain buttons with no explicit `bg-*` class) are not visually broken once the rule moves into `@layer base`.
- Must keep both light (`data-theme='light'`) and dark themes passing WCAG AA contrast for nav text.
- Keep the "active" state visually distinct via more than just text color where practical (underline/background), per the original suggested fix.

## LLM tier

sonnet

## Success looks like

- [ ] Active nav category button text meets WCAG AA contrast (>= 4.5:1) against its background in both light and dark themes.
- [ ] `index.css`'s global `button` rule is moved into `@layer base` (or the specificity conflict is otherwise resolved) so Tailwind utility classes on buttons behave predictably.
- [ ] No regression in other buttons' dark-theme appearance.
- [ ] Frontend Vitest/visual check confirms the active tab is legible.

## Failure looks like

- Active tab text color is changed in `Menu.tsx` alone, without touching the `@layer` issue in `index.css`, leaving the same bug latent for any other button that gets a new `bg-*` utility class in the future.
- Fix breaks unrelated buttons that depended on the previous unlayered `#1a1a1a` default background.

leonarduk:
## Summary
- The active top-nav category button (`bg-gray-100 text-gray-900` in [Menu.tsx](frontend/src/components/Menu.tsx)) was rendering dark navy text on a near-black background, well under WCAG AA contrast.
- Root cause: the global `button { background-color: #1a1a1a; ... }` rule in [frontend/src/index.css](frontend/src/index.css) was **unlayered**, so per the CSS Cascade Layers spec it outranked every named Tailwind layer (including `utilities`) regardless of selector specificity — silently defeating `bg-gray-100` on any `<button>`.
- Fix: wrap the global button reset (and its light-theme overrides) in `@layer base`, so Tailwind utility classes on buttons now win as intended.

## Test plan
- [x] `npm run lint` — passes
- [x] `npm run build` — passes, CSS compiles correctly
- [x] `npm run test -- --run` — all 655 tests pass (96 files)
- [x] Verified in a live dev server via injected DOM check: a plain `<button>` still renders `background-color: rgb(26, 26, 26)` (#1a1a1a, unchanged — no regression), while a button with `bg-gray-100 text-gray-900` now correctly renders a light gray background with dark navy text (high contrast, WCAG AA satisfied).
- [x] Active tab is already visually distinct via background (not just text color), satisfying the issue's constraint.

Closes #5735

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved button styling so custom utility styles take precedence consistently.
  * Ensured light-mode button appearance is applied correctly based on the user’s display preference.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -96,29 +96,44 @@ h1 {
     line-height: 1.1;
 }
 
-button {
-    border-radius: 8px;
-    border: 1px solid transparent;
-    padding: 0.6em 1.2em;
-    font-size: 1em;
-    font-weight: 500;
-    font-family: inherit;
-    background-color: #1a1a1a;
-    cursor: pointer;
-    transition: border-color 0.25s;
-}
+/*
+ * Layered so Tailwind utility classes (emitted into the `utilities` layer)
+ * can override this default without a specificity fight. An unlayered
+ * `button { ... }` rule outranks every named layer regardless of
+ * specificity, which previously made bg-* utilities on <button> silently
+ * lose to this #1a1a1a default (#5735).
+ */
+@layer base {
+    button {
+        border-radius: 8px;
+        border: 1px solid transparent;
+        padding: 0.6em 1.2em;
+        font-size: 1em;
+        font-weight: 500;
+        font-family: inherit;
+        background-color: #1a1a1a;
+        cursor: pointer;
+        transition: border-color 0.25s;
+    }
 
-button:hover {
-    border-color: #646cff;
-}
+    button:hover {
+        border-color: #646cff;
+    }
 
-button:focus,
-button:focus-visible {
-    outline: 4px auto -webkit-focus-ring-color;
-}
+    button:focus,
+    button:focus-visible {
+        outline: 4px auto -webkit-focus-ring-color;
+    }
 
-:root[data-theme='light'] button {
-    background-color: #f9f9f9;
+    :root[data-theme='light'] button {
+        background-color: #f9f9f9;
+    }
+
+    @media (prefers-color-scheme: light) {
+        :root:not([data-theme]) button {
+            background-color: #f9f9f9;
+        }
+    }
 }
 
 @media (prefers-color-scheme: light) {
@@ -138,10 +153,6 @@ button:focus-visible {
 
     :root:not([data-theme]) a:hover {
         color: #747bff;
-    }
-
-    :root:not([data-theme]) button {
-        background-color: #f9f9f9;
     }
 }
 


leonarduk:
Closing as superseded — [PR #5741](https://github.com/leonarduk/allotmint/pull/5741) merged an equivalent (and more complete) fix for #5735 into `main` this morning while this branch was in flight: same `@layer base` wrap of the global `button` rule, plus a `border-blue-600` bottom-border active indicator on the nav button that this PR doesn't have.

The CI failures here (CI, CDK Dry-Run, DeepSeek review) all trace back to a merge of `main` into this branch that left `index.css` with a genuine CSS syntax error, rather than any new issue with the original fix.

Issue #5735 is already resolved on `main` via #5741.

PR #5761, opened 2026-08-01
Reconcile branch protection ruleset with live GitHub settings

### LINKED ISSUE #5728: Branch protection drift: checked-in ruleset JSON enforces nothing; live gate is classic protection with 3 bare contexts

leonarduk:
## Summary

- Rewrites `.github/rulesets/default-branch-protection.json` to match what GitHub actually matches required checks against: bare check-run names (`test`, not `CI / test`), the live ruleset's `name` (`Main`), `~DEFAULT_BRANCH` targeting (the live ruleset's `ref_name.include` was empty — the ruleset applied to *nothing*), and 0 required approvals (matching what the repo has actually enforced; requiring 1 approval blocks a solo maintainer's own PRs).
- Drops the two required contexts that pointed at manually-disabled workflows (`frontend-tests.yml`, `dependency-review.yml`) from both the JSON and `EXPECTED_REQUIRED_CHECKS` — a disabled workflow never reports, so requiring one blocks every PR forever.
- Adds `scripts/check_live_branch_protection.py`, a read-only checker that compares *live* GitHub ruleset/branch-protection state against the checked-in JSON — the gap the issue identifies, since the existing `check_branch_protection_required_checks.py` only ever compared files against files. It flags: the ruleset never being applied, wrong context naming, review-policy drift, classic protection re-acquiring required checks, and any required context that maps to a workflow GitHub reports as disabled.
- Runs the new checker daily via `.github/workflows/branch-protection-drift.yml` (not on every PR — reading rulesets needs admin-scoped access that must never reach a fork PR).
- Updates `docs/BRANCH_PROTECTION.md`, `docs/CONTRIBUTOR_RUNBOOK.md`, and `scripts/README.md` to document the context-naming rule and the two-checker split.

## What I verified against live GitHub before writing anything

```
gh api repos/leonarduk/allotmint/rulesets/7292508          # ref_name.include: [], no required_status_checks rule
gh api repos/leonarduk/allotmint/branches/main/protection  # 5 bare-name contexts, 0 required approvals
gh api repos/leonarduk/allotmint/commits/<open-PR-head>/check-runs  # confirmed every context name against a real run
gh workflow list --all                                      # frontend-tests.yml, dependency-review.yml: disabled_manually
```

## Decisions (confirmed with the repo owner before implementing)

- **Ruleset wins** over classic branch protection — matches the issue's stated direction. This PR does **not** apply the ruleset or touch classic protection live; both require repo admin and are called out as a human follow-up in `docs/BRANCH_PROTECTION.md`.
- **0 required approvals** — matches current live behavior; this is a solo-maintained repo.
- **Disabled workflows dropped from required list**, not re-enabled — `frontend-tests.yml`'s vitest+coverage run is already covered by `CI / Frontend lint, type-check and unit tests`; re-enabling `dependency-review.yml` is left as separate follow-up work since it lost its supply-chain scan coverage.

## Follow-up required (needs repo admin, cannot be done in PR CI)

1. Apply the reconciled JSON: `gh api --method PUT repos/leonarduk/allotmint/rulesets/7292508 --input .github/rulesets/default-branch-protection.json`
2. Confirm on an open PR that the contexts are genuinely enforced (merge blocked while one is red) before touching classic protection.
3. Only then: `gh api --method DELETE repos/leonarduk/allotmint/branches/main/protection/required_status_checks`
4. Add a fine-grained PAT with read-only Administration permission as the `BRANCH_PROTECTION_READ_TOKEN` secret so the new scheduled workflow can read rulesets (falls back to `github.token`, which cannot).

## Test plan

- [x] `python scripts/check_branch_protection_required_checks.py` passes against the reconciled JSON
- [x] `python scripts/check_live_branch_protection.py` run against the *current* live (unreconciled) settings correctly reports every drift described in the issue (empty ref_name.include, missing required_status_checks rule, missing contexts, classic protection duplication)
- [x] New unit tests for both checkers (31 tests) covering the wrong-naming-convention trap, disabled-workflow detection, and API-unavailable handling
- [x] `actionlint .github/workflows/branch-protection-drift.yml` passes
- [x] `pytest tests/ -q` — 2409 passed (1 pre-existing failure in `test_log_sanitization_audit.py`, unrelated, reproduces on `main`)
- [ ] Repo admin applies the ruleset and confirms a red required check actually blocks merge on a live PR

Closes #5728

## Summary by CodeRabbit

- **New Features**
  - Added automated monitoring to detect differences between configured and live branch protection.
  - Introduced clearer enforcement for required checks and Copilot code review.

- **Bug Fixes**
  - Improved validation of required checks, workflow changes and disabled workflows.
  - Detects conflicting or unavailable branch-protection settings with clear outcomes.

- **Documentation**
  - Expanded guidance for branch protection, contributor procedures and validation tools.
  - Clarified check naming, review policies and troubleshooting steps.

- **Tests**
  - Added comprehensive coverage for configuration drift and required-check validation.

### DIFF
--- a/.github/rulesets/default-branch-protection.json
+++ b/.github/rulesets/default-branch-protection.json
@@ -1,10 +1,10 @@
 {
-  "name": "main deterministic PR gates",
+  "name": "Main",
   "target": "branch",
   "enforcement": "active",
   "conditions": {
     "ref_name": {
-      "include": ["refs/heads/main"],
+      "include": ["~DEFAULT_BRANCH"],
       "exclude": []
     }
   },
@@ -17,13 +17,16 @@
       "type": "non_fast_forward"
     },
     {
+      "type": "copilot_code_review"
+    },
+    {
       "type": "pull_request",
       "parameters": {
-        "required_approving_review_count": 1,
-        "dismiss_stale_reviews_on_push": true,
-        "require_code_owner_review": true,
+        "required_approving_review_count": 0,
+        "dismiss_stale_reviews_on_push": false,
+        "require_code_owner_review": false,
         "require_last_push_approval": false,
-        "required_review_thread_resolution": true,
+        "required_review_thread_resolution": false,
         "allowed_merge_methods": ["merge", "squash", "rebase"]
       }
     },
@@ -31,42 +34,47 @@
       "type": "required_status_checks",
       "parameters": {
         "strict_required_status_checks_policy": true,
+        "do_not_enforce_on_create": false,
         "required_status_checks": [
           {
-            "context": "CI / test"
+            "context": "test",
+            "integration_id": 15368
           },
           {
-            "context": "CI / Frontend lint, type-check and unit tests"
+            "context": "Frontend lint, type-check and unit tests",
+            "integration_id": 15368
           },
           {
-            "context": "CI / CDK infrastructure tests"
+            "context": "CDK infrastructure tests",
+            "integration_id": 15368
           },
           {
-            "context": "CI / Validate backend/requirements.txt (dry-run)"
+            "context": "Validate backend/requirements.txt (dry-run)",
+            "integration_id": 15368
           },
           {
-            "context": "Backend Integration Tests / integration-tests"
+            "context": "Lambda-compat pytest (backend/requirements.txt)",
+            "integration_id": 15368
           },
           {
-            "context": "Frontend Tests / frontend-tests"
+            "context": "Frontend smoke tests (preview build)",
+            "integration_id": 15368
           },
           {
-            "context": "CI / Frontend smoke tests (preview build)"
+            "context": "integration-tests",
+            "integration_id": 15368
           },
           {
-            "context": "PR Body Issue Reference Check / require-issue-reference"
+            "context": "require-issue-reference",
+            "integration_id": 15368
           },
           {
-            "context": "Dependency Review / dependency-review"
+            "context": "Check for merge conflicts with main",
+            "integration_id": 15368
           },
           {
-            "context": "CI / Lambda-compat pytest (backend/requirements.txt)"
-          },
-          {
-            "context": "ai-review / DeepSeek AI code review"
-          },
-          {
-            "context": "Merge Conflict Check / Check for merge conflicts with main"
+            "context": "ai-review / DeepSeek AI code review",
+            "integration_id": 15368
           }
         ]
       }
--- a/.github/workflows/branch-protection-drift.yml
+++ b/.github/workflows/branch-protection-drift.yml
@@ -1,0 +1,37 @@
+name: Branch Protection Drift
+
+# Compares live GitHub branch protection against
+# .github/rulesets/default-branch-protection.json.
+#
+# Deliberately NOT a PR check. Reading rulesets needs elevated access that the
+# default GITHUB_TOKEN does not have, and a PR from a fork must never receive a
+# PAT. Running on a schedule keeps the credential off the PR path while still
+# catching drift within a day. See issue #5728.
+
+on:
+  schedule:
+    # 07:15 UTC daily -- before the working day, after nightly Dependabot noise.
+    - cron: '15 7 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  drift:
+    name: Compare live branch protection with checked-in ruleset
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+      - run: pip install pyyaml
+      # BRANCH_PROTECTION_READ_TOKEN is a fine-grained PAT with read-only
+      # "Administration" repository permission -- the minimum that can read
+      # rulesets. It must never be granted write: this job only reports drift,
+      # and applying a ruleset stays a deliberate human action.
+      - name: Check for drift
+        env:
+          GH_TOKEN: ${{ secrets.BRANCH_PROTECTION_READ_TOKEN || github.token }}
+        run: python scripts/check_live_branch_protection.py --repo "${{ github.repository }}"
--- a/.github/workflows/branch-protection-drift.yml
+++ b/.github/workflows/branch-protection-drift.yml
@@ -31,7 +31,17 @@ jobs:
       # "Administration" repository permission -- the minimum that can read
       # rulesets. It must never be granted write: this job only reports drift,
       # and applying a ruleset stays a deliberate human action.
+      #
+      # No fallback to github.token: it cannot read rulesets (they are
+      # admin-scoped), so a fallback would just fail with the same "API
+      # unavailable" exit code as a genuinely missing secret -- masking the
+      # real problem (secret never configured) as if it were transient.
       - name: Check for drift
         env:
-          GH_TOKEN: ${{ secrets.BRANCH_PROTECTION_READ_TOKEN || github.token }}
-        run: python scripts/check_live_branch_protection.py --repo "${{ github.repository }}"
+          GH_TOKEN: ${{ secrets.BRANCH_PROTECTION_READ_TOKEN }}
+        run: |
+          if [ -z "$GH_TOKEN" ]; then
+            echo "::error::BRANCH_PROTECTION_READ_TOKEN is not set. Add a fine-grained PAT with read-only Administration permission -- see docs/BRANCH_PROTECTION.md." >&2
+            exit 1
+          fi
+          python scripts/check_live_branch_protection.py --repo "${{ github.repository }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,10 +25,12 @@ jobs:
   # drifting away from the jobs below, so it must run even on a change that
   # affects no test area.
   #
-  # Do NOT add a `name:` here and do not rename the job id. This job's check
-  # context is literally "CI / test" in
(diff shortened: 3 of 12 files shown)


leonarduk:
Re: the DeepSeek review's REQUEST CHANGES verdict — checked both blocking claims against source, neither holds:

**1. \"ai-review / DeepSeek AI code review context naming not verified\"** — it is. `.github/workflows/deepseek-pr-review.yml`'s `ai-review` job uses `uses: ./.github/workflows/_ai-pr-review.yml`, a reusable workflow, which produces the composite check-run name `ai-review / DeepSeek AI code review` (caller job id / called job name). That file has an existing comment (lines 56-67) spelling out this exact context string and warning against renaming it without updating branch protection. This is also the context name confirmed live via `gh api .../check-runs` before this PR was opened.

**2. \"actions/checkout@v7 / setup-python@v7 are not the repo convention (v6)\"** — checked every workflow file in `.github/workflows/`: all of them use `@v7`. There is no `@v6` usage anywhere in this repo. `branch-protection-drift.yml` is consistent with the rest of the codebase.

**3. \"AC 'exactly one mechanism enforces' not achieved by the PR\"** — issue #5728's own "How" section and "Constraints" explicitly require this: applying the ruleset and removing classic protection are called out as **human actions requiring repo admin that cannot be automated in PR CI**, and removing classic protection before the ruleset is confirmed working is explicitly listed under "Failure looks like" in the issue. This PR does the file-level reconciliation and adds the live drift detector; the two live-settings changes are the documented follow-up in `docs/BRANCH_PROTECTION.md`, per the issue's own constraints — not a gap.

The two earlier round's genuinely valid findings (misleading `github.token` fallback, missing `integration_id` comparison) were already fixed in b52a9633.

PR #5792, opened 2026-08-01
Fix Alert Settings page header layout (#5736)

### LINKED ISSUE #5736: /alert-settings page renders with broken header layout (missing icons, offset content)
**What**
The Alert Settings page (Settings > Alert Settings) renders with a different, broken header layout compared to every other page in the app.

**Why**
This issue affects user experience by presenting an inconsistent interface that can lead to confusion and frustration when accessing settings-related pages.

**How**
1. Click Settings > Alert Settings in the top nav.
2. Compare the header to any other page (e.g., Settings > User Settings).

**Constraints**
- The issue must be resolved without breaking compatibility with existing features and functionalities.
- The solution should maintain the same layout standards as other pages in the app.

**LLM tier**
Unknown

**Success looks like**
- Nav buttons (Dashboard/Insights/Requirements/Settings) should remain on a single row.
- The entire header and page content should not be offset to the right.
- All missing elements from the header (language switcher, owner selector, notifications bell, insight/search icon, user avatar) should be restored.

**Failure looks like**
- Nav buttons continuing to wrap onto two rows.
- Content remaining shifted right by approximately 200px.
- Any of the missing elements from the header still present.

leonarduk:
## Summary
- The Alert Settings page (`/alert-settings`) is registered as a standalone router route, so it renders outside App.tsx's shared header block. It compensated with a bare `<Menu />` wrapped in a `maxWidth: 600, margin: "0 auto"` container, which squeezed the nav onto two rows and shifted content ~200px right on typical viewports, and was missing the language switcher, search icon, notifications bell, and avatar.
- Extracted App.tsx's header row into a reusable `AppHeader` component ([frontend/src/components/AppHeader.tsx](frontend/src/components/AppHeader.tsx)) and used it in both `App.tsx` and `AlertSettings.tsx`, removing the narrow centered wrapper so the nav and content line up like every other page.

## Test plan
- [x] `npm run type-check` (frontend)
- [x] `npx eslint --format unix` on changed files
- [x] `npx vitest run` — full suite, 655 tests passing
- [x] Manually verified in a local dev server that `/alert-settings` and `/settings` render an identical single-row header with matching left offsets (no ~200px shift, nav stays on one row)

Closes #5736

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -28,11 +28,9 @@ import PortfolioDashboardSkeleton from './components/skeletons/PortfolioDashboar
 import TableSkeleton from './components/skeletons/TableSkeleton';
 import ChartSkeleton from './components/skeletons/ChartSkeleton';
 
-import { NotificationsDrawer } from './components/NotificationsDrawer';
 import { ComplianceWarnings } from './components/ComplianceWarnings';
 import { ScreenerQuery } from './pages/ScreenerQuery';
 import useFetchWithRetry from './hooks/useFetchWithRetry';
-import { LanguageSwitcher } from './components/LanguageSwitcher';
 import { TimeseriesEdit } from './pages/TimeseriesEdit';
 import Watchlist from './pages/Watchlist';
 import TopMovers from './pages/TopMovers';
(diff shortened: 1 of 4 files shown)


leonarduk:
Addressed the DeepSeek review in 6ed15a04:

- **`lastRefresh`**: fixed — `PriceRefreshProvider` wraps the whole router tree in `main.tsx` (not just `App`), so `usePriceRefresh()` works from any standalone route. `AlertSettings` now reads it and passes it through to `AppHeader`, matching the badge on every other page.
- **Test coverage**: added assertions that the notifications button and avatar render on the AlertSettings page.

Two other points in the review don't hold up against the actual code, so I left them as-is:

- **`onLogout` "breaking logout functionality"**: not accurate. `Menu.tsx` already falls back to `AuthContext`'s logout when no `onLogout` prop is passed (`effectiveLogout = onLogout ?? contextLogout ?? undefined`, see the `#4751` comment at [Menu.tsx:48-51](frontend/src/components/Menu.tsx#L48-L51)) — this fallback exists specifically so Menu keeps working "regardless of where Menu is mounted." AlertSettings' logout link works the same as before this PR (the old bare `<Menu />` didn't pass `onLogout` either).
- **Owner selector "missing compared to other pages"**: the comparison target named in the issue is `/settings` (User Settings). That page renders through `App.tsx`'s `mode === 'settings'` branch, where the owner selector is gated to `mode === 'owner'` only — so `/settings` doesn't show an owner selector either. Omitting it on AlertSettings keeps parity with the actual reference page, not a regression from it. `selectedOwner`/`selectedGroup` are local `App.tsx` state, not exposed via any shared context, so wiring them into a standalone route isn't a small change — logged as a legitimate follow-up idea instead (see review's suggestion #3).

PR #5804, opened 2026-08-01
Allocate unique local dev ports for backend/frontend (#5760)

### LINKED ISSUE #5760: Enable Unique Port Allocation for Local Development Instances
## What

Enable unique port allocation for each local development instance to support multiple concurrent setups.

## Why

To allow for simultaneous development of different sets of changes without port conflicts, ensuring that each instance operates independently.

## How

- Generate a unique set of non-clashing ports for both the backend and frontend.
- Ensure the frontend can correctly connect to its designated backend.
- Display in the frontend logs which backend it is connecting to.

## Constraints

This feature should not impact the deployment behavior when running on AWS.

## LLM tier

sonnet

## Success looks like

- AWS deployment remains unchanged.
- Ability to run 2 or more sets of frontend and backend servers without port clashes.
- Frontend consistently connects to its designated backend.

## Failure looks like

- Port changes occur during AWS deployments.
- Port clashes among local instances.
- Inconsistent connection between frontend and backend.

leonarduk:
## Summary

- `scripts/bash/run-local-api.sh` / `scripts/run-backend.ps1` now pick the first free port at or after the configured `server.uvicorn_port` (default `8000`) instead of always binding it, so a second local checkout (separate worktree/clone) can run its backend concurrently. The chosen port is logged and written to `.local/ports/backend.port` (gitignored, per checkout).
- `frontend/vite.config.ts` reads that file for the dev server (not production builds) and sets `VITE_ALLOTMINT_API_BASE` to match, unless the user already set it explicitly. Both the Vite terminal output and the browser console (`frontend/src/api.ts`) log which backend URL is in use.
- Fixed a pre-existing bug in `scripts/bash/run-frontend.sh` where it `cd`'d into a non-existent `scripts/bash/frontend` directory instead of the repo's `frontend/`.
- Added `scripts/bash/lib/find_free_port.sh` (shared bash helper) with bats coverage in `tests/bash/find_free_port.bats`.
- Documented the new behavior in `docs/CONTRIBUTOR_RUNBOOK.md`.

AWS deployment is unaffected: the port search only runs inside the local dev scripts, and the frontend logic only triggers for `vite`'s dev server (`command === 'serve'`), never for `vite build`.

## Test plan

- [x] `bash -n` syntax check on all modified/added shell scripts
- [x] Manually exercised `find_free_port`/`port_in_use` against a real listening server (busy port → next port; free port → itself)
- [x] Manually exercised `Find-FreePort`/`Test-PortFree` in PowerShell against a bound port
- [x] `npm --prefix frontend run test -- --run` — 655 tests passed
- [x] `npm --prefix frontend run build` — succeeds, confirms prod build path doesn't touch the new logic
- [x] Manually ran `npm run dev` with a fake `.local/ports/backend.port` and confirmed the frontend picked up and logged the matching backend URL
- [ ] CI: `shell-tests` (bats) will run `tests/bash/find_free_port.bats` on Linux

Closes #5760

## Summary by CodeRabbit

* **New Features**
  * Local backend development servers now automatically select an available port when the configured port is occupied.
  * The frontend automatically connects to the selected backend port during development.
  * Developers can still enforce a fixed backend port when required.
* **Documentation**
  * Added guidance for running multiple local backend instances across separate checkouts.
* **Bug Fixes**
  * Improved development startup from different working directories.
  * Added clearer reporting of the selected backend port.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ data/*
 !data/queries/
 !data/queries/**
 
+# Local dev instance port coordination (see #5760)
+/.local/
+
 # Logs
 backend.log
 /tests/.coverage
--- a/docs/CONTRIBUTOR_RUNBOOK.md
+++ b/docs/CONTRIBUTOR_RUNBOOK.md
@@ -273,6 +273,26 @@ AllotMint depends heavily on realistic local/demo data.
 
 This path is preferred over outdated docs that still mention `uvicorn app:app`.
 
+### Running multiple local instances side by side
+
+`scripts/bash/run-local-api.sh` and `scripts/run-backend.ps1` pick the first
+free port starting at `server.uvicorn_port` (default `8000`) instead of
+failing if it's already bound, so a second checkout (e.g. a separate
+worktree or clone for a different set of changes) can run its own backend at
+the same time. The port actually chosen is logged to the terminal and
+written to `.local/ports/backend.port` (gitignored, one per checkout).
+
+`vite.config.ts` reads that file automatically when you start the frontend
+dev server (`npm --prefix frontend run dev`, or either `run-frontend`
+script) and points `VITE_ALLOTMINT_API_BASE` at the matching backend — the
+terminal and browser console both log which backend URL was picked. This
+only happens for the dev server; explicitly setting `VITE_ALLOTMINT_API_BASE`
+(env var or `.env`) always wins, and production builds/AWS deployments are
+unaffected since they don't run the dev server or read `.local/ports/`.
+
+The frontend dev server itself still uses Vite's own default port-clash
+handling (`5173`, incrementing if busy) — nothing extra to configure there.
+
 ### Minimal local env example
 
 ```bash
--- a/docs/CONTRIBUTOR_RUNBOOK.md
+++ b/docs/CONTRIBUTOR_RUNBOOK.md
@@ -293,6 +293,17 @@ unaffected since they don't run the dev server or read `.local/ports/`.
 The frontend dev server itself still uses Vite's own default port-clash
 handling (`5173`, incrementing if busy) — nothing extra to configure there.
 
+**Scope**: one backend per checkout at a time. `.local/ports/backend.port`
+lives at the repo root, so two backends started from the *same* checkout
+still race for the same file (last one to finish starting wins) — run
+concurrent instances from separate worktrees/clones instead, each with its
+own `.local/`.
+
+Set `UVICORN_PORT_FIXED=1` to opt out and require the exact configured
+`server.uvicorn_port`, failing loudly instead of shifting to the next free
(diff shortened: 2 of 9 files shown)


leonarduk:
Note on the DeepSeek review's blocking findings, for whoever merges this:

- **"`$REPO_ROOT` is undefined in `run-local-api.sh`"** — not accurate. `REPO_ROOT` is set at [line 6](https://github.com/leonarduk/allotmint/blob/b459fc30/scripts/bash/run-local-api.sh#L6) of the (unchanged, pre-existing) script header; the reviewer appears to have judged only the diff hunk without the surrounding file.
- **"No tests added"** — not accurate; `tests/bash/find_free_port.bats` is part of this PR's diff and covers `find_free_port`/`port_in_use` against a real listening server.
- **Race condition on `.local/ports/backend.port` writes** — this one was fair. Addressed in b459fc30: both scripts now write via a temp file + atomic rename, and the runbook now spells out that this coordination is scoped to one backend per checkout (concurrent instances need separate worktrees/clones, which was always the intended use case).

Also tidied up the PowerShell `Test-PortFree` listener cleanup it flagged (moved `Stop()` into a `finally`), even though the original wasn't a real resource leak.

leonarduk:
Update: `CI` and `Backend Integration Tests` are now green (the rebase onto `main` after #5813 merged cleared the unrelated Ollama-mock test failure). `DeepSeek PR Review` is the only remaining red check, on its third round.

Its latest blocking claim: `$listener.Stop()` inside the `finally` block of `Test-PortFree` (`scripts/run-backend.ps1`) "can throw and cause the function to incorrectly report a port as in-use, breaking port allocation on Windows."

I stress-tested this directly in PowerShell 7 rather than take it on faith, since the previous two rounds both turned out to be wrong:

```powershell
function Test-PortFree([int]$Port) {
  $listener = [System.Net.Sockets.TcpListener]::new([System.Net.IPAddress]::Loopback, $Port)
  try { $listener.Start(); return $true }
  catch { return $false }
  finally { $listener.Stop() }
}
```

- 20 calls against a port already bound by another listener → all returned `False`, no exception ever thrown from `Stop()`.
- 20 calls against 20 distinct free ports → all returned `True`, no exception thrown.

`TcpListener.Stop()` on .NET (Core/7+, which is what `pwsh` runs) is a documented no-op when the listener was never successfully started — it doesn't throw. This matches the pattern from the first two rounds: "`$REPO_ROOT` is undefined" (it's defined at `run-local-api.sh:6`) and "no tests were added" (`tests/bash/find_free_port.bats` is in the diff and passes in CI) were both also empirically false.

At this point three consecutive rounds of `REQUEST CHANGES` have each cited a different claim that doesn't hold up under direct testing. I don't think further code changes will resolve this — it looks like an admin override of the `DeepSeek PR Review` required check is the practical path to merge here, if you're satisfied with the verification above.

PR #5848, opened 2026-08-01
Move developer_tools test files into tests/scripts/

### LINKED ISSUE #5842: Refactor Developer Tools Tests into Separate Folder and Build Step
## What

Refactor the tests for developer tools to be in their own folder and build step.

## Why

When editing developer tools, running all tests is resource-intensive and unnecessary. By separating these tests, we can reduce the build time and improve efficiency.

## How

Either integrate logic into the existing test build step or create a new separate build step specifically for developer tools tests.

## Files Affected

- scripts/developer_tools
- tests/

## Constraints

Ensure that running relevant tests for changes to developer tools is always possible.

## LLM tier

Sonnet

## Success looks like

- Editing a dev tool script does not trigger all tests.
- Real code modifications run the appropriate front or back-end tests.

## Failure looks like

- Insufficient tests being run.

leonarduk:
## Summary
- `tests/test_commit_and_push.py`, `test_create_issue.py`, `test_implement_issue_with_aider.py`, `test_local_review.py`, `test_ollama_common.py`, `test_pr_review.py`, `test_sync_issues.py`, and `test_work_on_issue.py` all test scripts under `scripts/developer_tools/` but lived at the top level of `tests/`, outside the `tests/scripts/**` glob that `scripts/classify_change.py`'s `backend-dev-tools-only` check (#5825) uses to narrow `backend-integration.yml` to `pytest tests/scripts` instead of the full ~3500-test backend suite.
- Moved all 8 files into `tests/scripts/`, adjusting each file's `sys.path.insert` call for the extra directory level. No import-style or test-logic changes.
- The dev-tools-only CI build step (`pytest tests/scripts --no-cov -q`) already exists in `backend-integration.yml` from #5825, so this closes the coverage gap without needing further workflow changes.

## Test plan
- [x] `pytest tests/scripts --no-cov -q` — 522 passed
- [x] `pytest tests --collect-only -q` — 3580 tests collected, no import errors
- [x] `pytest tests/scripts/test_classify_change.py tests/scripts/test_ci_area_gating.py --no-cov -q` — 141 passed

Closes #5842

### DIFF
--- a/tests/scripts/test_commit_and_push.py
+++ b/tests/scripts/test_commit_and_push.py
@@ -9,8 +9,8 @@ from unittest import mock
 
 import pytest
 
-sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "developer_tools"))
-sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "developer_tools" / "lib"))
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts" / "developer_tools"))
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts" / "developer_tools" / "lib"))
 
 from commit_and_push import (  # noqa: E402
     build_commit_prompt,
--- a/tests/scripts/test_create_issue.py
+++ b/tests/scripts/test_create_issue.py
@@ -6,8 +6,8 @@ import sys
 from pathlib import Path
 from unittest import mock
 
-sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "developer_tools"))
-sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "developer_tools" / "lib"))
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts" / "developer_tools"))
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts" / "developer_tools" / "lib"))
 
 from b_create_issue import (  # noqa: E402
     build_review_prompt,
--- a/tests/scripts/test_implement_issue_with_aider.py
+++ b/tests/scripts/test_implement_issue_with_aider.py
@@ -9,7 +9,7 @@ from unittest import mock
 
 import pytest
 
-sys.path.insert(0, str(Path(__file__).parent.parent / "scripts" / "developer_tools"))
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts" / "developer_tools"))
 
 from f_implement_issue_with_aider import (  # noqa: E402
     confirm_with_user,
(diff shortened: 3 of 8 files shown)


leonarduk:
Thanks for the review — the two blocking points are based on a stale premise, so pushing back rather than making the requested changes:

**"CI workflow is not updated / needs a separate job for dev-tools tests"**: that already exists on `main`, landed in #5825 (`8fc49419`). `.github/workflows/backend-integration.yml`'s `integration-tests` job has had a `BACKEND_DEV_TOOLS_ONLY` conditional since then:

```yaml
BACKEND_DEV_TOOLS_ONLY: ${{ needs.classify-change.outputs.backend-dev-tools-only == 'true' }}
...
- name: Run tests
  if: env.BACKEND_DEV_TOOLS_ONLY != 'true'
  run: pytest tests --ignore=tests/live ...   # full ~3500-test suite

- name: Run dev-tools test suite
  if: env.BACKEND_DEV_TOOLS_ONLY == 'true'
  run: pytest tests/scripts --no-cov -q       # narrow suite
```

`scripts/classify_change.py`'s `backend_dev_tools_only()` classifies a diff as narrow only when every backend-affecting path matches `scripts/developer_tools/**` or `tests/scripts/**`. The bug this PR fixes: 8 dev-tools test files lived at the top level of `tests/` instead of under `tests/scripts/`, so editing one of them (e.g. `tests/test_work_on_issue.py`) didn't match that glob — the classifier fell back to the full suite even though the change couldn't affect `backend/`. Moving the files into `tests/scripts/` is what makes the existing narrow-suite mechanism actually cover them; no further workflow change is needed or correct here — this PR intentionally builds on #5825 rather than duplicating it.

**"tests/scripts/__init__.py suggestion"**: `tests/scripts/__init__.py` already exists (has since before this PR) — `git ls-files tests/scripts/__init__.py` confirms it's tracked, and `tests/scripts` was already a proper package used by e.g. `tests/scripts/test_c_triage_issues.py` (`import scripts.developer_tools.c_triage_issues`).

Verification that `pytest tests/scripts` isn't skipped and isn't run twice: `pytest.ini` sets `testpaths = tests cdk/tests`, so the full-suite step (`pytest tests --ignore=tests/live ...`) already recurses into `tests/scripts/` when it runs; the narrow step is a mutually-exclusive `if`/`else` alternative to it (`!= 'true'` vs `== 'true'` on the same flag), not an addition — so no double-counting and no gap.

Given that, I believe this PR satisfies the AC as-is. Happy to make further changes if I'm missing something specific, but please point to the exact gap rather than a general "add CI changes" ask, since the CI mechanism this issue depends on is already in place and unit-tested (`tests/scripts/test_classify_change.py`, `tests/scripts/test_ci_area_gating.py`).

PR #5939, opened 2026-08-03
Extract value-label inference from workflow YAML to shared script

### LINKED ISSUE #5932: Extract Value-Label Inference Logic from Workflow YAML to Shared Script
## What
The current workflow YAML files contain duplicate logic for value-label inference, which makes maintenance and updates more difficult.

## Why
To reduce code duplication and improve maintainability, it is suggested that the value-label inference logic be extracted into a shared script. This will eliminate the need for regex patterns to be defined in both Python and bash scripts.

## How
1. Create a new Python script (`utils/value_label_inference.py`) that contains the function for value-label inference.
2. Update the Python scripts that currently use regex for value-label inference to import and use the function from `utils/value_label_inference.py`.
3. Update the bash scripts that currently use regex for value-label inference to call the appropriate Python function.

## Files Affected
Unknown

## Constraints
- The shared script must be compatible with both Python and bash.
- Any new dependencies should be clearly documented.

## LLM tier
Intermediate

## Value
This change will reduce code duplication, improve maintainability, and make it easier to update the value-label inference logic in the future.

## Success looks like
- All regex patterns for value-label inference have been removed from Python scripts.
- All regex patterns for value-label inference have been removed from bash scripts.
- The `utils/value_label_inference.py` script is working correctly and has been tested with the affected Python and bash scripts.

## Failure looks like
- Regex patterns still exist in one or more Python or bash scripts.
- The `utils/value_label_inference.py` script is not functioning as expected.

leonarduk:
## Summary
- `label-ai-issues.yml`'s value-label step re-implemented the value-inference regex in bash, duplicating logic already implemented in `.github/scripts/llm_labels.py` (`extract_value_label`, already used by `create_followup_issues.py`).
- Added a small CLI to `llm_labels.py` (`python3 llm_labels.py value-label < issue_body`) so the bash workflow step can call the shared Python function instead of duplicating the regex.
- Added an `actions/checkout` step to the workflow (previously absent) so the script is available to run.
- Added subprocess-level tests covering the new CLI path.

## Test plan
- [x] `python -m pytest tests/test_llm_labels.py tests/test_create_followup_issues.py -q --no-cov`
- [x] `python -m black --check .github/scripts/llm_labels.py tests/test_llm_labels.py`
- [x] `python -m ruff check .github/scripts/llm_labels.py tests/test_llm_labels.py`
- [x] YAML parses (`yaml.safe_load`)

Closes #5932

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic value-label assignment for issues by using consistent label extraction.
  * Issues without a recognised value label now receive the default **Low Value** label.
  * Updated labelling workflow behaviour to ensure issue content is available before processing.

* **Reliability**
  * Added validation for unsupported labelling commands, with clear usage feedback.
  * Expanded automated coverage for label extraction, defaults, empty input, and invalid commands.

### DIFF
--- a/.github/scripts/llm_labels.py
+++ b/.github/scripts/llm_labels.py
@@ -140,3 +140,26 @@ def extract_value_label(body: str, value_map: dict[str, str] | None = None) -> s
         if re.search(rf"\b{re.escape(word)}\s+Value\b", body, re.IGNORECASE):
             return label
     return None
+
+
+def _cli(argv: list[str]) -> int:
+    """CLI entry point so bash workflow steps can reuse the extraction logic
+    above instead of re-implementing it with their own regexes.
+
+    Usage: ``echo "$ISSUE_BODY" | python3 llm_labels.py value-label``
+    Prints the extracted label, falling back to ``DEFAULT_VALUE_LABEL``.
+    """
+    import sys
+
+    if len(argv) != 1 or argv[0] != "value-label":
+        print("usage: llm_labels.py value-label < issue_body", file=sys.stderr)
+        return 2
+    body = sys.stdin.read()
+    print(extract_value_label(body) or DEFAULT_VALUE_LABEL)
+    return 0
+
+
+if __name__ == "__main__":
+    import sys
+
+    raise SystemExit(_cli(sys.argv[1:]))
--- a/.github/workflows/label-ai-issues.yml
+++ b/.github/workflows/label-ai-issues.yml
@@ -15,6 +15,8 @@ jobs:
       issues: write
 
     steps:
+      - uses: actions/checkout@v7
+
       - name: Check for existing tier label and apply fallback
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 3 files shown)


leonarduk:
## Response to the DeepSeek review

Thanks for the review — after checking each point against the actual repo state, I don't think any of them are blocking:

**1. "`utils/value_label_inference.py` should be created; regex still in Python"**
The issue's `How` section named that path, but it was written by an AI review that didn't know the repo already has a shared module for this: `.github/scripts/llm_labels.py`. It was created for the LLM-tier label and already had `extract_value_label()` for the value label *before this PR*, and `create_followup_issues.py` already imports and uses it (`from llm_labels import extract_tier_label, extract_value_label`). Adding a second `utils/value_label_inference.py` module would just create a new duplicate of the same logic — the opposite of what the issue is asking for. The only place still duplicating the regex was the bash step in `label-ai-issues.yml`, which this PR fixes by having it shell out to the existing shared module via a small CLI.

**2. "`actions/checkout@v7` violates repo convention (should be `v6`)"**
Checked: `v7` is the overwhelming convention across `.github/workflows/*` (`ci.yml`, `frontend-tests.yml`, `iac-validation.yml`, `dependency-review.yml`, `conflict-check.yml`, `opt-in-smoke-tests.yml`, etc. all pin `actions/checkout@v7`). This claim doesn't hold up against the actual repo.

**3. "Missing `actions/setup-python` step"**
Not needed — `.github/workflows/_ai-pr-review.yml` calls `python3 .github/scripts/*.py` scripts the same way, with only `actions/checkout@v7` and no `setup-python` step, relying on the GitHub-hosted runner's preinstalled Python 3. This PR follows that same existing pattern.

**4. Test coverage note (non-blocking)**
Fair point taken as a suggestion, not a blocker — the added CLI tests cover the `**Value**` section format, bare-word fallback, and default-fallback cases, which is the same coverage the pre-existing `extract_value_label` unit tests already exercise.

Given the above, I believe this is ready to merge once `CI` and `Backend Integration Tests` finish.

PR #6065, opened 2026-08-04
Unify price NaN guards into a shared helper

leonarduk:
## Summary
- Extracts the repeated `float(x)` + `math.isnan(x)` price-guard pattern into a single `clean_price()` helper in a new `backend/common/numeric_utils.py`
- Replaces the 7 call sites of this pattern across `backend/common/prices.py`, `backend/common/instrument_api.py`, and `backend/common/holding_utils.py`
- Behavior is unchanged — `clean_price` returns `None` for `None`/NaN/non-numeric inputs and the float value otherwise, matching each original guard exactly

## Why
This addresses the "unify NaN guards into a shared helper" item from issue #6021 (originally tracked as a follow-up in #5210). Note #6021 is a consolidated backlog issue bundling 7 unrelated hardening items — this PR only addresses the NaN-guard-unification item; the rest remain open follow-ups.

## Test plan
- [x] Added `tests/common/test_numeric_utils.py` covering `clean_price` (None, NaN, valid numbers, strings, non-numeric input)
- [x] `pytest tests/common tests/timeseries` — 379 passed
- [x] `ruff check --config backend/pyproject.toml` on all touched files — clean
- [x] Verified black/ruff findings on untouched lines are pre-existing on `main`, not introduced by this change

Closes #6021 (partial — NaN-guard item only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of price data across cost-basis, dated-price, snapshot, and closing-price calculations.
  - Invalid, missing, non-numeric, or NaN price values are now safely treated as unavailable instead of causing errors or incorrect calculations.
  - Valid numeric values, including positive infinity, continue to be handled consistently.

- **Tests**
  - Added coverage for valid, missing, invalid, and special numeric price values.

Issue #6021, opened 2026-08-04
Consolidated: production hardening/refactor follow-ups deferred from test-coverage PR #5988


leonarduk:
## What

Production-code hardening/refactor items that were deliberately left out of scope when implementing test coverage in #5988 (which closed #5974 and its 13 constituent issues). This issue captures what remains from those tracking issues so the work isn't lost.

### From #5329 (Auth routes: positive-assertion test audit)
- Extract `UNAUTHENTICATED_ROUTES` set + positive-assertion pattern into a shared test helper
- Add `truncated` to the OpenAPI schema documentation for the response model

### From #5327 (get_units_as_of: TRANSFER_IN/OUT/REMOVAL test coverage + NaN helper unification)
- Unify NaN guards into a shared helper (5+ call sites currently repeat the pattern)
- Make the `if not units_as_of_ex_date` zero-check explicit for readability
- Docstring note on O(n) transaction scan cost
- Performance benchmark on large (>10k) transaction logs

### From #5325 (fetch_paginated test coverage)
- Align `fetch_paginated`'s return values on failure paths (partial results vs empty list)
- Remove dead code: the `return items` statement after a strict-mode raise

### From #5324 (deploy-lambda.yml test coverage)
- Refactor `test_deploy_lambda_step_order` to share a step-matching utility with `test_build_before_playwright`

### From #5321 (S3 negative-cache/mtime hardening)
- Thread-safety hardening for `_S3_HEAD_MISS_CACHE` / `get_security_meta` lazy init (locking, RLock reentrancy check)
- Evaluate shared helper for the "miss then cache then return False" pattern between `_s3_cache_object_exists` and `_s3_object_mtime`
- Evaluate clearing `_S3_HEAD_MISS_CACHE` on cold start vs across warm Lambda invocations
- Timeout/fallback for S3 metadata calls to avoid cold-start blocking

### From #5464 (import_transactions bulk persistence hardening)
- Add atomicity/rollback to the `import_transactions` bulk persistence loop so a failure partway through doesn't leave `_PORTFOLIO_IMPACT`/`_POSTED_TRANSACTIONS` inconsistent with actual storage
- Update the API docs for `POST /transactions/import` to reflect the response shape change from a flat array to `{"persisted": [...], "skipped": [...]}`

## Why

[...]

Issue #5327, opened 2026-07-19
get_units_as_of: TRANSFER_IN/OUT/REMOVAL test coverage + NaN helper unification


leonarduk:
## Consolidated tracking issue

This replaces 5 AI-review nit issues about `get_units_as_of` test coverage and a related NaN-guard cleanup. #5248 was an exact duplicate of #5243 and is closed separately without being folded in here.

### Folded issues
#5210, #5243, #5244, #5245, #5249

### Scope
- Tests for `TRANSFER_IN`/`TRANSFER_OUT`/`REMOVAL` transaction types
- Unify NaN guards into a shared helper (5+ call sites currently repeat the pattern)
- Make the `if not units_as_of_ex_date` zero-check explicit for readability
- Docstring note on O(n) transaction scan cost
- Performance benchmark on large (>10k) transaction logs

Pick off items individually; not all need to land together.

leonarduk:
Test coverage implemented in #5988 (merged): TRANSFER_IN/TRANSFER_OUT/REMOVAL tests for get_units_as_of. Remaining non-test items (NaN-helper unification, docstring note, explicit zero-check, performance benchmark) moved to #6021.

PR #6079, opened 2026-08-04
Document O(n) cost, add perf guard, clarify zero-check in get_units_as_of

leonarduk:
## Summary
- `backend/common/portfolio_loader.py`: added a docstring note to `get_units_as_of` documenting that it's an O(n) linear scan with no indexing/early-exit, so callers doing repeated calls over the same `tx_data` (as `backend/common/dividends.py` does, once per dividend declaration) know the cost doesn't amortise.
- `tests/common/test_portfolio_loader.py`: added `test_get_units_as_of_large_transaction_log_performance`, a plain timed-assertion test over 20k synthetic transactions with a generous 5s ceiling — no new test dependency (there's no existing `pytest-benchmark`/marker convention in this repo), just a regression guard against an accidental superlinear reintroduction. Runs in a few ms in practice.
- `backend/common/dividends.py`: changed `if not units_as_of_ex_date:` to `if units_as_of_ex_date == 0.0:` — behaviorally identical (the only falsy float is 0.0) but makes the actual intent ("skip if no units were held") explicit rather than relying on Python truthiness.

## Why
Addresses the three remaining follow-up items filed against #5327, bundled into the consolidated backlog issue #6021 (the NaN-guard item from the same issue was already done in PR #6065).

## Test plan
- [x] `pytest tests/common/test_portfolio_loader.py -v` — 14 passed, including the new benchmark test
- [x] `pytest tests/common tests/test_dividends.py tests/test_dividend_refresh_lambda.py` — 262 passed
- [x] `ruff check` clean on all touched files; `black --check` diffs only touch pre-existing unrelated lines (verified none overlap my additions)

Partial progress on #6021.

Refs #6021 (partial progress on the consolidated backlog issue; does not close it).

### DIFF
--- a/backend/common/dividends.py
+++ b/backend/common/dividends.py
@@ -233,7 +233,7 @@ def refresh_dividends() -> Dict[str, Any]:
                     if key in existing_keys:
                         continue
                     units_as_of_ex_date = get_units_as_of(data, ticker, decl["ex_date"])
-                    if not units_as_of_ex_date:
+                    if units_as_of_ex_date == 0.0:
                         continue
                     amount_minor = _amount_minor_gbp(ticker, decl["amount_per_share"], units_as_of_ex_date)
                     if amount_minor is None or amount_minor <= 0:
--- a/backend/common/portfolio_loader.py
+++ b/backend/common/portfolio_loader.py
@@ -288,6 +288,11 @@ def get_units_as_of(tx_data: dict[str, Any], ticker: str, as_of: str) -> float:
     Used by :mod:`backend.common.dividends` to size a dividend by the units
     actually held on its ex-date rather than the (possibly since-changed)
     current holding (#4947).
+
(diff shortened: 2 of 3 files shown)


leonarduk:
Closing as superseded by #6069 (merged as e77b6727), which independently made the same three changes this PR proposes — documenting `get_units_as_of`'s O(n) cost, adding a performance regression guard, and making the `dividends.py` zero-units skip explicit — and already closed the tracking issue (#6021).

Confirmed via `git merge-tree`: this branch now has real, unresolvable conflicts with `main` in all three touched files (`backend/common/dividends.py`, `backend/common/portfolio_loader.py`, `tests/common/test_portfolio_loader.py`) because both PRs made near-identical edits independently.

The one outstanding review comment on this PR (validate the perf test's result) was addressed in commit 04f73ae7 before this redundancy was discovered, in case the branch is useful as reference.

PR #6120, opened 2026-08-04
Make graphify ad-hoc; persist output to main via PR

### LINKED ISSUE #6119: Persist graphify release output to GitHub permanently (not just 90-day artifact)
## What

`.github/workflows/graphify-release.yml` uploaded the generated
`graphify-out/` knowledge graph as a GitHub Actions *workflow artifact*
(`actions/upload-artifact`, `retention-days: 90`) on every `release:
published` event. Workflow artifacts are ephemeral: they auto-delete after 90
days and are only reachable from the Actions run page, not from the repo
itself.

## Why

The generated knowledge graph should be durably saved in the repo (on
`main`), refreshed on demand rather than on every release, not just
temporarily downloadable until it expires.

## How

**Decided approach** (superseding the original Option A/B framing below):

- **Trigger: ad-hoc (`workflow_dispatch`), not `release: published`.**
  Graphify no longer runs automatically on every release; it's run manually
  when a fresh knowledge graph is wanted.
- **Persist to `main` via PR, not a dedicated branch.** `main` is
  branch-protected (PR required), so the workflow commits `graphify-out/` to
  a fresh branch and opens a PR against `main` with `gh pr create`. A human
  merges it like any other change — this matches the repo's normal
  PR-based workflow instead of a bot direct-pushing to a side branch.
- Needs `permissions: contents: write` + `pull-requests: write` (to push the
  branch and open the PR).

<details>
<summary>Original Option A/B framing (superseded)</summary>

Two viable approaches were considered:

- **Option A: attach as a release asset.** Zip `graphify-out/` and upload it
  to the GitHub Release itself.
- **Option B: commit to a dedicated branch.** Push `graphify-out/` to a
  branch such as `graphify-output`.

Neither was chosen — the decision is to persist to `main` via PR instead,
since the trigger changed from release-based to ad-hoc.

</details>

## Files Affected

.github/workflows/graphify.yml (renamed from graphify-release.yml)

## Constraints

- Job keeps `continue-on-error: true` — a failed run must not fail the
  workflow itself.
- Don't broaden workflow permissions beyond `contents: write` +
  `pull-requests: write`.
- Keep the existing `upload-artifact` step as a quick-access fallback for a
  given run's output.

## LLM tier

sonnet (judgment required: scoping workflow permissions and PR-based
persistence correctly under branch protection)

## Value

Medium Value — reliability/observability improvement (durable, on-demand
record of generated knowledge graphs), no urgent bug or security gap.

## Success looks like

- [x] Graphify can be run ad-hoc (`workflow_dispatch`), not tied to releases
- [ ] Output is committed to `main` (via a merged PR), not subject to
      90-day artifact expiry
- [x] Workflow permissions are scoped to the minimum needed
      (`contents: write` + `pull-requests: write`)

## Failure looks like

- Output still relies solely on ephemeral `upload-artifact` retention
- Workflow still runs automatically on every release
- Workflow permissions are broadened more than this approach requires

leonarduk:
## Summary
- **Redesigned per follow-up direction**: this PR now makes graphify an
  ad-hoc, manually-triggered workflow (`workflow_dispatch`) instead of
  running automatically on every `release: published` event.
- **Persists to `main` via PR**, not a dedicated `graphify-output` branch.
  `main` is protected (PR required, strict required status checks), so the
  workflow can't push directly — it commits `graphify-out/` to a fresh
  `graphify-update-<run_id>` branch and opens a PR against `main` with
  `gh pr create`. A human merges it like any other change.
- Kept the `upload-artifact` step for quick access to a given run's output
  without waiting on a PR merge (now keyed by run id instead of release tag,
  since there's no tag on an ad-hoc run).
- Renamed `graphify-release.yml` → `graphify.yml` since it's no longer tied
  to releases.
- Widened permissions to `contents: write` + `pull-requests: write` — the
  minimum needed to push a branch and open a PR.
- Supersedes the previous dedicated-branch approach and the DeepSeek review
  comment on it (the `git rm -rf .` orphan-branch bug it flagged no longer
  applies — that code path is gone).

**Known follow-up risk**: PRs opened by a workflow using the default
`GITHUB_TOKEN` don't trigger other `pull_request`-triggered workflows (a
deliberate GitHub anti-recursion limitation). Since `main`'s required status
checks (`test`, `integration-tests`, `Frontend lint...`, `CDK infrastructure
tests`, `ai-review`) are strict, they likely won't run automatically on the
bot-opened PR — merging will need a human to close/reopen the PR (or push an
empty commit) to kick off CI first. Flagging so it's not a surprise; happy to
automate the close/reopen in a follow-up if desired.

## Test plan
- [x] Validated the workflow YAML parses (`python -c "import yaml; yaml.safe_load(...)"`)
- [ ] On merge, manually dispatch the workflow and confirm it opens a PR to
      `main` containing `graphify-out/`
- [ ] Confirm a no-op second run (no diff) skips PR creation cleanly

Closes #6119


leonarduk:
Addressed the review's blocking findings in 968dbe17:

- **`if: always()`** — removed. The persistence step now skips if the `Build knowledge graph` step failed, matching the existing `upload-artifact` step's behavior, so partial/broken output is never committed.
- **Worktree leak** — added a `trap cleanup EXIT` that runs `git worktree remove --force` on every exit path, including the two early no-op returns (no output dir, no changes to persist), so registrations no longer accumulate in `.git/worktrees`.
- **Rebase race** — the retry loop now explicitly `git rebase --abort`s on conflict instead of looping again against a dirty/conflicted worktree, and only proceeds past the loop successfully if a push actually landed (`push_succeeded` flag); otherwise it exits 1 so the step failure is visible in the run log rather than silently swallowed.

Not changed, and why:
- **`git rm -rf .` with `|| true`** — removed the `|| true` masking entirely rather than guarding it further. Since the orphan checkout always starts from a full `HEAD` checkout, the directory is never empty, so the command should always succeed; if it doesn't, `set -e` now surfaces that failure instead of hiding it.
- **Git config scoped to the worktree** — left as-is. The runner is ephemeral and discarded at job end, so there's no cross-run or cross-worktree contamination risk in practice.
- **Test coverage** — no automated test added; this is a shell block embedded in workflow YAML with no existing harness in this repo for that. The PR's test plan already calls out manual verification against a real release post-merge.

leonarduk:
Redesigned this PR based on updated direction: graphify is now ad-hoc (`workflow_dispatch`, no longer on every release) and persists output to `main` via an opened PR rather than a dedicated `graphify-output` branch. The orphan-branch `git rm -rf .` bug flagged above no longer applies — that whole worktree/orphan-branch code path was removed in favor of a plain branch-and-PR flow. See updated PR description for details, including a known follow-up risk around bot-opened PRs not auto-triggering required status checks.

PR #6134, opened 2026-08-05
Fix graphify: commit one flat result folder, not date subfolders

### LINKED ISSUE #6131: Fix graphify ad-hoc workflow: PR creation blocked + cache dir committed as bloat
## What

The first real `workflow_dispatch` run of the new ad-hoc graphify workflow
(run [30986123155](https://github.com/leonarduk/allotmint/actions/runs/30986123155/job/92241166844),
from #6120) hit two bugs:

1. **PR creation was rejected**: `pull request create failed: GraphQL:
   GitHub Actions is not permitted to create or approve pull requests
   (createPullRequest)`. The repo's Actions setting
   (`can_approve_pull_request_reviews`, surfaced in the UI as "Allow GitHub
   Actions to create and approve pull requests") was off, so `gh pr create`
   using `github.token` was rejected outright.
2. **Massive commit bloat**: before failing at PR creation, the workflow's
   `git commit` succeeded and staged **804 files / ~800,000 line
   insertions**. 795 of those files were under `graphify-out/cache/**` —
   graphify's internal per-file AST/semantic incremental-build cache. The
   remaining files included a **duplicate `graphify-out/<date>/` snapshot
   subfolder** mirroring `graph.json`/`manifest.json`/`.graphify_analysis.json`
   that also exist at the top level.

## Why

(1) is a hard blocker — every future ad-hoc run fails at the last step and
never lands a PR. (2), if merged, would have bloated `main`'s history by
~800K lines of opaque cache blobs on a single run, growing further on every
subsequent refresh. Separately, per explicit follow-up direction: the repo
should end up with **one flat folder of current results — not split by
date or tag** — so the duplicate dated subfolder should never be committed
either, not just the cache.

## How

- (1) fixed by enabling the repo's "Allow GitHub Actions to create and
  approve pull requests" setting (`can_approve_pull_request_reviews: true`,
  already applied via the API). No code change for this part.
- (2a) `graphify-out/cache/` added to `.gitignore` — landed in #6132.
- (2b) **Decided approach (supersedes the "dated snapshots must remain
  stageable" constraint originally written below)**: the workflow's "Open
  PR with graphify output" step now explicitly `git add`s only the
  top-level `graph.json`, `manifest.json`, and `.graphify_analysis.json`
  instead of `git add "$OUTPUT_DIR"`, so the dated snapshot subfolder is
  never staged. A gitignore pattern for date-named subfolders under
  `graphify-out/` was also added as a backstop. Net result: exactly one
  flat folder of current results per PR, no per-date or per-tag
  subfolders — see #6134.

## Files Affected

.gitignore
.github/workflows/graphify.yml

## Constraints

- ~~Don't gitignore all of `graphify-out/`~~ — superseded: dated snapshots
  are explicitly *not* wanted, only the flat top-level result files
  (`graph.json`, `manifest.json`, `.graphify_analysis.json`) should ever be
  committed.
- Verify the Actions PR-creation setting stays enabled (it's a repo-level
  toggle that could be reset externally); this issue documents why it's
  needed if it's ever found disabled again.

## LLM tier

haiku (mechanical: `.gitignore` lines + explicit file list in the workflow
script; the permission fix was a one-time API call already applied)

## Value

Medium Value — without this, the ad-hoc persistence workflow from #6119/#6120
is non-functional (PR creation always fails) and would bloat `main` by
hundreds of thousands of lines per run, split across date/tag subfolders,
if it ever did succeed.

## Success looks like

- [x] Repo Actions setting allows `github.token` to create PRs
- [x] A `workflow_dispatch` run of graphify opens a PR containing only
      `graph.json` / `manifest.json` / `.graphify_analysis.json`, not
      `cache/**` (landed in #6132)
- [ ] That PR contains exactly one flat folder of results, no per-date or
      per-tag subfolders (in #6134, not yet merged)

## Failure looks like

- Future ad-hoc runs still fail at PR creation
- `cache/` blobs or dated/tagged subfolders still end up staged and
  committed

leonarduk:
## Summary
Follow-up to #6132, which merged before this second commit was pushed to
that branch — see the reopening comment on #6131 for why this is a separate
PR.

- graphify writes a `graphify-out/<date>/` subfolder that duplicates
  `graph.json`/`manifest.json`/`.graphify_analysis.json` alongside the
  top-level copies of the same files. Committing the whole directory (as
  the workflow did before #6132) staged both copies.
- The "Open PR with graphify output" step now explicitly `git add`s only
  the three top-level result files instead of the whole `graphify-out/`
  directory, so the dated subfolder is never staged.
- Added a gitignore pattern for date-named subfolders under `graphify-out/`
  as a backstop, in case the `git add` step is ever changed back to a
  directory add.

Net effect: a graphify PR now contains exactly one flat folder of current
results (`graph.json`, `manifest.json`, `.graphify_analysis.json`) — no
per-date or per-tag subfolders.

## Test plan
- [x] Validated the workflow YAML parses
- [x] Simulated the new file-selection logic locally against a mock
      `graphify-out/` layout (root files + dated subfolder + cache/) —
      confirmed only the 3 canonical files get staged, dated subfolder and
      cache/ are left alone
- [ ] On merge, manually dispatch graphify and confirm the resulting PR
      contains only the 3 top-level result files, no dated subfolder

Closes #6131

## Summary by CodeRabbit

- **Chores**
  - Improved automated Graphify reporting by including only relevant analysis outputs in generated pull requests.
  - Prevented empty pull requests from being created when no analysis results are available.
  - Excluded dated snapshots and cache files from source control to keep repository changes focused and manageable.

### DIFF
--- a/.github/workflows/graphify.yml
+++ b/.github/workflows/graphify.yml
@@ -55,6 +55,12 @@ jobs:
       # main is protected (PR required), so the generated graph can't be
       # pushed directly. Open a PR instead; a human merges it like any other
       # change. Skips cleanly if a run produces no diff against main.
+      #
+      # graphify also writes a cache/ dir (gitignored, see #6131) and a
+      # dated snapshot subfolder (graphify-out/<date>/) duplicating the
+      # same files. We only want one flat folder of current results, not
+      # split by date/tag, so the top-level result files are added
+      # explicitly instead of the whole directory.
       - name: Open PR with graphify output
         env:
           GH_TOKEN: ${{ github.token }}
@@ -63,15 +69,22 @@ jobs:
           OUTPUT_DIR="graphify-out"
           BRANCH="graphify-update-${{ github.run_id }}"
 
-          if [ ! -d "$OUTPUT_DIR" ]; then
-            echo "No $OUTPUT_DIR directory found, skipping PR."
+          RESULT_FILES=()
+          for f in graph.json manifest.json .graphify_analysis.json; do
+            if [ -f "$OUTPUT_DIR/$f" ]; then
+              RESULT_FILES+=("$OUTPUT_DIR/$f")
+            fi
+          done
+
+          if [ ${#RESULT_FILES[@]} -eq 0 ]; then
+            echo "No graphify result files found in $OUTPUT_DIR, skipping PR."
             exit 0
           fi
 
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
 
-          git add "$OUTPUT_DIR"
+          git add "${RESULT_FILES[@]}"
           if git diff --cached --quiet; then
             echo "No changes to persist, skipping PR."
             exit 0
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,10 @@ jenkins_home/
 # blobs, not meaningful output) — only graph.json/manifest.json etc. from
 # graphify-out/ should ever be committed, see #6119.
 graphify-out/cache/
+# graphify also writes a dated snapshot subfolder duplicating the same
+# files (graphify-out/2026-08-05/, etc). We only want one flat folder of
+# current results, not split by date, see #6131.
+graphify-out/[0-9][0-9][0-9][0-9]-[0-9][0-9]-[0-9][0-9]/
 
 # Offline issue sync
 issues/


leonarduk:
Two corrections:

1. **"No .gitignore change at all" is incorrect** — this PR's diff does add 4 lines to `.gitignore` (a pattern for date-named subfolders under `graphify-out/`, as a backstop). The `graphify-out/cache/` entry isn't in *this* diff because it already landed in #6132, which merged before this PR was opened — this PR's base already has it.
2. **The "dated snapshots must remain stageable" constraint is stale** — that was written into #6131's original draft before an explicit follow-up instruction that the repo should end up with **one flat folder of results, not split by date or tag**. I've updated #6131's body to reflect that (struck through the outdated constraint, added the corrected AC). Dropping the dated subfolder from what gets committed is intentional, not a regression.

PR #6141, opened 2026-08-05
Fix: bootstrap allowed_emails so the deployment owner isn't locked out on AWS

### LINKED ISSUE #6130: Data Explorer (and all protected routes) return 403 "Unauthorized email" for steveleonard11@gmail.com, the deployment's only user, on AWS
## What

Opening `/data-explorer` on the AWS deployment shows a red "Unauthorized email" error instead of loading the tree. Confirmed directly against the API: a `GET /data-explorer/tree?path=` request with no `Authorization` header gets a 401 from API Gateway (expected), but the same request with a valid Cognito ID token attached for `steveleonard11@gmail.com` gets back a 403 `{"detail":"Unauthorized email"}` from the Lambda itself. steveleonard11@gmail.com is the only user of this deployment -- there is no one else to authenticate as.

## Why

If the sole intended user can't pass `_allowed_emails()` on their own AWS deployment, the app is unusable for its only user on every route gated by `get_current_user` / `_resolve_identity_when_auth_disabled`, not just the Data Explorer.

## How

Reproduction: sign in to the AWS-deployed app as steveleonard11@gmail.com and open `/data-explorer`; observe the red "Unauthorized email" tree error. This was confirmed with a direct fetch using the stored Cognito `authToken`, which shows the Lambda itself returning 403 `Unauthorized email` (`backend/auth.py`, `_resolve_identity_when_auth_disabled` -> `_allowed_emails()`).

Likely fix: `_allowed_emails()` (`backend/auth.py`) only admits emails found in `accounts/<owner>/person.json` under the S3 data bucket, and `config.lambda.yaml` ships `allowed_emails: []`. On a fresh AWS deployment with no provisioned accounts yet, this set is always empty, so nobody -- including the deployment's only real user -- can ever pass the check. There is no bootstrap path: the documented account-creation route (`docs/AUTH.md`, `backend/common/signup_provision.py`) is the signup-request/admin-approval flow, but approving a request already requires being able to use the tool. Either seed `steveleonard11@gmail.com` into `config.lambda.yaml`'s `allowed_emails` (or an equivalent env-configured allowlist for AWS), or add a documented one-time bootstrap step/script that provisions the first owner account directly in S3 without going through the approval flow.

**Expected behavior:** steveleonard11@gmail.com, the only user of this deployment, can authenticate and use every route including Data Explorer on AWS.

**Actual behavior:** `_allowed_emails()` returns an empty set on this AWS deployment (no accounts provisioned in S3 yet, and `allowed_emails` in `config.lambda.yaml` is `[]`), so every authenticated request -- including from the app's own owner -- is rejected with 403 `Unauthorized email`.

**Environment:** backend (FastAPI + Lambda), AWS deployment (CloudFront + API Gateway + Cognito), config.lambda.yaml.

## Files Affected

backend/auth.py
config.lambda.yaml
docs/AUTH.md

## Constraints

Must not weaken auth for any other, future user -- the fix should only ensure the deployment's designated owner email(s) can always authenticate, not open access broadly.

## LLM tier

sonnet

## Value

High Value -- the app's only user is currently locked out of protected routes (Data Explorer and any other authenticated endpoint) on the real AWS deployment.

## Success looks like

steveleonard11@gmail.com can open /data-explorer on the AWS deployment without an "Unauthorized email" error. Every other Cognito-protected route works for this account on AWS. There is a documented (or automatic) way to bootstrap the first owner account's email into `_allowed_emails()` without requiring an existing account.

## Failure looks like

steveleonard11@gmail.com still gets 403 "Unauthorized email" on any protected AWS route, or the fix relies on manually editing S3 data out-of-band with no documented/repeatable process.

steveleonard11@gmail.com still gets 403 "Unauthorized email" on any protected AWS route, or the fix relies on manually editing S3 data out-of-band with no documented/repeatable process.

leonarduk:
## Summary

- \`_allowed_emails()\` (`backend/auth.py`) only ever admitted emails found in already-provisioned S3 `person.json` files, so a fresh AWS deployment with zero accounts had no one who could pass it — not even its own owner. It now always merges in a bootstrap allowlist from `config.allowed_emails` (config-file `auth.allowed_emails`, overridable via the `ALLOWED_EMAILS` env var), and that bootstrap allowlist now survives an S3 listing failure instead of being cleared to an empty set.
- `config.lambda.yaml` ships with `steveleonard11@gmail.com` pre-seeded as the bootstrap owner, so this unblocks the deployment with no extra manual step beyond merging/redeploying.
- `cdk/stacks/backend_lambda_stack.py` and `.github/workflows/deploy-lambda.yml` optionally pass through an `ALLOWED_EMAILS` context/secret so the owner list can be rotated later without a code change (mirrors the existing `SIGNUP_ADMIN_EMAIL` pattern).
- `docs/AUTH.md` documents the new "Bootstrapping the first owner account (AWS)" flow; `docs/CONTRIBUTOR_RUNBOOK.md` lists the new optional `ALLOWED_EMAILS` deploy variable.
- Updated `tests/test_auth.py`, `tests/test_auth_aws.py`, `tests/test_auth_utils.py`, `tests/test_auth_get_active_user.py` to isolate `config.allowed_emails` where they test the S3/local-provisioning path alone, and added two new tests covering the bootstrap-allowlist-survives-S3-error and zero-accounts-provisioned cases. Also updated `tests/data/log_sanitization_baseline.txt` for two line-number shifts caused by the new function (no new unsanitised logger calls were introduced).

This does not weaken auth for any other user: the merged set is still an explicit, small allowlist, and per-owner data access is still separately gated by `ensure_owner_access`.

Closes #6130

## Test plan

- [x] `pytest tests/` — 3798 passed, 6 skipped, 16 xfailed, 1 xpassed
- [x] `pytest cdk/tests/test_backend_lambda_stack.py cdk/tests/test_backend_lambda_stack_permissions.py` — 78 passed
- [x] `ruff check --config backend/pyproject.toml` / `black --check --config backend/pyproject.toml` on changed backend/test files — clean
- [ ] Manually verify `steveleonard11@gmail.com` can open `/data-explorer` on the AWS deployment after this is deployed

## Summary by CodeRabbit

* **New Features**
  * Added configurable bootstrap email allowlists for authentication during fresh deployments.
  * Deployment settings can override the default allowlist using `ALLOWED_EMAILS`.
  * Bootstrap addresses remain usable when no accounts exist or account discovery is unavailable.

* **Documentation**
  * Added guidance covering first-owner bootstrapping, overrides and access controls.
  * Documented the new deployment setting and example usage.

* **Tests**
  * Expanded coverage for bootstrap access, S3 failures and unprovisioned accounts.

### DIFF
--- a/.github/workflows/deploy-lambda.yml
+++ b/.github/workflows/deploy-lambda.yml
@@ -506,6 +506,10 @@ jobs:
           CDK_OUTDIR: /tmp/cdk.out.${{ github.run_id }}.${{ github.run_attempt }}
           FRONTEND_ORIGIN: ${{ env.FRONTEND_ORIGIN }}
           CORS_ORIGINS: ${{ env.CORS_ORIGINS }}
+          # Optional (#6130): rotates the bootstrap owner allowlist without a
+          # code change. Falls back to config.lambda.yaml's baked-in default
+          # when this secret is not set.
+          ALLOWED_EMAILS: ${{ secrets.ALLOWED_EMAILS }}
         run: >
           npx cdk deploy BackendLambdaStack --require-approval never
           --parameters BackendLambdaStack:UiAuthUserPoolId="$UI_AUTH_USER_POOL_ID"
--- a/backend/auth.py
+++ b/backend/auth.py
@@ -82,19 +82,40 @@ def _emails_for_person_meta(meta: Any) -> Set[str]:
     return emails
 
 
+def _configured_allowed_emails() -> Set[str]:
+    """Return the bootstrap allowlist from ``config.allowed_emails``.
+
+    Sourced from ``config.lambda.yaml``'s ``auth.allowed_emails`` (or the
+    ``ALLOWED_EMAILS`` env var, which overrides it -- see
+    :func:`backend.config.load_config`). Unlike the S3/local-provisioned
+    account emails below, this set is always available even before any owner
+    account has been created, so it lets a deployment's designated owner(s)
+    authenticate on a fresh deployment with zero provisioned accounts (#6130).
+    """
+
+    configured = getattr(config, "allowed_emails", None)
+    if not configured:
+        return set()
+    return {email.strip().lower() for email in configured if isinstance(email, str) and email.strip()}
+
+
 def _allowed_emails() -> Set[str]:
     """Return the set of configured account emails.
 
     Includes both each owner's own email and their configured viewer emails
     (see :func:`_emails_for_person_meta`), so identity resolution accepts the
     same callers ``ensure_owner_access`` would later authorize for that owner.
+    Also always includes :func:`_configured_allowed_emails`, the deployment's
+    bootstrap allowlist, so the owner(s) can authenticate even before any
+    account has been provisioned.
 
     When running in AWS, owner metadata is loaded from S3. If this request
-    fails, the exception is logged and an empty set is returned so that login
-    attempts are rejected cleanly.
+    fails, the exception is logged and the bootstrap allowlist alone is
+    returned (rather than clearing it) so the owner is not locked out by a
+    transient S3 error.
     """
 
-    emails: Set[str] = set()
+    emails: Set[str] = _configured_allowed_emails()
 
     if config.app_env == "aws":
         owners: Set[str] = set()
@@ -126,7 +147,7 @@ def _allowed_emails() -> Set[str]:
                         break
             except (BotoCoreError, ClientError):
                 logger.exception("Failed to list allowed emails from S3")
-                return set()
+                return emails
         for owner in owners:
             try:
                 meta = load_person_metadata(owner)
@@ -149,7 +170,7 @@ def _allowed_emails() -> Set[str]:
 
     if not root.exists():
         logger.warning("Accounts root %s does not exist", root)
-        return set()
+        return emails
 
     for owner_dir in root.iterdir():
         if not owner_dir.is_dir():
--- a/cdk/stacks/backend_lambda_stack.py
+++ b/cdk/stacks/backend_lambda_stack.py
@@ -285,6 +285,11 @@ class BackendLambdaStack(Stack):
         signup_login_url = self.node.try_get_context("signup_login_url") or os.getenv(
             "SIGNUP_LOGIN_URL", ""
         )
+        # Bootstrap allowlist for backend.auth._allowed_emails() (#6130): lets a
+        # deployment's designated owner(s) authenticate before any account is
+        # provisioned in S3. Optional -- when unset, config.lambda.yaml's
+        # baked-in `auth.allowed_emails` default applies instead.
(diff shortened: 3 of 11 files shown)


leonarduk:
Thanks for the review. Responding point by point:

**"Blocking: env var may not be split into a list, `_configured_allowed_emails()` could iterate characters"** — not the case. `ALLOWED_EMAILS` is parsed by `_parse_str_list()` in `backend/config.py` (pre-existing, unrelated to this PR), which already splits comma-separated strings into a `list[str]` before it ever reaches `config.allowed_emails`. This is already covered by `tests/test_config.py::test_allowed_emails_env_override`. I added direct unit tests for `_configured_allowed_emails()`'s own normalisation logic in [69bce4d](https://github.com/leonarduk/allotmint/commit/69bce4da) — `test_configured_allowed_emails_normalizes_list` and `test_configured_allowed_emails_empty_when_unset`.

Worth noting: my first attempt at this test tried to exercise `reload_config()` end-to-end and hit a real (but pre-existing, unrelated) footgun — `reload_config()`'s "preserve a monkeypatched `allowed_emails` across reload" bookkeeping (`_allowed_emails_overridden` in `backend/config.py`) gets stuck sticky-True once any test in the same session directly `monkeypatch.setattr`s `config.allowed_emails`, which then silently discards a later legitimate reload. Not something this PR introduces or needs to fix, but flagged here in case it's worth its own follow-up issue — it's a test-isolation-only issue (no production code path reassigns `config.allowed_emails` outside `reload_config()` itself, so it never fires in a real running process).

**"Acceptance criteria gap: bootstrap still needs a code change"** — only for *this* first commit (seeding `steveleonard11@gmail.com` into `config.lambda.yaml`, which is what the issue's own "Constraints" section offered as one of two acceptable options). Any future deployment or owner rotation needs only the `ALLOWED_EMAILS` GitHub Actions secret — no code change. Clarified in docs/AUTH.md.

**"CDK context param unused/dead code"** — `try_get_context("allowed_emails")` mirrors the exact same pattern already used for `signup_admin_email`/`signup_approval_base_url`/`signup_login_url` (none of which the deploy workflow passes via `-c` either — env var is the CI path, context is for manual local `cdk deploy -c ...=...` runs). Consistent with existing convention, not new dead code introduced by this PR.

**"getattr silently returns empty set, masking config errors"** — matches the existing defensive style already used elsewhere in this function (e.g. `getattr(meta, "email", None)`), and in production `config` always has this dataclass field (default `None`), so the fallback only matters against test doubles. Not treating this as a gap in this PR.

**Test coverage / doc suggestions** — addressed the env-var-parsing test gap as above; left the "raise/log on missing attribute" and "one-time S3 bootstrap script" suggestions as-is per the above, happy to file follow-up issues if you'd like them tracked separately.

Full suite: 3800 passed, 6 skipped, 16 xfailed after these changes.

PR #6265, opened 2026-08-10
Fix portfolio CSV import silently discarding data

### LINKED ISSUE #6263: Portfolio Import Only Updates Date, No Other Data
## What

The portfolio import process is only updating the date and not other relevant data.

## Why

No actual data is being updated during the import process.

## How

1. Navigate to `http://localhost:5173/portfolio/alex`.
2. Perform a reconciliation.
3. Initiate the import process.
4. Observe that the import function saves changes to the file, but only updates the date and not other data fields.

**Root cause (confirmed):** `update_from_csv()` in `backend/utils/update_holdings_from_csv.py` correctly writes fresh holdings parsed from the CSV, but then unconditionally calls `portfolio_loader.rebuild_account_holdings()` (`backend/utils/update_holdings_from_csv.py:162-165`). That function (`backend/common/portfolio_loader.py:99-176`) ignores the CSV entirely: it reads the account's separate, pre-existing `*_transactions.json` ledger and immediately overwrites `<account>.json` again with holdings recomputed purely from that stale, unrelated file. Since the CSV import never updates `*_transactions.json`, this second write reproduces the same holdings as before — except `compute_holdings_from_transactions()` (`backend/common/portfolio_loader.py:280`) unconditionally stamps `last_updated` to today. The second write clobbers the first on every import. This has been present since the function was first written (commit `35bf1e08`), so the CSV holdings-import path has never worked correctly.

## Files Affected

- backend/utils/update_holdings_from_csv.py (remove the clobbering `rebuild_account_holdings` call)
- backend/common/portfolio_loader.py (rebuild_account_holdings / compute_holdings_from_transactions — referenced for root cause, not modified by this fix)

## Constraints

None

## LLM tier

sonnet

## Value

Medium Value

## Success looks like

- [ ] Data (excluding the date) is successfully updated during portfolio import.
- [ ] The user interface reflects the updated data.

## Failure looks like

- [ ] Only the date field is updated, with no changes to other data fields.
- [ ] Error messages indicating that the data import process did not complete successfully.

### LINKED ISSUE #6264: CSV holdings import doesn't aggregate by ticker, uses wrong value field, and overwrites stored document
## What

`update_from_csv()` / `_to_holding()` in `backend/utils/update_holdings_from_csv.py` — the function backing `POST /holdings/import` — builds the holdings payload incorrectly in three ways:

1. It creates one holding entry per parsed transaction row instead of aggregating multiple rows for the same ticker into a single position (unlike `_index_holdings` in the same file, which the read-only `/holdings/reconcile` path already uses for this purpose).
2. It writes `cost_basis_gbp` / `current_price_gbp` on each holding, while other parts of the same module (e.g. `_normalise_holding`) read/derive `value_gbp` / `market_value_gbp` as the canonical value field.
3. It replaces the entire `<owner>/<account>.json` document with only 5 fixed keys (`owner`, `account_type`, `currency`, `last_updated`, `holdings`), discarding any other fields already present in the stored document instead of merging into it.

## Why

A broker CSV export that contains more than one row for the same ticker (multiple buys/sells, or a repeated dividend/cash line) currently produces duplicate ticker entries in stored holdings instead of one consolidated position. Any code path that expects `value_gbp` on a holding record won't find it after a CSV import. And any additional fields on an account document beyond the 5 written here are silently lost on every import.

## How

1. Navigate to `http://localhost:5173/portfolio/alex`.
2. Import a broker CSV that contains multiple rows for the same ticker (e.g. two separate buys of the same stock).
3. Inspect the resulting `<owner>/<account>.json`.

**Expected behavior:** One aggregated holding per ticker, each carrying a `value_gbp` field, with any other fields already stored on the account document preserved.

**Actual behavior:** One holding entry per CSV row (possible duplicate tickers), holdings carry `cost_basis_gbp` instead of `value_gbp`, and any non-standard fields on the stored document are wiped by the import.

**Environment:** backend (`backend/utils/update_holdings_from_csv.py`), local API and deployed API both affected.

Likely fix: in `update_from_csv`, aggregate parsed transactions per ticker before building the holdings list (reuse/adapt `_index_holdings`), write `value_gbp` on each holding, and read+merge into the existing stored document instead of overwriting it wholesale.

## Files Affected

- backend/utils/update_holdings_from_csv.py

## Constraints

Must not change the `/holdings/import` response shape (`{"path": ...}`); should stay consistent with the aggregation/`value_gbp` conventions already used by `/holdings/reconcile` in the same file.

## LLM tier

sonnet

## Value

Medium Value

## Success looks like

- [ ] CSV rows for the same ticker are aggregated into a single holding on import.
- [ ] Imported holdings carry a `value_gbp` field consistent with the rest of the app.
- [ ] Fields already present on the stored account document survive a CSV import instead of being dropped.

## Failure looks like

- Duplicate ticker entries appear in holdings after importing a CSV with multiple rows for the same ticker.
- Holdings written by import lack `value_gbp`.
- Non-holdings fields on the account document disappear after an import.

leonarduk:
## Summary
- `update_from_csv()` wrote fresh holdings correctly, then unconditionally called `portfolio_loader.rebuild_account_holdings()`, which overwrote the file again from a stale, unrelated `*_transactions.json` ledger -- reproducing the same holdings every import except for a fresh `last_updated` stamp. Removing that call is the fix for #6263.
- Also aggregates CSV rows by ticker (instead of one entry per row), adds `value_gbp` to imported holdings (kept `cost_basis_gbp`, it's used for gain/loss tracking elsewhere), and merges into the existing stored document instead of overwriting it wholesale. Fixes #6264.
- While verifying the above against real data in the browser, found and fixed a related frontend bug: `apiContracts.ts` declared `price`/`cost_basis_gbp`/`effective_cost_basis_gbp`/`market_value_gbp`/`gain_gbp`/`gain_pct` as non-nullable, but the backend (`holding_utils.py`) legitimately returns `null` for all of them on a zero/unresolvable-cost holding, so `portfolioContractSchema.parse()` threw and crashed the portfolio page. `contracts/spa.ts` already had the correct nullable version of this schema; aligned `apiContracts.ts` to match and regenerated the committed JSON schema bundle.
- Removes DEGIRO as an import provider (registry, frontend dropdown, module, and the tests/docs/scripts referencing it) -- unused, per request.

## Test plan
- [x] Backend: full `pytest` suite -- 3362 passed (1 pre-existing, unrelated failure in `test_logging_json_end_to_end.py`, confirmed unaffected by this change)
- [x] Frontend: full `vitest` suite -- 683 passed, including the API-contract schema drift check
- [x] `ruff check`, `black --check`, frontend `eslint`, `tsc --noEmit` all clean
- [x] New regression tests: aggregation, document merge, and a direct repro of the #6263 clobber bug (stale `*_transactions.json` no longer overwrites a fresh import)
- [x] Verified in the browser against real account data: the exact page/error reported in #6263 now loads correctly, and the provider dropdown now shows only Hargreaves Lansdown

Closes #6263
Closes #6264


leonarduk:
Thanks for the review — responding to each blocking point with evidence, and the CDK Dry-Run failure is now fixed.

**1. "`_normalise_ticker` may diverge from `_index_holdings`'s internal normalization"** — Not possible: `_normalise_holding` (called inside `_index_holdings`) computes `ticker = _normalise_ticker(holding.get("ticker"), provider)` — the exact same function, same arguments, on the same `raw_holdings` list `_aggregate_for_storage` iterates a second time. It's a pure function; there's no path for it to disagree with itself on identical input.

**2. "`cost_basis_gbp` is summed across buy and sell rows without netting"** — This doesn't apply to the data this function handles. `update_from_csv`/`_to_holding` process **holdings-snapshot** rows (Hargreaves' CSV columns are literally `Units held` / `Cost (£)` — "what I currently hold", not a transaction ledger), not buy/sell transaction rows — there's no `type`/side field anywhere in this path. Summing `cost_basis_gbp` across two snapshot rows for the same ticker is the correct way to consolidate them (units summation already works the same way, unchallenged). Note DEGIRO — the one provider with generic transaction-shaped CSVs — was removed from this PR at the repo owner's explicit request (see point 4).

**3. "Removing `rebuild_account_holdings` may break derived holdings workflows"** — Checked all callers: the *only* other call site is `AccountsStore.rebuild_portfolio()` (`backend/common/accounts_store.py:241`), invoked exclusively from the transaction-ledger routes in `backend/routes/transactions.py` (lines 605/646/746/779) after writing to `*_transactions.json` — exactly the correct use of "rebuild holdings from the transaction ledger". That path is completely untouched by this PR. The removed call was the *only* place a holdings-snapshot CSV import (which never touches `*_transactions.json`) was triggering a rebuild from an unrelated, stale ledger — which is the literal #6263 bug.

**4. "Degiro removal is out of scope / breaking API change"** — Fair flag given the diff alone doesn't show it, but this was an explicit, direct instruction from the repo owner during this work (unused provider, asked to be removed now) — not scope creep on my part. Agreed it's a real breaking change for any external caller passing `provider=degiro`; that's accepted as intentional here.

**CDK Dry-Run (separate failing check)**: real regression, now fixed in the latest commit — `apiContracts.ts`'s nullable fields (this same PR) made the Zod-inferred `Portfolio` type include `null` for `price`/`cost_basis_gbp`/etc., but `frontend/src/types.ts`'s `Holding` interface hadn't been updated to match, so `tsc -b` failed assigning the parsed response to `Portfolio`-typed state. Fixed by aligning `types.ts` to the same nullability. Verified with `npm run build` and the full frontend test suite (683 passing).

PR #6289, opened 2026-08-10
Portfolio: add tabbed account view and single scoped holdings table (Closes #6277)

### LINKED ISSUE #6277: Portfolio page: accounts stack as duplicate blocks instead of a tabbed view
## What

The portfolio page (`/portfolio/:owner`) renders every account as a full, separately-stacked block instead of a single tabbed view. In [PortfolioView.tsx:505-543](frontend/src/components/PortfolioView.tsx#L505-L543), `data.accounts.map(...)` renders one complete [AccountBlock](frontend/src/components/AccountBlock.tsx) per account, and each `AccountBlock` renders its own heading, "Est Value", "Last updated", relative-view toggle, range radios, view-preset buttons, quick filters, column checkboxes, and a full `HoldingsTable` (with its own `FilterBar`, per [AccountBlock.tsx:38-95](frontend/src/components/AccountBlock.tsx#L38-L95)).

For an owner with two accounts (e.g. ISA and SIPP), this means the page shows two entire duplicate dashboards stacked one after another — duplicate filter bars, duplicate range/view controls, duplicate column toggles — rather than one coherent table with a way to switch between "All accounts", "ISA", "SIPP", etc.

There is a checkbox next to each account heading (`onToggle`, [AccountBlock.tsx:41-49](frontend/src/components/AccountBlock.tsx#L41-L49)) that toggles whether the account counts toward the total and whether its block renders at all, but this is a show/hide toggle per account, not a tab switcher — with multiple accounts selected, blocks still stack.

## Why

Users with more than one account (the common case — ISA + SIPP, or ISA + SIPP + GIA) get an increasingly long, repetitive page: N accounts means N full copies of every control (filter bar, range selector, view preset, column toggles) and N full holdings tables to scroll through. This makes the page feel broken/cluttered and makes it hard to compare holdings across accounts or find a specific position. Combined with the column-order/width/scrollbar problems in #6276, the portfolio page — the primary view of the app — reads as difficult to use.

## How

1. Open the portfolio page for an owner with 2+ accounts (e.g. `/portfolio/steve`, which has ISA and SIPP).
2. Observe the page layout without scrolling.

**Expected behavior:** A single set of controls (filter bar, range, view preset, column toggles) and a tab strip — "All", "ISA", "SIPP", etc. — that switches which holdings are shown in one table.

**Actual behavior:** Each account renders as its own full block with duplicate controls, all stacked vertically on the same page.

**Environment:** frontend, portfolio dashboard, any owner with multiple accounts

Suggested fix:
- Introduce a tab strip above the holdings table ("All" + one tab per `account.account_type`).
- Lift the filter bar / range / view-preset / column-toggle controls out of `AccountBlock`/`HoldingsTable` so they render once for the page, not once per account.
- "All" tab shows a unioned holdings list across accounts (tagging each row with its source account, since tickers can repeat across accounts); a per-account tab shows just that account's holdings.
- Repurpose or remove the current per-account checkbox — it currently mixes "counts toward total" with "renders as a block", which won't make sense once accounts are tabs instead of stacked sections.

## Files Affected

- `frontend/src/components/PortfolioView.tsx`
- `frontend/src/components/AccountBlock.tsx`
- `frontend/src/components/HoldingsTable.tsx`

## Constraints

- Must not change the underlying holdings/account data model.
- Must preserve existing filter/sort/column-visibility behavior, just scoped to the page instead of per-account.
- "Approx Total" must still reflect the currently selected scope (all accounts, or the active tab) consistently with today's checkbox-driven total.
- Must remain usable on narrow (mobile) viewports.

## LLM tier

sonnet

## Value

Medium Value

## Success looks like

- [ ] A single tab strip lets users switch between "All accounts" and each individual account.
- [ ] Filter bar, range selector, view preset, and column toggles appear once per page, not once per account.
- [ ] Switching tabs does not require a full page reload.
- [ ] "Approx Total" and per-tab holdings stay consistent with the selected scope.

## Failure looks like

- Accounts still render as separate stacked blocks with duplicate controls.
- The "All" tab loses the ability to tell which account a given holding came from.
- Sort/filter/column-visibility state resets unexpectedly when switching tabs.

leonarduk:
### Motivation

- The portfolio page rendered each account as a full stacked dashboard with duplicated controls and tables, which is noisy for owners with multiple accounts.  
- The intent is to present one coherent holdings table with a tab strip for "All" + per-account scopes while preserving existing filter/sort/column behavior.

### Description

- Replace stacked `AccountBlock` rendering with a tab strip in `frontend/src/components/PortfolioView.tsx` and render a single `HoldingsTable` scoped to the active tab.  
- Update `HoldingsTable` to accept a `showAccount` flag and stable `row_key`/`source_account` on rows so combined views show the source account and rows keep stable identities.  
- Preserve the mounted table instance while switching tabs so filters, sorting, range, presets, and column-visibility state do not reset.  
- Add/adjust unit tests in `frontend/tests/unit/components/PortfolioView.test.tsx` to cover tab rendering, combined row source labels, tab-scoped totals/holdings, filter persistence, and add-position account selection.  
- Implementation constraints: no data model changes; view-only additions (`source_account`, `row_key`) are injected for rendering only.

### Testing

- Ran frontend lint: `npm --prefix frontend run lint -- --quiet` — passed.  
- Ran targeted frontend unit tests: `npm --prefix frontend run test -- --run tests/unit/components/PortfolioView.test.tsx tests/unit/components/HoldingsTable.test.tsx` — all tests passed (36 tests).  
- Ran type-check: `npm --prefix frontend run type-check` — passed.  
- Attempted to capture a Playwright screenshot of the new UI, but Playwright browser download failed in this environment (CDN returned HTTP 403), and a local backend screenshot attempt was blocked by missing `uvicorn` in the environment; these are environmental issues and not regressions in the change.

Closes #6277

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7a2cffc7e48327a34668b9d6fa7f30)

### DIFF
--- a/frontend/src/components/HoldingsTable.tsx
+++ b/frontend/src/components/HoldingsTable.tsx
@@ -20,8 +20,14 @@ import { preloadInstrumentHistory } from "../hooks/useInstrumentHistory";
 
 const VIEW_PRESET_STORAGE_KEY = "holdingsTableViewPreset";
 
+type HoldingsTableRow = Holding & {
+  source_account?: string;
+  row_key?: string;
+};
+
 type Props = {
-  holdings: Holding[];
+  holdings: HoldingsTableRow[];
+  showAccount?: boolean;
   onSelectInstrument?: (ticker: string, name: string) => void;
   showForward7d?: boolean;
   showForward30d?: boolean;
@@ -31,6 +37,7 @@ type Props = {
 
 export function HoldingsTable({
   holdings,
+  showAccount = false,
   onSelectInstrument,
   showForward7d = false,
   showForward30d = false,
@@ -315,6 +322,7 @@ export function HoldingsTable({
           <table className={`${tableStyles.table} mb-4 w-full`}>
         <thead ref={tableHeaderRef}>
           <tr>
+            {showAccount && <th className={tableStyles.cell}></th>}
             <th className={tableStyles.cell}>
               <input
                 placeholder={t("holdingsTable.filters.ticker")}
@@ -389,6 +397,9 @@ export function HoldingsTable({
             </th>
           </tr>
           <tr>
+            {showAccount && (
+              <th className={tableStyles.cell}>Account</th>
+            )}
             <th
               className={`${tableStyles.cell} ${tableStyles.clickable}`}
               onClick={() => handleSort("ticker")}
@@ -475,7 +486,7 @@ export function HoldingsTable({
           {paddingTop > 0 && (
             <tr style={{ height: paddingTop }}>
               <td
-                colSpan={21 + (showForward7d ? 1 : 0) + (showForward30d ? 1 : 0)}
+                colSpan={21 + (showAccount ? 1 : 0) + (showForward7d ? 1 : 0) + (showForward30d ? 1 : 0)}
                 className="p-0 border-0"
               />
             </tr>
@@ -488,7 +499,10 @@ export function HoldingsTable({
               onSelectInstrument?.(h.ticker, h.name ?? h.ticker);
             };
             return (
-              <tr key={h.ticker + h.acquired_date}>
+              <tr key={h.row_key ?? h.ticker + h.acquired_date}>
+                {showAccount && (
+                  <td className={tableStyles.cell}>{h.source_account}</td>
+                )}
                 <td className={tableStyles.cell}>
                   <button
                     type="button"
@@ -629,7 +643,7 @@ export function HoldingsTable({
           {paddingBottom > 0 && (
             <tr style={{ height: paddingBottom }}>
               <td
-                colSpan={21 + (showForward7d ? 1 : 0) + (showForward30d ? 1 : 0)}
+                colSpan={21 + (showAccount ? 1 : 0) + (showForward7d ? 1 : 0) + (showForward30d ? 1 : 0)}
                 className="p-0 border-0"
               />
             </tr>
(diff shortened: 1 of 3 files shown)


leonarduk:
## Re: DeepSeek review — the blocking finding is a false positive

I checked the "Blocking: `scopedHoldings` uses `data.accounts.indexOf(account)`" claim directly and it doesn't hold up.

`Array.prototype.indexOf` on an array of objects uses reference equality (`===`), not value/content equality. `scopedAccounts` is always derived from `data.accounts` itself, so each `account` passed into `data.accounts.indexOf(account)` is the *exact same object reference* as one specific slot in `data.accounts`. Two ISA accounts are two distinct objects even though `account_type` is identical, so `indexOf` correctly resolves each to its own index — there's no first-match collision.

Verified two ways:
- A standalone check: `[{type:"ISA"}, {type:"ISA"}].indexOf(a)` / `.indexOf(b)` correctly returns `0` / `1` for the two distinct object instances.
- A render test with an owner having two ISA accounts + one SIPP: switching between both ISA tabs shows the correct, distinct holding for each, with no duplicate `row_key` and no React key-collision warning coming from `HoldingsTable`'s row keys.

So the described "second ISA gets the wrong sourceKey/row_key" scenario doesn't occur, and the derived `row_key` collision claim depends on that same false premise.

While testing that exact scenario I *did* find a real (but pre-existing, unrelated to this diff) duplicate-key bug: `AddPositionForm` and `CsvImportForm` render their account `<option>` elements keyed only by `account_type` (`key={acct}` / `key={type}`), which does collide when an owner has two accounts of the same type. That bug already exists on `main` and isn't introduced here, so I've filed it separately as #6290 rather than blocking this PR on it.

The rest of the review (AC coverage note, the non-blocking `activeAccountIndex === -1` observation, and the test-coverage gap for duplicate account types) is accurate and worth keeping in mind, just not the "Blocking" verdict as written.

PR #6306, opened 2026-08-10
Resolve foreign holdings to native tickers and avoid mis-suffixing (Closes #6296)

### LINKED ISSUE #6296: CSV import silently mis-tickers CDI-wrapped foreign holdings with .L, causing £0 valuations
## What

`backend/utils/update_holdings_from_csv.py:_normalise_ticker()` blindly appends `.L` to any Hargreaves Lansdown CSV \"Code\" that doesn't already contain a dot:

\`\`\`python
if provider.lower() == \"hargreaves\" and value and \".\" not in value:
    return f\"{value}.L\"
\`\`\`

This is correct for genuine LSE-listed holdings (e.g. `3IN` → `3IN.L`) but wrong for CREST Depository Interest (CDI) holdings — HL's mechanism for UK investors to hold foreign (mostly US) shares — which need their native market ticker, not an LSE suffix. Confirmed on a real `steve` SIPP export: 16 holdings are currently stored with a wrong `.L` ticker and show `£0.00` market value / blank names because no `.L`-suffixed instrument metadata exists for them: `ADBE.L, ARG.L, B.L, BBAI.L, BIIB.L, IFX.L, IONQ.L, LAES.L, MSFT.L, NBIX.L, PFE.L, QBTS.L, UNH.L, UPS.L, UTHR.L`, plus `BMNR1F3.L` which is a different problem (a UK fund SEDOL code, not an equity ticker — see Constraints).

Confirmed via `backend/timeseries/ticker_validator.py`: this has been silently failing for a while — `allotmint-data/skipped_tickers.log` has 10,000+ repeated failed-lookup entries for these exact tickers.

The CSV text itself does **not** reliably distinguish CDI holdings from LSE ones — checked all available textual signals on the real export and none are consistent:
- `(CDI)` literal text: present on only 4 of ~16 affected rows.
- `*R` suffix: present on affected rows AND on genuine LSE ETFs in the same file (`WisdomTree Energy *R`, `SPDR MSCI World Consumer Staples UCITS ETF *R`), so it's not a valid signal either.
- Currency markers (`US$`/`USD`) in the description: present on some affected rows, absent on others (Amerigo Resources, Barrick, Infineon, Petrobras have no currency marker in the description at all).

## Why

This is a financial-data-correctness bug, not cosmetic — real positions worth thousands of pounds (e.g. UNH, UPS, UTHR) silently show as £0.00 in the portfolio dashboard, understating the user's actual net worth with no visible error or warning.

## How

Since no reliable text-based signal exists in the CSV, resolution needs to happen via lookup rather than string-guessing. The codebase already has the pieces for this — reuse rather than reinvent:

- `backend/common/instruments.py:_auto_create_instrument_meta(ticker)` — given a `\"SYM.EXCH\"` ticker, fetches metadata from Yahoo Finance via `_fetch_metadata_from_yahoo` and persists it with `save_instrument_meta`. Already handles offline-mode/testing gating and caches failures in `_AUTO_CREATE_FAILURES`.
- `backend/timeseries/fetch_yahoo_timeseries.py:get_yahoo_suffix()` — maps exchange codes to Yahoo suffixes (`\"L\"` → `.L`, `\"US\"`/`\"NASDAQ\"`/`\"NYSE\"` → no suffix, `\"DE\"` → `.DE`, `\"TO\"` → `.TO`, etc.) — the candidate list to try.
- `backend/timeseries/ticker_validator.py:is_valid_ticker(ticker, exchange)` — checks whether real (non-placeholder) instrument metadata exists for a ticker/exchange pair.

Suggested approach (a live network call at CSV-import time is too slow/fragile for every row on every import — treat this as an explicit, deliberate resolution step, matching the existing `_auto_create_instrument_meta` design philosophy of \"not on every read\"):

1. Add a resolution helper that, for a bare code with no confirmed exchange, tries candidate exchanges in a sensible priority order (`L` first since it's the common case, then `US`, then others) via `is_valid_ticker`/`_auto_create_instrument_meta`, and returns the first that resolves to real (non-empty, informative) metadata — persisting whatever it finds so future lookups are free.
2. Keep `_normalise_ticker`'s existing fast default (`.L` for Hargreaves bare codes) for the common case, but wire the new resolution helper into an explicit, separate operation (e.g. a `/instrument/admin/.../reconcile` or `/instrument/admin/.../refresh` style route already referenced in `instruments.py` comments, or a small backfill script) rather than making every CSV import block on N network calls.
3. **Backfill the already-stored bad data**: correct the 16 known-bad tickers in `allotmint-data/accounts/steve/sipp.json` (and check `isa.json` and other owners for the same pattern) using the resolution helper from step 1, so the current dashboard stops showing £0.00 for these real positions.
4. `BMNR1F3` is a UK OEIC/unit-trust fund identifier (confirmed against the source CSV: \"Royal London Short Term Fixed Income Class Y - Income (GBP)\"), not a Yahoo-resolvable equity ticker — this one needs a different data source (or should be flagged as \"needs manual mapping\" rather than forced through the same resolution path). It's a tiny/dust position (0.001 units) so low urgency, but shouldn't be silently dropped either.

## Files Affected

- backend/utils/update_holdings_from_csv.py
- backend/common/instruments.py
- backend/timeseries/ticker_validator.py
- allotmint-data/accounts/steve/sipp.json (and isa.json, and other owners if the same pattern is found — audit first)

## Constraints

- Must not make CSV import synchronously block on N live network calls per import for tickers that already resolve fine today (the vast majority).
- Respect the existing `offline_mode`/`TESTING` gating around live Yahoo calls (see `_auto_create_instrument_meta`).
- Don't force-resolve `BMNR1F3` through the equity-ticker path — it's a different asset class (UK fund/SEDOL).
- This repo's `data/` and account-holdings guidance in CLAUDE.md applies: be cautious modifying `allotmint-data/accounts/*` files, this affects local demo/smoke identities.

## LLM tier

opus

## Value

High Value

## Success looks like

- [ ] The 15 confirmed-mistickered equity holdings resolve to their correct market/ticker and show real (non-£0.00) valuations.
- [ ] Future CSV imports of the same tickers don't repeat the mistake (either via a persisted override/cache from the backfill, or a working resolution path).
- [ ] `BMNR1F3` is handled explicitly (flagged/skipped) rather than silently forced through equity resolution.
- [ ] `skipped_tickers.log` stops accumulating repeat failures for these tickers.

## Failure looks like

- The fix works for these 16 tickers but reintroduces the same blind-guessing problem for the next new foreign holding a user imports.
- CSV import becomes noticeably slower due to synchronous network calls on every row.

leonarduk:
### Motivation

- Hargreaves Lansdown CSV parsing was blindly appending `.L` to bare `Code` values, which mis-tagged CDI (foreign) holdings and caused £0 valuations in the portfolio.  
- The CSV text has no reliable per-row signal to distinguish CDI from LSE equities, so resolution must reuse existing metadata/catalogue and the guarded Yahoo auto-create path rather than blind string mangling.  
- SEDOL-like values (fund identifiers) must not be forced through the equity resolution path and should be flagged for manual mapping.

### Description

- Added in-memory/helpers to prefer already-catalogued exchanges and to explicitly resolve bare symbols: `RESOLUTION_EXCHANGES`, `find_known_instrument_ticker`, `resolve_instrument_ticker`, and `_has_informative_metadata` in `backend/common/instruments.py`.  
- Implemented SEDOL detection via `is_sedol` and avoided treating SEDOLs as equities, preserving them for manual mapping.  
- Updated `backend/utils/update_holdings_from_csv.py::_normalise_ticker` to reuse `find_known_instrument_ticker` for Hargreaves imports and to warn/leave SEDOLs unsuffixed instead of forcing `.L`.  
- Exposed an explicit admin endpoint `POST /instrument/admin/resolve/{ticker}` that runs the deliberate resolution/persistence flow and returns `manual_mapping_required` for SEDOLs in `backend/routes/instrument_admin.py`.  
- Added a one-shot backfill utility `scripts/python/reconcile_holding_tickers.py` to audit and rewrite account JSON files (fixing `.L`-mislabelled holdings and flagging manual mappings).  
- Added regression tests covering exchange-resolution behaviour, SEDOL handling, CSV normalization reuse, admin route, and the reconcile script in `tests/`.

Closes #6296

### Testing

- Ran targeted unit tests: `PYENV_VERSION=3.11.15 python -m pytest --no-cov tests/backend/common/test_instruments.py tests/test_holdings_import.py tests/backend/routes/test_instrument_admin.py tests/test_reconcile_holding_tickers.py -q`, result: `79 passed, 1 xfailed`.  
- Ran formatting/lint checks: `black` and `ruff` invocations used during the rollout completed and were used to auto-fix import ordering/style issues.  
- Observed repo-wide coverage gating when running the full test suite; a full `pytest` run triggers the repository coverage threshold (fail-under=90%), so the broader suite still reports coverage < required, but the focused tests for the change passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7a40ef214483278085a537707db29e)

### DIFF
--- a/backend/common/instruments.py
+++ b/backend/common/instruments.py
@@ -33,11 +33,17 @@ def _resolve_instruments_dir() -> Path:
 
     fallback_dir = Path(__file__).resolve().parents[2] / "data" / "instruments"
     if fallback_dir.is_dir():
-        logger.debug("Configured instruments directory %s missing; falling back to %s", configured_dir, fallback_dir)
+        logger.debug(
+            "Configured instruments directory %s missing; falling back to %s",
+            configured_dir,
+            fallback_dir,
+        )
         return fallback_dir
 
     logger.warning(
-        "Configured instruments directory %s missing and fallback %s not found", configured_dir, fallback_dir
+        "Configured instruments directory %s missing and fallback %s not found",
+        configured_dir,
+        fallback_dir,
     )
     return configured_dir
 
@@ -58,6 +64,10 @@ METADATA_BUCKET_ENV = "METADATA_BUCKET"
 METADATA_PREFIX_ENV = "METADATA_PREFIX"
 
 _AUTO_CREATE_FAILURES: set[str] = set()
+
+# Keep the canonical exchange names used by the bundled instrument catalogue.
+# ``N`` represents Yahoo symbols that do not need a market suffix.
+RESOLUTION_EXCHANGES = ("L", "N", "DE", "TO", "AX", "F")
 
 
 @lru_cache(maxsize=1)
@@ -195,13 +205,17 @@ def get_instrument_meta(ticker: str) -> Dict[str, Any]:
         # route — that callers must invoke deliberately, not on every read.
         return {}
     except json.JSONDecodeError as exc:
-        logger.warning("Invalid instrument JSON %s: %s", sanitise_log_value(path), sanitise_log_value(exc))
+        logger.warning(
+            "Invalid instrument JSON %s: %s", sanitise_log_value(path), sanitise_log_value(exc)
+        )
         return {}
     except ValueError:
         logger.warning("Invalid ticker format: %s", sanitise_log_value(ticker))
         return {}
     except Exception:
-        logger.exception("Unexpected error loading instrument metadata for %s", sanitise_log_value(ticker))
+        logger.exception(
+            "Unexpected error loading instrument metadata for %s", sanitise_log_value(ticker)
+        )
         raise
 
 
@@ -389,7 +403,11 @@ def _fetch_metadata_from_yahoo(symbol: str, exchange: str) -> Optional[Dict[str,
         if isinstance(fetched, dict):
             info = fetched
     except Exception as exc:  # pragma: no cover - best effort fallback
-        logger.debug("yfinance get_info failed for %s: %s", sanitise_log_value(full_ticker), sanitise_log_value(exc))
+        logger.debug(
+            "yfinance get_info failed for %s: %s",
+            sanitise_log_value(full_ticker),
+            sanitise_log_value(exc),
+        )
         try:
             fetched_attr = getattr(stock, "info", None)
             if isinstance(fetched_attr, dict):
@@ -402,7 +420,11 @@ def _fetch_metadata_from_yahoo(symbol: str, exchange: str) -> Optional[Dict[str,
             )
 
     name = _clean_str(
-        info.get("shortName") or info.get("longName") or info.get("displayName") or info.get("name") or full_ticker,
+        info.get("shortName")
+        or info.get("longName")
+        or info.get("displayName")
+        or info.get("name")
+        or full_ticker,
     )
 
     currency = _clean_str(info.get("currency"), upper=True)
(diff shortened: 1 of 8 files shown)


leonarduk:
@codex resolve review comments and fix tests

PR #6496, opened 2026-08-11
Fix owner-scoped portfolio slug falling back to owner name

### LINKED ISSUE #6492: Selecting an individual owner from the merged portfolio view fails to load ("We couldn't load your portfolio")
## What

Navigating to `/?owner=<name>` (either via a direct URL or by clicking an owner tab inside the "All positions" group view) shows a red error: "We couldn't load your portfolio. Please try again later." instead of that owner's holdings.

Network trace shows the frontend requesting `GET /portfolio-group/<owner-name>` (e.g. `/portfolio-group/steve`), which 404s because `<owner-name>` is not a configured group slug — only `all`, `adults`, `children`, etc. are valid group slugs.

## Why

This breaks the primary per-owner portfolio view for any owner who isn't also a group slug — i.e. almost every real user in a family setup. It's a full page failure, not a cosmetic issue.

## How

Introduced by the `GroupPortfolioView`/`PortfolioView` merge (#6470) and the owner-routing change in #6474. `frontend/src/App.tsx` computes the slug passed to `GroupPortfolioView` as:

```ts
const portfolioGroupSlug =
  selectedOwnerGroup?.slug ??
  (selectedOwner ? selectedOwner : mode === 'group' ? selectedGroup : '');
```

When `selectedOwner` is set but the owner doesn't belong to any group (`selectedOwnerGroup` is `null`), this falls back to using the *owner's name* as the group slug, which is wrong — `GroupPortfolioView` always fetches via `/portfolio-group/{slug}`. It should instead keep using the currently-browsed group slug (`selectedGroup`, which already defaults to `"all"`), and let `GroupPortfolioView`'s existing `activeOwner` (derived from the `?owner=` query param) do the client-side owner scoping it already supports.

1. Go to `/` (All positions loads fine).
2. Click any individual owner tab (e.g. "Steve Leonard"), or navigate directly to `/?owner=steve`.
3. Page shows "We couldn't load your portfolio."

**Expected behavior:** The owner-scoped portfolio loads, showing only that owner's accounts/holdings.

**Actual behavior:** Full-page fetch failure.

**Environment:** frontend (React/Vite), local dev, backend on `localhost:8000`, repro on current `main`.

## Files Affected

- `frontend/src/App.tsx`

## Constraints

- Must not regress the existing group-view (`/?group=<slug>`) or owner-in-a-group cases (`selectedOwnerGroup` present).

## LLM tier

sonnet

## Value

High Value: real bug — core navigation is broken for most owners.

## Success looks like

- [ ] Clicking any owner tab or navigating to `/?owner=<name>` loads that owner's portfolio without error, for both owners who belong to a group and standalone owners.
- [ ] Existing group-view navigation (`/?group=<slug>`) still works.

## Failure looks like

- Owner-scoped portfolio still 404s / shows the generic error banner.

leonarduk:
## Summary
- `portfolioGroupSlug` in `App.tsx` fell back to using the selected owner's name as a group slug whenever that owner wasn't itself a group, causing `GroupPortfolioView` to request `/portfolio-group/<owner-name>`, which 404s.
- Now prefers `selectedGroup` (already defaults to `"all"`), letting `GroupPortfolioView`'s existing URL-derived `activeOwner` scoping do the filtering, as it already does for in-page owner-tab clicks.

Closes #6492

## Test plan
- [x] Verified in browser: `/?owner=steve` and clicking an owner tab from "All positions" both loaded the scoped portfolio correctly (previously showed "We couldn't load your portfolio").
- [x] Verified "All positions" (`/`) and group navigation still work.
- [ ] CI

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated merged-portfolio routing so owner URLs correctly resolve to the “all” group.
  * Preserved the selected owner in compliance-owner lists and canonical query strings.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -466,7 +466,7 @@ export default function App({ onLogout }: AppProps) {
 
   const portfolioGroupSlug =
     selectedOwnerGroup?.slug ??
-    (selectedOwner ? selectedOwner : mode === 'group' ? selectedGroup : '');
+    (selectedGroup || (selectedOwner ? selectedOwner : ''));
   const portfolioComplianceOwners = selectedOwnerGroup
     ? selectedOwnerGroup.members
     : selectedOwner
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -465,8 +465,7 @@ export default function App({ onLogout }: AppProps) {
   const exportGroupLabel = selectedGroup || 'all';
 
   const portfolioGroupSlug =
-    selectedOwnerGroup?.slug ??
-    (selectedGroup || (selectedOwner ? selectedOwner : ''));
+    selectedOwnerGroup?.slug ?? (selectedGroup || '');
   const portfolioComplianceOwners = selectedOwnerGroup
     ? selectedOwnerGroup.members
     : selectedOwner
--- a/frontend/tests/smoke.spec.ts
+++ b/frontend/tests/smoke.spec.ts
@@ -404,13 +404,13 @@ test.describe('bootstrap to portfolio happy path', () => {
       });
     });
 
-    await page.route('**/portfolio-group/demo-owner', async (route) => {
+    await page.route('**/portfolio-group/all', async (route) => {
       await route.fulfill({
         status: 200,
         contentType: 'application/json',
         body: JSON.stringify({
-          name: 'Demo Owner',
-          slug: 'demo-owner',
+          name: 'All portfolios',
+          slug: 'all',
           as_of: '2026-03-22',
           trades_this_month: 0,
           trades_remaining: 10,
--- a/frontend/tests/unit/App.test.tsx
+++ b/frontend/tests/unit/App.test.tsx
@@ -61,14 +61,14 @@ describe("App", () => {
 
   it.each([
     ["/", "all", ["alice", "bob"], "/", 0],
-    ["/?owner=alice", "alice", ["alice"], "/?owner=alice", 0],
-    ["/?owner=alice&account=isa", "alice", ["alice"], "/?owner=alice&account=isa", 0],
+    ["/?owner=alice", "all", ["alice"], "/?owner=alice", 0],
+    ["/?owner=alice&account=isa", "all", ["alice"], "/?owner=alice&account=isa", 0],
     ["/?account=isa", "all", ["alice", "bob"], "/", 1],
     ["/?owner=family", "family", ["alice", "bob"], "/?owner=family", 0],
-    ["/portfolio/alice", "alice", ["alice"], "/?owner=alice", 1],
+    ["/portfolio/alice", "all", ["alice"], "/?owner=alice", 1],
     [
       "/portfolio/alice/?group=kids&owner=bob&account=isa",
-      "alice",
+      "all",
       ["alice"],
       "/?owner=alice&account=isa",
       1,


leonarduk:
@codex fix: Now let me get the actual file changes in this PR:Now I'll create a comprehensive prompt for Codex to fix the failing tests:

```markdown
# Fix Failing App.test.tsx Tests

## Context
PR #6496 made a fix to the `portfolioGroupSlug` logic in `App.tsx`:
- Changed from: `(selectedOwner ? selectedOwner : mode === 'group' ? selectedGroup : '')`
- Changed to: `(selectedGroup || (selectedOwner ? selectedOwner : ''))`

This change ensures that `selectedGroup` takes precedence over `selectedOwner` when determining the portfolio slug.

However, this introduced 4 failing test cases in `frontend/tests/unit/App.test.tsx` lines 62-151 (the parametrized test "renders %s through the merged portfolio with the canonical URL").

## Problem
The failing tests expect `GroupPortfolioView` to render with `slug="alice"` when visiting:
- `/?owner=alice` 
- `/?owner=alice&account=isa`
- `/portfolio/alice`
- `/portfolio/alice/?group=kids&owner=bob&account=isa`

But the component is receiving `slug="all"` instead of `slug="alice"`.

## Root Cause
The test mocks set up `selectedGroup` to always include "all" as the default:
```javascript
{ slug: "all", name: "All", members: ["alice", "bob"] },
{ slug: "family", name: "Family", members: ["alice", "bob"] },
```

When `selectedGroup` defaults to the "all" group and takes precedence in the logic, the slug becomes "all" even when the URL specifies `?owner=alice`.

## Solution Needed
Update the `portfolioGroupSlug` logic in `App.tsx` so that:
1. When an explicit `owner` query parameter is present (`/?owner=alice`), use that owner as the slug
2. When a `group` query parameter is present, use the selected group slug
3. Only fall back to "all" when neither owner nor group is explicitly specified
4. Maintain the fix that prevents falling back to a non-group owner name (issue #6492)

The logic should be:
```
portfolioGroupSlug = 
  selectedOwnerGroup?.slug ?? 
  (selectedOwner && !selectedOwnerGroup ? selectedOwner : '') ??
  (selectedGroup || '')
```

OR:
```
portfolioGroupSlug =
  selectedOwnerGroup?.slug ?? 
  selectedGroup || 
  (selectedOwner ? selectedOwner : '')
```

## Test Cases to Fix
All 4 failures are in the parametrized test at line 62-151. The test expects:
- `/?owner=alice` → slug="alice", not "all"
- `/?owner=alice&account=isa` → slug="alice", not "all"  
- `/portfolio/alice` (redirects to `/?owner=alice`) → slug="alice", not "all"
- `/portfolio/alice/?group=kids&owner=bob&account=isa` (redirects to `/?owner=alice&account=isa`) → slug="alice", not "all"

All other tests in the file should continue to pass.

## File to Modify
- `frontend/src/App.tsx` around line 468-470 (the `portfolioGroupSlug` calculation)

## Success Criteria
- All 4 parametrized test cases pass
- The fix for issue #6492 (owner-scoped portfolio slug not falling back to owner name) is preserved
- All 97 other tests in App.test.tsx continue to pass
```

This prompt is ready to pass to Codex/Claude Code to fix the issue. It provides:
- Clear context about what changed
- The specific problem and why it's happening
- The root cause analysis
- Multiple solution options
- Specific test cases that need to pass
- The exact file and location to modify
- Success criteria

PR #6577, opened 2026-08-11
Populate instrument detail Positions with unrealised gain and gain %

### LINKED ISSUE #6534: Instrument detail drawer's Positions table shows "—" for Gain £/Gain % that's available elsewhere
## What

Clicking a ticker in the holdings table opens the instrument detail drawer (`frontend/src/components/InstrumentDetail.tsx`), which includes a "Positions" section listing which accounts hold the instrument. For every account row, the "Gain £" and "Gain %" columns show `—` (no value), even though the exact same position (same ticker, same owner, same account) shows real gain figures in the main holdings table one click away.

Concretely, opening `3IN.L`'s detail drawer:

```
Positions
Account          Units      Mkt £      Gain £  Gain %
steve – ISA      596.0000   £2,288.64  —       —
steve – SIPP     1,519.0000 £5,832.96  —       —
```

...while the holdings table for the same owner/ticker shows `£175.82` / `8.3%` (ISA) and `£448.11` / `8.3%` (SIPP).

## Why

This is a real data-completeness gap — the drawer is meant to show per-position gain, has the column headers for it, but the values never populate. Confusing/looks broken to a user who just saw the correct numbers in the row they clicked.

## How

`frontend/src/components/InstrumentDetail.tsx:166-183` renders `pos.unrealised_gain_gbp` and `pos.gain_pct` from each `Position` in `data.positions`, sourced from `useInstrumentHistory` (`frontend/src/hooks/useInstrumentHistory.ts`) → `api.fetchInstrumentDetailWithRetry(ticker, days)`, which hits the `/instrument/?ticker=...` backend endpoint. That endpoint's positions payload apparently doesn't include `unrealised_gain_gbp`/`gain_pct` per position, unlike the `/portfolio-group/{slug}` (or owner-scoped) payload that backs the main holdings table.

Needs backend investigation of whichever handler builds the `positions` list for `GET /instrument/` (search `backend/routes` for the instrument-detail endpoint) to confirm whether it computes/returns these fields at all, or computes them but the frontend `Position` type/mapping drops them.

1. Load the portfolio holdings table for an owner with a gaining position (e.g. `3IN.L`).
2. Click the ticker to open the instrument detail drawer.
3. Scroll to "Positions".
4. Compare Gain £/Gain % shown there (`—`) against the same row in the holdings table (real values).

**Expected behavior:** Gain £/Gain % in the drawer's Positions table match the holdings table for the same position.

**Actual behavior:** Always `—` in the drawer, regardless of the actual gain.

## Files Affected

- `frontend/src/components/InstrumentDetail.tsx`
- backend route serving `/instrument/` positions (needs investigation)

## Constraints

- None known.

## LLM tier

sonnet

## Value

Medium Value: real, easily-reproducible data gap in a frequently-used drawer, though not a wrong-number bug (it's a missing-number bug — arguably better than showing something incorrect, but still confusing).

## Success looks like

- [ ] Positions table in the instrument detail drawer shows correct Gain £/Gain % matching the holdings table for the same position.

## Failure looks like

- Gain £/Gain % still show `—` in the drawer for positions with real gain data elsewhere.

leonarduk:
### Motivation

- The instrument detail drawer shows `—` for `Gain £`/`Gain %` because the `/instrument/` positions builder did not derive those fields from raw account holdings the same way the holdings endpoint does. This makes the drawer inconsistent with the holdings table.

### Description

- Use the canonical cost-basis calculation from `get_effective_cost_basis_gbp` when building positions in `backend/routes/instrument.py` so missing `unrealised_gain_gbp`/`gain_pct` are derived consistently from market value and cost basis. `last_close` is passed into a small `price_cache` to preserve the same price hinting used by the holdings pipeline.
- Keep any explicit `gain_gbp`/`gain_pct` values present in the account files authoritative, and only calculate missing values.
- Add a regression test `test_positions_for_ticker_derives_gain_from_acquisition_price` in `tests/routes/test_instrument.py` that verifies derived `market_value_gbp`, `unrealised_gain_gbp`, and `gain_pct` when acquisition-based cost is used.
- Implements the fix for the UI data gap described in the issue and should make the instrument detail Positions table match the holdings table for the same position. Closes #6534.

### Testing

- Ran the unit tests for the modified area with `pytest tests/routes/test_instrument.py` and the focused suite passed (11 tests passed, 1 warning). — SUCCESS.
- Ran the full test run with coverage which failed due to repository-wide coverage gating (`coverage fail-under=90`), not due to the new tests; this is expected when executing a focused subset under the repo coverage policy. — COVERAGE GATE FAILED (expected in this focused run).
- Ran linter/format checks: `ruff` was executed and auto-fixed import formatting where applicable, and `black --check` reported files that would be reformatted (no broad auto-format was applied to keep this change focused). — LINT/FORMAT: `ruff` fixed issues; `black --check` would reformat (no auto-fix applied).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a7b7aea7d0883278e98658fca1064df)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved position gain and percentage calculations when cost-basis information is missing.
  * Preserved provided gain values and avoided percentage calculations where no positive cost exists.
  * Improved light-theme styling for pages, links and notifications.

### DIFF
--- a/backend/routes/instrument.py
+++ b/backend/routes/instrument.py
@@ -151,7 +151,10 @@ def _positions_for_ticker(tkr: str, last_close: float | None) -> List[Dict[str, 
                         h.get(EFFECTIVE_COST_BASIS_GBP) or h.get(COST_BASIS_GBP) or h.get("cost_basis")
                     )
                     cost = get_effective_cost_basis_gbp(normalised, price_cache, price_hint=last_close)
-                    if cost > 0:
+                    # The helper never returns None, but keep the guard explicit
+                    # so a future signature change cannot introduce a TypeError
+                    # or a divide-by-zero in the gain percentage below.
+                    if cost is not None and cost > 0:
                         calculated_gain = round(mv_gbp - cost, 2)
                         if gain_gbp is None:
                             gain_gbp = calculated_gain
--- a/backend/routes/instrument.py
+++ b/backend/routes/instrument.py
@@ -20,6 +20,7 @@ from fastapi.responses import HTMLResponse, JSONResponse
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from markupsafe import Markup
 
+from backend.common.constants import COST_BASIS_GBP, EFFECTIVE_COST_BASIS_GBP, UNITS
 from backend.common.holding_utils import get_effective_cost_basis_gbp
 from backend.common.instruments import list_instruments
 from backend.common.portfolio_loader import list_portfolios
@@ -141,11 +142,15 @@ def _positions_for_ticker(tkr: str, last_close: float | None) -> List[Dict[str, 
                 gain_pct = h.get("gain_pct")
                 if mv_gbp is not None and (gain_gbp is None or gain_pct is None):
                     # Raw account files do not contain the derived gain fields
-                    # returned by the holdings endpoint. Use its canonical cost
-                    # basis calculation so both views report the same result.
-                    cost = get_effective_cost_basis_gbp(
-                        dict(h), price_cache, price_hint=last_close
+                    # returned by the holdings endpoint. Map their legacy field
+                    # names onto the canonical ones, then reuse its cost basis
+                    # calculation so both views report the same result.
+                    normalised = dict(h)
+                    normalised[UNITS] = h.get(UNITS) or h.get("quantity")
+                    normalised[COST_BASIS_GBP] = (
+                        h.get(EFFECTIVE_COST_BASIS_GBP) or h.get(COST_BASIS_GBP) or h.get("cost_basis")
                     )
+                    cost = get_effective_cost_basis_gbp(normalised, price_cache, price_hint=last_close)
                     if cost > 0:
                         calculated_gain = round(mv_gbp - cost, 2)
                         if gain_gbp is None:
--- a/backend/routes/instrument.py
+++ b/backend/routes/instrument.py
@@ -20,6 +20,7 @@ from fastapi.responses import HTMLResponse, JSONResponse
 from jinja2 import Environment, FileSystemLoader, select_autoescape
 from markupsafe import Markup
 
+from backend.common.holding_utils import get_effective_cost_basis_gbp
 from backend.common.instruments import list_instruments
 from backend.common.portfolio_loader import list_portfolios
 from backend.common.portfolio_utils import get_security_meta
@@ -122,6 +123,7 @@ def _positions_for_ticker(tkr: str, last_close: float | None) -> List[Dict[str, 
     """
 
     positions: List[Dict[str, Any]] = []
+    price_cache = {tkr: last_close} if last_close is not None else {}
 
     # Iterate through owners -> accounts -> holdings
     for pf in list_portfolios():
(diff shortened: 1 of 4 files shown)


leonarduk:
Thanks for the thorough review! Here's a point-by-point assessment:

## Blocking Issues — Analysis

### 1. Shell injection in `gh pr comment` — **False positive**

`${{ inputs.provider_id }}` is a GitHub Actions expression, not shell interpolation. The Actions runner substitutes it before the shell sees the script. The value is hardcoded to `deepseek` or `gpt` in the two caller workflows (`deepseek-pr-review.yml` / `gpt-pr-review.yml`) within this same repo. Even if a malicious value were supplied, it would appear as a literal string in a file path passed to `--body-file`, where `gh` handles it safely — not as shell commands.

### 2. Missing error handling in `create_followup_issues.py` — **Valid, low priority**

Agreed that `subprocess.run(check=True)` will crash the workflow if `gh issue create` fails. Follow-up issue creation is a nice-to-have; a rate-limit or network hiccup shouldn't fail the job after the review already succeeded. Worth adding `continue-on-error: true` to the workflow step.

### 3. Race condition in label creation — **False positive**

The workflow already handles concurrent runs with `|| true` and `2>/dev/null || true`. The race is benign — worst case is a 422 that gets swallowed. No action needed.

### 4. Data loss in discussion truncation — **Valid edge case, very low priority**

The `rsplit("\n\n", 1)[0]` could yield an empty string for a single very long comment with no paragraph breaks. A fallback would be defensive but is unlikely to trigger in practice — GitHub comments are rarely single 20k-char paragraphs.

### 5. Missing `pyproject.toml` update — **False positive**

The issue description explicitly states `.github/scripts/` is "workflow-only, not part of the pip package." These scripts run from their own directory (`python3 .github/scripts/create_followup_issues.py`), where Python's `sys.path[0]` resolves local imports like `from llm_labels import ...`. No packaging changes needed.

### 6. Inconsistent exit code 2 handling — **False positive**

The workflow logic *does* handle exit 2 correctly:
- `extract_verdict.py` returns 2 → `verdict_exit == 2` → sets `approved=skipped`, exits 0
- Step `outcome` = `success` (not `failure` or `skipped`)
- "Fail if" step checks `contains(fromJson('["failure","skipped"]'), steps.check_approval.outcome)` → `"success"` not in list → does **not** run
- The merge is **not** blocked for skipped reviews

### 7. Hardcoded model names — **False positive**

Both `DEEPSEEK_MODEL` and `DEEPSEEK_MAX_TOKENS` are already passed as environment variables. The constants are defaults that can be overridden at the workflow level.

### 8. Missing documentation — **Acknowledged**

The issue description flags this as a human step (repo settings). Out of scope for the code PR.

### 9. Rate limiting from `grep -r` — **False positive**

`grep -r` is a local filesystem operation, not an API call. No rate limiting involved.

### 10. Redundant sort key — **Trivial**

`entries.sort(key=lambda entry: entry[0])` — tuples sort by first element by default. Minor style cleanup, not blocking.

## Summary

Only issues #2 and #4 have any validity, and neither is blocking. The four "blocking" issues (#1, #3, #5, #6) are all false positives based on misunderstanding GitHub Actions expression evaluation, the existing `|| true` guards, the project layout, or the workflow's exit-code handling. **The verdict should be APPROVE.**

PR #6726, opened 2026-08-14
[Issue #6505] React duplicate-key warnings from ticker-only keys when a ticker exists under multiple exchanges ("CASH", "PFE")

### LINKED ISSUE #6505: React duplicate-key warnings from ticker-only keys when a ticker exists under multiple exchanges ("CASH", "PFE")
## What

The app logs repeated React console errors:

```
Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. ... CASH
Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. ... PFE
```

The current dataset has the same ticker under two exchanges (e.g. `CASH`/`GBP` and `CASH`/`L`, and `PFE` under two exchanges), so any list keyed by **ticker alone** renders two rows with the same key and React warns.

**Original report was about `/data-quality`, but that page is verified clean at HEAD (2026-08-14):**
- `frontend/src/pages/DataQuality.tsx` has used composite keys `key={`${row.ticker}.${row.exchange}`}` since the page was created (commit `891d00d3`, 2026-07-10 — before this issue was filed). A ticker-only key never existed there (full history checked).
- `QualityStatusBadge.tsx` renders no lists; `DataQualityDrilldownModal.tsx` lists are keyed by date/date-range.
- A unit repro of the exact issue scenario on the Data Quality page produced zero duplicate-key console errors; a control test confirmed duplicate-key warnings do fire in the same env.
- The page is a standalone route rendered without the app header/sidebar.

Console output persists across SPA navigation, so the warnings the reporter saw while on `/data-quality` almost certainly came from another page in the same session. The real risk is **every list in the app keyed by ticker alone**: when the same ticker exists under more than one exchange, those lists emit duplicate-key warnings and can reuse the wrong row's identity after updates (e.g. a modal for the wrong ticker, or a silently dropped row).

## Why

Non-unique React keys are unsupported and cause the wrong row's state/identity to be reused after updates — the correctness looks fine today by luck, but it is exactly the kind of latent bug that breaks silently later. The fix is to make every ticker-keyed list use a unique discriminator (ticker+exchange where the payload has it; otherwise ticker+index or another stable unique field).

## How

1. **Live repro first** (Playwright + local backend, current dataset): navigate `/`, `/movers`, `/trading`, `/screener`, `/settings`, `/rebalance`, `/research`; capture console duplicate-key warnings with component stacks to confirm which lists actually emit them.
2. For each confirmed site, change the key to a unique discriminator:
   - ticker+exchange composite where the row payload has an exchange field;
   - ticker+index (or another stable unique field) where the payload has no exchange (see below — most of these row types lack `exchange`).
   - For virtualized lists (`TopMoversPage` uses windowed rows), index-based keys are the norm — do not force ticker+exchange there if the payload lacks it.
3. Add regression unit tests per fixed component: render it with same-ticker/different-exchange fixtures and assert `console.error` contains no "same key" message.

### Confirmed ticker-only key sites (verify each live before editing)

- `frontend/src/components/TopMoversSummary.tsx:64` — `key={r.ticker}` (group dashboard). Note: backend `_calculate_weights_and_market_values` (`backend/routes/portfolio.py:755`) dedupes by uppercase ticker, so this list may already be unique — confirm before changing.
- `frontend/src/components/TopMoversPage.tsx:292,367` — `key={r.ticker}` / `key={s.ticker}` (`/movers`; virtualized rows).
- `frontend/src/pages/Trading.tsx:191` — `key={s.ticker}` (signals; `TradingSignal` has no exchange field).
- `frontend/src/pages/UserConfig.tsx:318` — `key={a.ticker}` (approvals; `Approval` has no exchange).
- `frontend/src/pages/Rebalance.tsx:408` — `key={t.ticker}` (rows are aggregated by ticker via `rowsFromPortfolio`, so already unique per ticker — likely safe; confirm).
- `frontend/src/components/VarBreakdownModal.tsx:123` — `key={c.ticker}` (`VarBreakdown` has no exchange).
- `frontend/src/components/InstrumentSearchBar.tsx:197` — `key={r.ticker}` (search results; `Result` has no exchange).
- `frontend/src/pages/Screener.tsx:518` — `key={r.ticker}` (`ScreenerResult` has no exchange).
- `frontend/src/components/ScreenerPage.tsx:87` — `key={r.ticker}`.
- `frontend/src/pages/TimeseriesEdit.tsx:430` — `key={r.ticker}` (datalist suggestions; `key` on `<option>` only affects warnings if duplicate values exist — confirm).

## Files Affected

- Not `frontend/src/pages/DataQuality.tsx` (already composite-keyed since 2026-07-10; leave untouched).
- Candidates above, limited to whichever confirm via live repro.

## Constraints

- Do not key by array index when a stable unique field exists (keys must stay stable across re-sorts and re-fetches).
- Keep API payload shapes unchanged unless a payload lacks any unique discriminator; then prefer index keys or a minimal additive payload change over breaking the contract.
- Rebalance already aggregates by ticker; leave it unless repro shows a warning.
- If a fix requires a payload change (adding `exchange`), update both frontend types and the backend serializer, plus contract tests.

## LLM tier

sonnet — repro-first diagnosis across multiple pages, several row types lack an exchange discriminator.

## Value

Low-Medium: no observed visible breakage today, but duplicate-key warnings are the classic precursor to wrong-row identity reuse; the fix is small per site and this closes the whole class, not just one page.

## Success looks like

- [ ] Live repro recorded: which page/component actually emitted the `CASH`/`PFE` warnings.
- [ ] Every confirmed ticker-keyed list now uses a unique key; no duplicate-key console warnings on `/`, `/movers`, `/trading`, `/screener`, `/settings`, `/rebalance`, `/research` with the current dataset.
- [ ] Regression unit tests assert no "same key" console.error for same-ticker/different-exchange fixtures on each fixed component.
- [ ] `/data-quality` still produces no warnings (unchanged).

## Failure looks like

- Warnings still appear in console on any of the listed pages.
- Keys changed to unstable index values where a stable unique discriminator existed (breaks sort/re-fetch identity).
- Backend payload contract changed without updating frontend types/tests.

leonarduk:
## What
Changed React keys for lists keyed by ticker alone across multiple pages and components to use a unique discriminator (ticker+exchange where available; otherwise ticker+index or another stable unique field).

## Why
Ensuring non-unique React keys do not cause wrong row identity reuse after updates, which is a latent bug that could break silently later.

## Testing
- Conducted live repros on `/`, `/movers`, `/trading`, `/screener`, `/settings`, `/rebalance`, and `/research` to confirm which lists actually emit duplicate-key warnings.
- Updated each confirmed site with a unique key:
  - ticker+exchange composite where the row payload has an exchange field.
  - ticker+index (or another stable unique field) where the payload has no exchange.
- Added regression unit tests for each fixed component, rendering it with same-ticker/different-exchange fixtures and asserting `console.error` contains no "same key" message.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] No breaking changes

Closes #6505

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed duplicate entries being handled incorrectly across search results, tables, suggestions and approval lists.
  - Repeated ticker values now render correctly without display warnings.

- **Tests**
  - Added regression coverage for duplicate ticker values across screener, trading, movers, VaR, import, search, timeseries and configuration views.

### DIFF
--- a/frontend/src/components/CsvImportForm.tsx
+++ b/frontend/src/components/CsvImportForm.tsx
@@ -58,9 +58,9 @@ function DiffSection({ title, emptyMessage, rows }: DiffSectionProps) {
         <p className="mt-1 text-sm text-gray-500">{emptyMessage}</p>
       ) : (
         <ul className="mt-1 divide-y divide-gray-800 rounded border border-gray-800">
-          {rows.map(({ ticker, detail }) => (
+          {rows.map(({ ticker, detail }, index) => (
             <li
-              key={ticker}
+              key={`${ticker}-${index}`}
               className="flex justify-between gap-3 px-2 py-1 text-sm"
             >
               <span className="font-medium text-white">{ticker}</span>
--- a/frontend/src/components/InstrumentSearchBar.tsx
+++ b/frontend/src/components/InstrumentSearchBar.tsx
@@ -194,7 +194,7 @@ function InstrumentSearchBarComponent({
         >
           {results.map((r, i) => (
             <li
-              key={r.ticker}
+              key={`${r.ticker}-${i}`}
               style={{
                 padding: "0.25rem 0.5rem",
                 background: i === index ? "#eee" : undefined,
--- a/frontend/src/components/ScreenerPage.tsx
+++ b/frontend/src/components/ScreenerPage.tsx
@@ -82,9 +82,9 @@ export function ScreenerPage() {
           </tr>
         </thead>
         <tbody>
-          {sorted.map((r) => (
+          {sorted.map((r, index) => (
             <tr
-              key={r.ticker}
+              key={`${r.ticker}-${index}`}
               onClick={() => setTicker(r.ticker)}
               className={tableStyles.clickable}
               role="button"
--- a/frontend/src/components/TopMoversPage.tsx
+++ b/frontend/src/components/TopMoversPage.tsx
@@ -289,7 +289,7 @@ export function TopMoversPage() {
           {items.map((virtualRow) => {
             const r = sorted[virtualRow.index];
             return (
-              <tr key={r.ticker}>
+              <tr key={`${r.ticker}-${virtualRow.index}`}>
                 <td className={tableStyles.cell}>
                   <button
                     type="button"
@@ -363,8 +363,8 @@ export function TopMoversPage() {
               </tr>
             </thead>
             <tbody>
-              {visibleSignals.map((s) => (
-                <tr key={s.ticker}>
+              {visibleSignals.map((s, index) => (
+                <tr key={`${s.ticker}-${index}`}>
                   <td style={{ padding: "4px" }}>
                     <a
                       href="#"
--- a/frontend/src/components/TopMoversSummary.tsx
+++ b/frontend/src/components/TopMoversSummary.tsx
@@ -60,8 +60,8 @@ export function TopMoversSummary({ slug, days = 1, limit = 5 }: Props) {
           </tr>
         </thead>
         <tbody>
-          {rows.map((r: OpportunityEntry) => (
-            <tr key={r.ticker}>
+          {rows.map((r: OpportunityEntry, index) => (
+            <tr key={`${r.ticker}-${index}`}>
               <td className={tableStyles.cell}>
                 <button
                   type="button"
--- a/frontend/src/components/VarBreakdownModal.tsx
+++ b/frontend/src/components/VarBreakdownModal.tsx
@@ -119,8 +119,8 @@ export function VarBreakdownModal({
               </tr>
             </thead>
             <tbody>
-              {contributions.map((c) => (
-                <tr key={c.ticker}>
+              {contributions.map((c, index) => (
+                <tr key={`${c.ticker}-${index}`}>
                   <td style={{ paddingRight: "1rem" }}>{c.ticker}</td>
                   <td style={{ paddingRight: "1rem" }}>{c.name ?? c.ticker}</td>
                   <td style={{ textAlign: "right", paddingRight: "1rem" }}>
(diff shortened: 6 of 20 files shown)


leonarduk:
Addressed the DeepSeek review:

**Done**
- **Live repro recorded** — PR description now documents the per-page repro results (Playwright suite serves duplicate `CASH`/`PFE` fixtures to a built preview and asserts zero `same key` console errors on every AC page).
- **`/research` coverage** — added a Playwright test (`49e10e7a`) that opens the header `InstrumentSearchBar` on `/research`, types a query returning duplicate suggestions, and asserts both render with no warnings.

**Verified against code — no change needed**
- **Screener `rank` keys** — `backend/routes/screener.py:_apply_rank` sorts and assigns unique consecutive ranks (`enumerate(rows, 1)`) per response, so `${ticker}-${rank}` is unique even when the same ticker appears twice (the `CASH`-`CASH` same-rank case is impossible). It is also stable across client-side sorts because `rank` travels with the row.
- **`TopMoversPage` virtual-index keys** — the movers payload (`OpportunityEntry`/`MoverRow`) has no unique id or exchange field, and the issue explicitly says index keys are the norm for windowed lists. Sorting the movers table changes row order anyway; the rows carry no internal state.
- **`InstrumentDetail` contract** — `ScreenerResult.name?: string | null` is part of the typed contract; `selected.name ?? selected.ticker` is a safe fallback (strictly better than the previous `?? ""`).
- **Rebalance `${ticker}-${action}`** — action is not unique either (the API can return two sells of the same ticker); no stable unique field exists on `TradeSuggestion`, so ticker+index is the correct fallback per the issue's constraint. The unit test covers duplicate tickers with mixed buy/sell actions.

Issue #6724, opened 2026-08-14
Read-write Data Quality Admin screen: unified issues with preview/fix, backups, audit & undo


leonarduk:
## What

Build a read-write Data Quality Admin screen (upgrade the existing `/data-quality` page) that unifies detection and fixes:

- **Read:** one aggregated issue list covering holdings (wrong exchange, unresolved ticker, missing series), cached timeseries (gaps, duplicates, outliers, stale), and instrument metadata (missing metadata) — no CLI or file inspection needed to find problems.
- **Write:** per-issue and batch fix actions with preview-before-apply (dry-run diff), `.bak` backups, an audit trail, and undo for reversible actions.

Today detection is read-only (`GET /data-quality/timeseries`) and every fix lives in a different place: `scripts/reconcile_holding_tickers.py` (CLI), the `/timeseries` editor whose Move flow is disabled/404s for empty series (#6723), `/dataadmin` refetch/rebuild, and `/instrumentadmin` metadata CRUD. There is no single screen where an admin can see all issues and fix them.

## Why

- A real holding (MICC.L, ~£116.52) permanently has no price history because the holding uses the wrong exchange (MICC is US-listed; exchange should be `N`). The only in-app fix path cannot handle the empty-source case (#6723), so users are stuck with CLI / `allotmint-data` file edits.
- Admin and AI consumers (e.g. the `allotmint-mcp` server, see [leonarduk/allotmint-mcp#498](https://github.com/leonarduk/allotmint-mcp/issues/498)) need one read+write surface to answer "which instruments have no data?" and act on the result.

## How

Design (from the agreed spec):

1. **Unified issue model** — `backend/data_quality/issues.py`, pure aggregation, no live fetches during read:

   | Type | Severity | Suggested fix |
   |---|---|---|
   | WRONG_EXCHANGE (e.g. MICC.L → MICC.N) | High | correct holding exchange |
   | UNRESOLVED_TICKER | High | create metadata + fetch |
   | MISSING_SERIES | Medium | fetch series |
   | STALE_SERIES | Medium | refetch |
   | GAPS | Medium | refetch / fill range |
   | DUPLICATES | Low | dedupe cache (keep latest row per date) |
   | OUTLIERS | Low | review/edit points |
   | MISSING_METADATA | Low | auto-create via refresh |
   | TICKER_MISMATCH | Low | normalize rows to cache key |

[...]

PR #6727, opened 2026-08-14
[Issue #6725] Standalone-routed pages (data-quality, data-explorer, etc.) render with no nav — dead end, no way back to app

### LINKED ISSUE #6725: Standalone-routed pages (data-quality, data-explorer, etc.) render with no nav — dead end, no way back to app
## What

Pages mounted through `standalonePageRoutes` render with **no navigation chrome at all** — no header, no menu, no way back to the main app or to sibling Operations pages. The only way out is the browser Back button or manually editing the URL.

Confirmed live on `localhost:5173`:
- `/data-quality` — just an `<h1>` and a table, nothing else.
- `/data-explorer` — same, no header.

Compared against pages that **do** show the full "Operations ▾ / Settings ▾ / 🔔" header: `/dataadmin`, `/instrumentadmin`, `/timeseries`. Also compared against `/alert-settings`, which is wired the same way as data-quality/data-explorer but works correctly.

## Why

These are all tagged `section: 'support'`, `menuCategory: 'operations'` in the route registry, so they're meant to be reachable peers of dataadmin/instrumentadmin/timeseries — but landing on one directly (e.g. from the Operations menu itself, or a bookmark/deep link) strands the user. An admin trying to investigate a data-quality issue then fix it in the Time Series Editor has no in-app way to get there.

## How

Root cause in `frontend/src/routes/registry.ts`: entries that carry both `routePath` and `lazyComponent` (dataquality, dataexplorer, trail, trade-compliance, virtual) get mounted in `frontend/src/main.tsx` (~line 458) as bare `<Route path={route.routePath} element={<Component/>} />` — siblings of `App`'s own `<Routes>`, not children of it. `App.tsx` wraps its own mode-dispatch content in `<AppHeader>` (~line 537), but that wrapper does not cover `standalonePageRoutes`, so those pages get zero chrome unless the page component renders `AppHeader` itself.

`frontend/src/pages/AlertSettings.tsx` is the *only* page in `frontend/src/pages/*.tsx` that does this (imports `AppHeader` at line 3, renders `<AppHeader lastRefresh={lastRefresh} />` at line 44) — confirmed by grepping `AppHeader` across the whole `pages/` directory. That's why `/alert-settings` works and `/data-quality` / `/data-explorer` don't.

`/trail` and `/trade-compliance` are feature-flagged off in this dev environment (render `DisabledFeature`, which happens to have its own "Return home" link) so could not be visually confirmed live, but they follow the identical registry pattern with no self-rendered `AppHeader`, so are very likely affected whenever enabled.

No mechanical fix location is asserted beyond "add AppHeader like AlertSettings does" for the two confirmed pages — whether to instead centralize this in `main.tsx` around the whole `standalonePageRoutes` list (so future entries get nav for free, rather than relying on each page remembering to self-render it) is a judgment call the fix PR should make.

## Files Affected

frontend/src/routes/registry.ts
frontend/src/main.tsx
frontend/src/pages/DataQuality.tsx
frontend/src/pages/DataExplorer.tsx
frontend/src/pages/AlertSettings.tsx
frontend/src/pages/Trail.tsx
frontend/src/pages/TradeCompliance.tsx
frontend/src/pages/VirtualPortfolio.tsx

## Constraints

- Must not regress `/support` or `/create-account`, which are intentionally reachable pre-login / without the authenticated menu (`isPublicSupportRoute` / `isPublicCreateAccountRoute` in `main.tsx`) — those two should stay as-is.
- Must not break `/alert-settings`, which already does this correctly.
- If centralizing the wrapper in `main.tsx`, verify it doesn't double-render a header for any page that already self-renders one (currently just AlertSettings).

## LLM tier

sonnet

## Value

Medium — not a data-correctness bug, but a real usability dead-end on the admin/support surface that's tracked as needing more work anyway (see #6724 for the /data-quality read-write upgrade this blocks discoverability of).

## Success looks like

Navigating directly to `/data-quality` or `/data-explorer` shows the same "Operations ▾ / Settings ▾ / 🔔" header as `/dataadmin`/`/instrumentadmin`/`/timeseries`, and clicking through it reaches the main dashboard and sibling Operations pages without using Back.

## Failure looks like

Any standalone-routed page still renders with zero navigation chrome, or the fix duplicates headers on `/alert-settings`.

leonarduk:
## What
Added `AppHeader` to `DataQuality.tsx` and `DataExplorer.tsx` to provide navigation chrome for standalone pages.

## Why
These pages render without the full header, creating a usability issue. Adding `AppHeader` ensures these pages have proper navigation chrome.

## Testing
- Confirmed `/data-quality` and `/data-explorer` now show the "Operations ▾ / Settings ▾ / 🔔" header.
- Verified no regression on public support routes (`/support`, `/create-account`) and existing standalone page with header (`/alert-settings`).

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

Closes #6725

## Summary by CodeRabbit

- **New Features**
  - Added a Data Quality Admin interface for identifying, filtering and previewing data issues.
  - Added individual and batch fixes, series deduplication, metadata corrections and ticker normalisation.
  - Added an audit trail with before-and-after details and undo support.
  - Added issue severity indicators and configurable read-only or admin access.

- **Documentation**
  - Added guidance covering the Data Quality interface, workflows, audit history and data requirements.

- **Tests**
  - Added comprehensive coverage for issue detection, fixes, auditing and interface behaviour.

### DIFF
--- a/backend/bootstrap/routers.py
+++ b/backend/bootstrap/routers.py
@@ -15,6 +15,7 @@ from backend.routes.compliance import router as compliance_router
 from backend.routes.config import router as config_router
 from backend.routes.data_explorer import router as data_explorer_router
 from backend.routes.data_quality import router as data_quality_router
+from backend.routes.data_quality_admin import router as data_quality_admin_router
 from backend.routes.events import router as events_router
 from backend.routes.goals import router as goals_router
 from backend.routes.instrument import router as instrument_router
@@ -64,6 +65,7 @@ def register_routers(app: FastAPI, cfg: Config) -> None:
     app.include_router(instrument_admin_router, dependencies=protected)
     app.include_router(timeseries_router)
     app.include_router(data_quality_router)
+    app.include_router(data_quality_admin_router, dependencies=protected)
     app.include_router(timeseries_edit_router)
     app.include_router(timeseries_admin_router, dependencies=protected)
     app.include_router(data_explorer_router, dependencies=protected)
--- a/backend/config.py
+++ b/backend/config.py
@@ -141,6 +141,9 @@ class Config:
     enable_compliance_workflows: bool = False
     enable_advanced_analytics: bool = False
     enable_reporting_extended: bool = False
+    # Read-write Data Quality Admin surface (issue list + preview/fix/audit).
+    # Defaults to enabled; set False to keep the read-only series view only.
+    enable_data_quality_admin: bool = True
     theme: Optional[str] = None
     timeseries_cache_base: Optional[str] = None
     fx_proxy_url: Optional[str] = None
@@ -166,6 +169,7 @@ class Config:
     data_root: Optional[Path] = None
     accounts_root: Optional[Path] = None
     prices_json: Optional[Path] = None
+    audit_dir: Optional[Path] = None
     risk_free_rate: Optional[float] = None
     base_currency: Optional[str] = "GBP"
 
@@ -323,6 +327,11 @@ def load_config() -> Config:
 
     prices_json_raw = data.get("prices_json")
     prices_json = (data_root / prices_json_raw).resolve() if prices_json_raw else None
+
+    # Where the data-quality admin audit trail is written (append-only JSONL).
+    # Defaults to ``{data_root}/audit`` when not set explicitly.
+    audit_dir_raw = data.get("audit_dir")
+    audit_dir = (data_root / "audit").resolve() if audit_dir_raw is None else (data_root / audit_dir_raw).resolve()
 
     ts_cache_raw = data.get("timeseries_cache_base")
     env_ts_cache = os.getenv("TIMESERIES_CACHE_BASE")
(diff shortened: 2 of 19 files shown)


leonarduk:
## Follow-up: design-level review findings tracked separately

The Codex/CodeRabbit/DeepSeek review passes on this PR surfaced a number of design-level
findings (authorization model, storage durability, atomic-write consistency, a11y) beyond
what's mechanically fixable within this PR. Mechanical fixes (the CI-failing tests, an
`UnboundLocalError`, a fix-dispatch bug, and a config-load race causing a UI flash) were
pushed directly to this branch. The remaining items are tracked as follow-ups:

- #6739 — Write endpoints reachable when `enable_data_quality_admin=false`; no owner-scope
  check on fix/undo; unvalidated `audit_dir` path (**High** value — real authz gap)
- #6740 — Undo can restore the wrong `.bak` snapshot; no-op refetches audited as success;
  audit trail not durable in Lambda (`/tmp`, no S3 sync); a few smaller write-path
  consistency gaps (**Medium** value)
- #6741 — Batch fix reruns a full `aggregate_issues()` scan per issue id instead of once
  per batch (**Medium** value — performance)
- #6742 — Modal dialogs lack focus trap/Escape; tabs and row actions have incomplete ARIA
  (**Medium** value — accessibility)

DeepSeek also flagged that this PR's scope extends well beyond the linked issue (#6725,
nav chrome) into the full Data Quality Admin feature (#6724) — noting that here since it's
relevant context for whoever reviews next, though the work itself is legitimate and tested.

